PR #230, opened 2013-09-28
Added autoheader support.

joachimmetz:
Added autoheader support; removes another auto-generated file.


joachimmetz:
no response from project

PR #279, opened 2013-12-13
Changed img layer read behavior to be POSIX compliant.

joachimmetz:
Currently the img layer reads do not have POSIX compliant behavior.

From man pread

RETURN VALUE
       On  success, the number of bytes read or written is returned (zero indicates that nothing was written, in the case of pwrite(), or end of file, in the case of pread()), or -1 on error, in which case errno is set to indicate the error.

Please consider this patch to correct this.


joachimmetz:
no response from project

PR #311, opened 2014-02-10
ResultSet closing.

raman-bt:


rcordovano:
This code no longer auto-merges and has been overcome by events. The current SleuthkitCase code has an equivalent fix, ensuring all result sets opened in SleuthkitCase methods are closed in finally blocks.

PR #323, opened 2014-03-19
Nt2unixtime fix

joachimmetz:
Fixes for behavior of nt2unixtime:
- use of time_t negative values
- use of 64-bit time_t values if possible

Removed some files in .gitignore that got changed

Fixed missing header files for deployment


joachimmetz:
no response from project

Issue #402, opened 2015-03-18
HFS+ support is partially broken in latest version


meder:
HFS+ istat in 'devel' branch and 4.1.3 release errors out on some files, but older version(3.2.3) istats and icats these files correctly.

Sample failing output 

```
...
hfs_cat_get_record_offset_cb: Leaf node want: 17 vs have: 16
hfs_cat_get_record_offset_cb: Leaf node want: 17 vs have: 17
hfs_cat_file_lookup: Looking up file record (parent: 2)
hfs_cat_traverse: starting at root node 173; nodesize = 8192
hfs_cat_traverse: node 173 @ 1417216 has 90 records
hfs_cat_get_record_offset_cb: Index node want: 2 vs have: 1
hfs_cat_get_record_offset_cb: Index node want: 2 vs have: 1963
hfs_cat_traverse: node 501 @ 4104192 has 145 records
hfs_cat_get_record_offset_cb: Index node want: 2 vs have: 1
hfs_cat_get_record_offset_cb: Index node want: 2 vs have: 16
hfs_cat_traverse: node 5652 @ 46301184 has 20 records
hfs_cat_get_record_offset_cb: Leaf node want: 2 vs have: 1
hfs_cat_get_record_offset_cb: Leaf node want: 2 vs have: 2
hfs_cat_get_record_offset_cb: Leaf node want: 2 vs have: 2
hfs_cat_get_record_offset_cb: Leaf node want: 2 vs have: 2
hfs_cat_file_lookup: found folder record valence 0, cnid 17
hfs_cat_file_lookup exiting
hfs_dinode_copy: called for file/folder 17
hfs_open: Creation times for key folders have been read and cached.
hfs_open: The root directory is accessible.
hfs_open: "/^^^^HFS+ Private Data" metadata folder is accessible.
hfs_open: "/HFS+ Private Directory Data^" metadata folder is accessible.
hfs_open: Optional Startup File is not present.
hfs_open: Extents File (and BadBlocks File) is present.
hfs_open: Attributes File is present.
iso9660_open img_info: 140737353703440 ftype: 2048 test: 1
raw_read: byte offset: 32768 len: 65536
raw_read: found in image 0 relative offset: 32768 len: 65536
iso_load_vol_desc: Bad volume descriptor: Magic number is not CD001
Trying RAW ISO9660 with 16-byte pre-block size
fs_prepost_read: Mapped 32768 to 37648
[...]

PR #409, opened 2015-03-24
Several HFS+ fixes

meder:
- Initialize HFS_INFO members earlier, since parsing code relies on them (#402)
- Do not bail when non-inline extended attributes are found
- Fix off-by-one bug

/cc @joachimmetz

### DIFF
--- a/tsk/fs/hfs.c
+++ b/tsk/fs/hfs.c
@@ -2675,7 +2675,7 @@ hfs_attr_walk_special(const TSK_FS_ATTR * fs_attr,
     }
 
     // Allocate two buffers of the compression unit size.
-    rawBuf = (char *) tsk_malloc(COMPRESSION_UNIT_SIZE);
+    rawBuf = (char *) tsk_malloc(COMPRESSION_UNIT_SIZE + 1);
     uncBuf = (char *) tsk_malloc(COMPRESSION_UNIT_SIZE);
     if (rawBuf == NULL || uncBuf == NULL) {
         error_returned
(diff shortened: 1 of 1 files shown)


tsk/fs/hfs.c
```diff
@@ -2675,7 +2675,7 @@ hfs_attr_walk_special(const TSK_FS_ATTR * fs_attr,
     }
 
     // Allocate two buffers of the compression unit size.
-    rawBuf = (char *) tsk_malloc(COMPRESSION_UNIT_SIZE);
+    rawBuf = (char *) tsk_malloc(COMPRESSION_UNIT_SIZE + 1);
     uncBuf = (char *) tsk_malloc(COMPRESSION_UNIT_SIZE);
```
karlmortensen: Why not add a +1 on the uncBuf size too?

bcarrier: Does len end up being COMPRESSION_UNIT_SIZE+1 if the unit is not compressed and the new bounds checks prevent it from processing those?

The other minor change I'll make to this code after it is merged (or you can do it if you want) would be to allocate and free uncBuf and rawBuf closer to the loop that they are used it.  THey seem to be allocated early and freed late, which increases the number of places that we need to have error handling to free them.

meder:
@karlmortensen i am not sure if it's needed there. the only time when i saw COMPRESSION_UNIT_SIZE + 1 being returned in len was when the COMPRESSION_UNIT_SIZE is uncompressed and 1 byte is added to indicate that. That is actually what the code further down on line 2866.

@bcarrier yes, len ended up being COMPRESSION_UNIT_SIZE + 1, but the bounds checking account for that. I'll push the uncBuf/rawBuf change shortly

meder:
PTAL

PR #420, opened 2015-04-06
Added new BlackboardArtifact constructor that allows us to skip the data...

APriestman:
...base lookup if the artifact is new.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
@@ -224,6 +224,30 @@ public class BlackboardArtifact implements SleuthkitVisitableItem {
 		this.artifactTypeName = artifactTypeName;
 		this.displayName = displayName;
 	}
+	
+	/**
+	 * Constructor for an artifact. Should only be used by SleuthkitCase
+	 * @param Case the case that can be used to access the database this artifact is part of
+	 * @param artifactID the id for this artifact
+	 * @param objID the object this artifact is associated with
+	 * @param artifactTypeID the type id of this artifact
+	 * @param artifactTypeName the type name of this artifact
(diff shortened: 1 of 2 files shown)


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -224,6 +224,30 @@ protected BlackboardArtifact(SleuthkitCase Case, long artifactID, long objID, in
...
+	 * Constructor for an artifact. Should only be used by SleuthkitCase
+	 * @param Case the case that can be used to access the database this artifact is part of
+	 * @param artifactID the id for this artifact
+	 * @param objID the object this artifact is associated with
+	 * @param artifactTypeID the type id of this artifact
+	 * @param artifactTypeName the type name of this artifact
+	 * @param displayName the display name of this artifact
+	 * @param isNew true if we are currently creating the artifact
+	 */
+	protected BlackboardArtifact(SleuthkitCase Case, long artifactID, long objID, int artifactTypeID, String artifactTypeName, String displayName, boolean isNew) {
```
millmanorama: if the intent is that only sleuthkitcase use this constructor, i think this is not the optimal design.  I would consider a few things.  protected is even weaker than the default access priveleges.  why should we let potential subclasses of BlackBoardArtifact have access to this.  Which leads to the question, why isn't BlackBoardArtifact final?  are there any use cases envisioned for a subclass?  making a class final eliminates a whole range of potential bugs.

Is a whole new constructor overkill for this? why not give package level access to loadedCacheFromDb and let sleuthkit case set it after the constructor call.

APriestman: The protected was a mistake - I forgot to remove it after the cut and
paste. It's supposed to be the default.

PR #437, opened 2015-04-23
Slash is added to parent file's parent path if it is missing

sidheshenator:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -2956,7 +2956,11 @@ public DerivedFile addDerivedFile(String fileName, String localPath,
 			connection.beginTransaction();
 
 			final long parentId = parentFile.getId();
-			final String parentPath = parentFile.getParentPath() + parentFile.getName() + '/'; //NON-NLS
+			String parentParentPath = parentFile.getParentPath();
+			if(!parentParentPath.endsWith("/")) {
```
karlmortensen: We can just use Java's built-in Path, right? Path xyz = Paths.get(...)

sidheshenator: The slashes are OS dependent. I think that the tsk_files.parent_path should have / as path separators irrespective of the OS.

PR #459, opened 2015-06-02
Patches for cross-compilation using MinGW

uckelman:
This PR contains four patches for compiling Windows builds of libtsk using the MinGW compiler. The issues fixed are:

* I removed narrowing casts in tests/fs_attrlist_apis.cpp which cause compilation to fail without `-fpermissive`.
* fiwalk won't build for Windows due to the absence of `getline` from <stdio.h>. I modified configure.ac to check for `getline` and decline to build fiwalk if `getline` is absent.
* I added `-no-undefined` as a flag for libtool when building using MinGW. Libtool will not produce DLLs without this flag.
* Adding the native /usr/local/include and /usr/local/lib to CPPFLAGS and LDFLAGS when cross-compiling is wrong.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,11 +84,18 @@ AX_PTHREAD([
     LDFLAGS="$LDFLAGS $PTHREAD_CFLAGS"
     CC="$PTHREAD_CC"],[])
 
-dnl Not all compilers include /usr/local in the include and link path
-if test -d /usr/local/include; then
+case "$host" in
+*-*-mingw*)
+  dnl Adding the native /usr/local is wrong for cross-compiling
+  ;;
+*)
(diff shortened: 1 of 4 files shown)


bcarrier:
This is applied to the collaborative branch. That won't be released for a while.  Develop is the more latest and greatest.  Changes to develop are periodically pushed to collaborative, but collaborative is not getting pushed to develop. Are you using collaborative?  It would help more people if this were applied to develop.

uckelman:
We're not using collaborative at present. I applied my fix to that because it looked to be ahead of develop. I'll close this and give you a patch against develop instead.

uckelman:
New PR against develop: https://github.com/sleuthkit/sleuthkit/pull/464. Thanks, Brian.

PR #481, opened 2015-06-25
better checking if we can't load a local file

bcarrier:


millmanorama:
I feel like this could be better handled by pushing the logic to read down into the different subclasses of AbstractFile: Why does AbstactFile know/care if a file is local or not?  Shouldn't that knowledge be encapsulated in the LocalFile subclass? but maybe that requires too much redesign.

bcarrier:
That was the original intention long ago, but I think it started to get complicated because DerivedFiles also use local copies of files.  However, DerivedFiles don't extend LocalFiles because there has always been an idea that DerivedFiles could be rederived as needed and never stored on disk (i.e. we would process a ZIP file as needed to extract a file from it in memory and not explode its contents to disk).  So, I think this web of inheritance lead to the place where it currently is.

PR #485, opened 2015-07-07
Acole76 patch 2

acole76:
enhancement for FLS, inserts FILE_NAME MAC times after STANDARD_INFORMATION MAC times.

### DIFF
--- a/tsk/fs/fs_name.c
+++ b/tsk/fs/fs_name.c
@@ -506,6 +506,7 @@ tsk_fs_name_print_long(FILE * hFile, const TSK_FS_FILE * fs_file,
     const char *a_path, TSK_FS_INFO * fs, const TSK_FS_ATTR * fs_attr,
     uint8_t print_path, int32_t sec_skew)
 {
+	int bNo_NTFS_TimeFields = 0;
     tsk_fs_name_print(hFile, fs_file, a_path, fs, fs_attr, print_path);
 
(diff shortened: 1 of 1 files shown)


tsk/fs/fs_name.c
```diff
@@ -506,6 +506,7 @@ tsk_fs_name_print_long(FILE * hFile, const TSK_FS_FILE * fs_file,
     const char *a_path, TSK_FS_INFO * fs, const TSK_FS_ATTR * fs_attr,
     uint8_t print_path, int32_t sec_skew)
 {
+	int bNo_NTFS_TimeFields = 0;
```
bcarrier: Is this ever used?

bcarrier:
So, this will add four columns to the 'fls -l' output, but just for NTFS.  I'd be concerned that this would break parsers that people have and would make them more complex because they would need to parse different number of columns depending on the file system. 

What is your use case for this?  'fls -m' will print sets of times, but on different lines.  Does that help?

acole76:
I see your concern about not wanting to modify the standard output of FLS
because it will beak existing parsers.  Perhaps a better idea would be to
only output the extra 4 fields if an optional argument were passed in (ex:
FLS -l -t).  My use case is to get both SI and FN timestamps on a single
line.  I think -m does get the job done, but w/ a little extra overhead.

PR #498, opened 2015-08-18
fix improper use of result set in SleuthkitCase

millmanorama:


millmanorama:
updated to use getArtifactsHelper

PR #551, opened 2015-10-23
Added --disable-multithreading option to configure.ac

uckelman:
Added --disable-multithreading option to configure.ac to permit the disabling of multithreading. This prevents TSK_MULTITHREAD_LIB from being defined when pthread is available.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -84,6 +84,13 @@ AX_PTHREAD([
     LDFLAGS="$LDFLAGS $PTHREAD_CFLAGS"
     CC="$PTHREAD_CC"],[])
 
+dnl Permit single-threaded builds
+AC_ARG_ENABLE([multithreading],
+    [AS_HELP_STRING([--disable-multithreading], [Build without multithreading support])])
+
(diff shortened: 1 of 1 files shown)


uckelman:
NB: There's a `#define TSK_MULTITHREAD_LIB` in tsk/base/tsk_os.h in a section guarded by `#ifdef _MSC_VER`. That means TSK_MULTITHREAD_LIB gets defined when compiling with Visual Studio regardless of the availability of pthreads, or anything else in configure.ac. This isn't a change I'm making---it was already like this---and I'm wondering if that's a bug, or at least if it should be made conditional on the multithreading flag my patch adds.

uckelman:
Aha, I see what's going on now---`TSK_MULTITHREAD_LIB` is set manually when compiling with Visual Studio because in that case you won't have a `tsk_incs.h` produced by running `configure`. So, nothing in `configure.ac` needs concern itself with Visual Studio. That satisfies me regarding my previous comment. Patch is good as-is.

PR #554, opened 2015-10-26
Fix memory leak in tsk_fs_path2inum

zrweger:


zrweger:
I noticed a call to tsk_fs_file_open() was leaking and isolated this offending function with valgrind. The fs_file allocated at line 259 isn't freed when the inum is found.

PR #556, opened 2015-11-02
Java home detection on OS X using the java_home tool.

RobertoMachorro:
*Java/JNI support* can now be found in OS X using the */usr/libexec/java_home* tool, if it isn't available (highly unlikely), then it falls back to prior behavior.


bcarrier:
I updated the m4 macro to the latest one and it then worked on my 10.11 system. Can you test if the develop branch works as is or is this patch still needed?  Thanks.

bcarrier:
Closing based on lack of response if the macro update was good enough.

PR #565, opened 2015-12-18
Added tsk_img_open_external for setting up handles for external image types

uckelman:
This patch adds tsk_img_open_external, which eases the setup of `TSK_IMG_INFO`s of type `TSK_IMG_TYPE_EXTERNAL` and---most importantly---ensures that the cache lock is initialized and deinitialized as appropriate. (It wasn't possible to handle the cache lock properly for external types before, due to the locking functions being internal to libtsk.)

There's also a one-line fix for two bad casts in tests/fs_attrlist_apis.cpp, which prevented compilation under MinGW.

### DIFF
--- a/tests/fs_attrlist_apis.cpp
+++ b/tests/fs_attrlist_apis.cpp
@@ -118,8 +118,8 @@ test_get_apis(TSK_FS_INFO * a_fs, TSK_INUM_T a_addr, int a_len)
 
         if (fs_attr != fs_attr2) {
             fprintf(stderr,
-                "Attribute from get_type not same addr as original %lu vs %lu from %"
-                PRIuINUM "\n", (unsigned long)fs_attr, (unsigned long)fs_attr2, a_addr);
+                "Attribute from get_type not same addr as original %p vs %p from %"
+                PRIuINUM "\n", fs_attr, fs_attr2, a_addr);
             tsk_error_print(stderr);
             return 1;
         }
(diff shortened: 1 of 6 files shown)


bcarrier:
Adding some comments here for my future reference.  Prior to this, tsk_img_malloc and tsk_img_free would initialize and de-initialize the cache lock.  Main change in here is that tsk_img_open now initializes it and tsk_img_close de-initializes it. 

Can you make these small changes:
- tsk_img_free() should have the line to de-initialize the cache lock removed because it is still being called from raw.c and such before the lock is initialized.  
- With the above change, then I think it it is cleanest if ewf_close and the like go back to their original code of calling tsk_img_free() instead of an explicit free and reset of tag.

uckelman:
You're still using `tsk_img_free` for cleanup on error in 23 places _after_ my changes, and in each of those places the lock needs to be released. If you remove the call to `tsk_deinit_lock` from `tsk_img_free`, then any time you have an error opening an image---say, in `raw_open`, where you have nine such calls---the lock will be wedged.

I hadn't noticed that `raw_open` was being called before the lock is taken. I'll fix that.

uckelman:
Now that I look closer, the lock is taken via the call to `tsk_img_malloc` near the top of `raw_open`. So `tsk_img_free` is _not_ being called before the lock is taken.

uckelman:
I added one more commit. This removes the call to `tsk_deinit_lock` from `tsk_img_free` and restores the call to `tsk_img_free` to `ewf_close` &c. as you requested, but also makes sure that `tsk_init_lock` is called only when `tsk_img_open is going to return nonnull.

PR #566, opened 2016-01-02
Tsk_DataModel_PostgreSQL.jar: not included in 'dist' (proposed fix)

dkarpo:
Hello.

When building Autopsy 4 against Sleuthkit, the Autopsy build errors that is cannot locate Tsk_DataModel_PostgreSQL.jar as follows:

<snip>
BUILD FAILED
/home/dk/sc/git-ext/autopsy/netbeans-plat/8.0.2/harness/suite.xml:187: The following error occurred while executing this line:
/home/dk/sc/git-ext/autopsy/Core/build.xml:36: The following error occurred while executing this line:
/home/dk/sc/git-ext/autopsy/Core/build.xml:25: Warning: Could not find file /home/dk/sc/git-ext/sleuthkit-dk/bindings/java/dist/Tsk_DataModel_PostgreSQL.jar to copy.
<snip>

The problem appears to be that 'check-build' in sleuthkit/bindings/java/build.xml checks for the targetfile Tsk_DataModel.jar (which sets 'up-to-date) and then carries on leaving only Tsk_DataModel.jar in the dist directory.  As far as 'dist-PostgreSQL' is concerned, the build is up-to-date so it doesn't need to run, therefore Tsk_DataModel_PostgreSQL.jar never makes it into dist.

  dk@anubis:~/sc/git-ext/sleuthkit-dk/bindings/java$ find dist/
  dist/
  dist/Tsk_DataModel.jar

I propose that 'check-build' be changed to check if Tsk_DataModel_PostgreSQL.jar is uptodate.  This ensure that both the Tsk_DataModel.jar and Tsk_DataModel_PostgreSQL.jar files make it into dist which makes Autopsy happy.  Here's what dist looks like after the change:

  dk@anubis:~/sc/git-ext/sleuthkit-dk/bindings/java$ find dist/
  dist/
  dist/Tsk_DataModel.jar
  dist/Tsk_DataModel_PostgreSQL.jar

There's probably a better way to do this (ie. using <mapper> with <uptodate> to check for multiple targets) but I'm no Ant expert and the proposed change seems to work fine.


dkarpo:
I don't want to merge my remote-tracking branches!  Argh!  Will fix and create a new pull request.

PR #567, opened 2016-01-02
Tsk_DataModel_PostgreSQL.jar: not included in 'dist' (proposed fix)

dkarpo:
Hello.

When building Autopsy 4 against Sleuthkit, the Autopsy build errors that is cannot locate Tsk_DataModel_PostgreSQL.jar as follows:

BUILD FAILED
/home/dk/sc/git-ext/autopsy/netbeans-plat/8.0.2/harness/suite.xml:187: The following error occurred while executing this line:
/home/dk/sc/git-ext/autopsy/Core/build.xml:36: The following error occurred while executing this line:
/home/dk/sc/git-ext/autopsy/Core/build.xml:25: Warning: Could not find file /home/dk/sc/git-ext/sleuthkit-dk/bindings/java/dist/Tsk_DataModel_PostgreSQL.jar to copy.

The problem appears to be that 'check-build' in sleuthkit/bindings/java/build.xml checks for the targetfile Tsk_DataModel.jar (which sets 'up-to-date) and then carries on leaving only Tsk_DataModel.jar in the dist directory.  As far as 'dist-PostgreSQL' is concerned, the build is up-to-date so it doesn't need to run, therefore Tsk_DataModel_PostgreSQL.jar never makes it into dist.

  dk@anubis:~/sc/git-ext/sleuthkit-dk/bindings/java$ find dist/
  dist/
  dist/Tsk_DataModel.jar

I propose that 'check-build' be changed to check if Tsk_DataModel_PostgreSQL.jar is uptodate.  This ensure that both the Tsk_DataModel.jar and Tsk_DataModel_PostgreSQL.jar files make it into dist which makes Autopsy happy. Here's what dist looks like after the change:

  dk@anubis:~/sc/git-ext/sleuthkit-dk/bindings/java$ find dist/
  dist/
  dist/Tsk_DataModel.jar
  dist/Tsk_DataModel_PostgreSQL.jar

There's probably a better way to do this (ie. using with to check for multiple targets) but I'm no Ant expert and the proposed change seems to work fine.

### DIFF
--- a/bindings/java/build.xml
+++ b/bindings/java/build.xml
@@ -1,14 +1,14 @@
 <project xmlns:ivy="antlib:org.apache.ivy.ant" name="DataModel" default="dist" basedir=".">
 	<description>
-		Sleuthkit Java DataModel 
+		Sleuthkit Java DataModel
 	</description>
-	<condition property="os.family" value="unix">  
-		<os family="unix"/>  
-	</condition>  
-	<condition property="os.family" value="windows">  
-		<os family="windows"/>  
-	</condition>  
-	<import file="build-${os.family}.xml"/> 
+	<condition property="os.family" value="unix">
+		<os family="unix"/>
+	</condition>
+	<condition property="os.family" value="windows">
+		<os family="windows"/>
+	</condition>
+	<import file="build-${os.family}.xml"/>
 
 	<!-- set global properties for this build -->
 	<property name="src" location="src/org/sleuthkit/datamodel"/>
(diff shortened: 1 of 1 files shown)


rcordovano:
After some debate, it was decided that developers who want to build SleuthKit and the SleuthKit Java bindings from source should not be required to build with the PostgreSQL libraries. The intent was to not raise the bar for building TSK, given that PostgreSQL support was added for the sake of supporting collaboration in Autopsy. Of course, this is a bit inconvenient for those building Autopsy from source, as you have discovered. The Autopsy developer must build either the Debug_PostgreSQL or Release_PostgreSQL targets for Sleuthkit and either the Debug-PostgreSQL or the dist-PostgreSQL target for the Java bindings. 

We have reviewed this decision and decided to not change it at this time, so although we appreciate this contribution (thanks!), we are not accepting it at this time.

PR #574, opened 2016-01-12
Custom artifact addition change revision

BasisOlivers:
Not much was really changed in the end, might not want properties change to stay, let me know.


rcordovano:
We'll hold off on this idea for now.

PR #575, opened 2016-01-13
Fixed sql queries for all getBlackboardArtifacts methods to work for …

BasisOlivers:
…BlackboardArtifact class, eliminated need for helper.


millmanorama:
this looks like a prime candidate to reduce code duplication by having  a more general (private)  method that these variations delegate to by passing appropriate arguments.

PR #599, opened 2016-02-11
Proper fix to TSK build error

eugene7646:
The intent of the original fix was to simply put the declaration of img_info at the top of tsk_img_open_external() method. I accidentally put the definition of img_info as one of input parameters into the tsk_img_open_external() method. This pull request fixes that error.


uckelman:
I spotted this. Fix looks good.

uckelman:
I spotted this. Fix looks good.

PR #608, opened 2016-02-21
Use GUID partition type as gpt partition description

shujianyang:
The mmls tool displayed partition information in an image. In previous edition, while processing GPT partitioned image, the "Description" column used the **name** field of a partition entry. But usually it did not provide very useful information.

Unlike GPT, while processing DOS partition, the mmls uses the flag in an entry to indicate the type of the partition and displays it in the "Description" column. This provides more accurate and useful information.

Thus, using similar method in GPT system may be a better idea. The program now reads partition type GUID, namely the first 16 bytes of each GPT partition entry, then finds out the corresponding partition type and displays it in mmls output.

### DIFF
--- a/tsk/vs/gpt.c
+++ b/tsk/vs/gpt.c
@@ -16,11 +16,19 @@
 #include "tsk_dos.h"
 
 
-/*
- *Check if GUID matches a given value.
+/**
+ * Check if GUID matches a given value.
+ *
+ * @param g The GUID to match.
+ * @param d1 First 4 bytes of the given value. 
+ * @param d2 Next 2 bytes of the given value following d1.
+ * @param d3 Next 2 bytes of the given value following d2.
+ * @param d4 Next 8 bytes of the given value following d3.
+ * @returns 1 if they match, 0 if they do not.
+ *
  */
 static int
-gpt_guid_match(GUID * g, uint32_t d1, uint16_t d2, uint16_t d3, uint64_t d4)
+gpt_guid_match(gpt_guid * g, uint32_t d1, uint16_t d2, uint16_t d3, uint64_t d4)
 {
     if(g->data_1 != d1 || g->data_2 != d2 || g->data_3 != d3)
         return 0;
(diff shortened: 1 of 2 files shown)


tsk/vs/tsk_gpt.h
```diff
@@ -28,6 +28,18 @@ extern "C" {
...
+/*Length of a GUID partition type description*/
+#define GUID_DESC_LEN 64
+
+/*Definiation of GUID struct.*/
+    typedef struct{
+        uint32_t data_1;
+        uint16_t data_2;
+        uint16_t data_3;
+        uint8_t data_4[8];
+    } GUID;
```
bcarrier: This should have a GPT-specific name.  To be consistent with other volume systems, something like gpt_guid.

tsk/vs/gpt.c
```diff
@@ -16,6 +16,73 @@
 #include "tsk_dos.h"
 
 
+/*
+ *Check if GUID matches a given value.
+ */
```
bcarrier: Can you add basic doxygen-style API docs.

/**
 * check if GUID matches a given value.
 * @param g GUID to match
* @param d1 ....
....
* @returns 1 if they match, 0 if they do not
*/

tsk/vs/gpt.c
```diff
@@ -16,6 +16,73 @@
...
+    else if(gpt_guid_match(g, 0x5808C8AA, 0x7E8F, 0x42E0, 0x85D2E1E90434CFB3))
+        snprintf(desc, GUID_DESC_LEN, "Logical Disk Manager metadata partition");
+    else if(gpt_guid_match(g, 0xAF9B60A0, 0x1431, 0x4F62, 0xBC683311714A69AD))
+        snprintf(desc, GUID_DESC_LEN, "Logical Disk Manager data partition");
+    else if(gpt_guid_match(g, 0x37AFFC90, 0xEF7D, 0x4E96, 0x91C32D7AE055B174))
+        snprintf(desc, GUID_DESC_LEN, "GPFS partition");
+    else if(gpt_guid_match(g, 0xE75CAF8F, 0xF680, 0x4CEE, 0xAFA3B001E56EFC2D))
+        snprintf(desc, GUID_DESC_LEN, "Storage Spaces partition");
+    else
+        snprintf(desc, GUID_DESC_LEN, "[Unkown type]");
```
bcarrier: Given that this list is not complete, I think the caller needs to know that the type was not matched so that it can then print the name.  On many systems, the name provides useful information, so I think that should be displayed if we can't map type.

bcarrier:
Thanks Shujianyang. I made some comments in the code.

shujianyang:
I made some improvements. Please check.

simsong:
Is this still relevant?

simsong:
Hi. I don't want to take this until it has a test case, given its age. Please rebase and add a test case.

tsk/vs/tsk_gpt.h
simsong: @bcarrier - in fact, it *is* a GPT-specific name. These are the names that Microsoft uses in its documentation!
https://learn.microsoft.com/en-us/windows/win32/api/guiddef/ns-guiddef-guid

simsong:
Moved to #3144

simsong:
`mmls` test is now failing because GPT is now identified with more info. Yay! The `mmls` test is working.
c.f. https://github.com/sleuthkit/sleuthkit/actions/runs/12493023090/job/34860986689?pr=3144

PR #628, opened 2016-03-22
Updated getMatchingAritfacts to use 1 db trip.

BasisOlivers:


rcordovano:
The SQL is incorrect, but we actually need to use a different approach to maintain backwards-compatibility of the API. We will do type info caching in the get type info method instead.

PR #649, opened 2016-05-12
Converted NTFS fs_attr_id from uint16 to int16

eugene7646:
Autopsy treats fs_id as SHORT. This code change adds ability to handle attribute ids that are larger than 32K by converting them to signed short values.

### DIFF
--- a/tsk/fs/fs_attr.c
+++ b/tsk/fs/fs_attr.c
@@ -275,7 +275,7 @@ tsk_fs_attr_set_str(TSK_FS_FILE * a_fs_file, TSK_FS_ATTR * a_fs_attr,
 uint8_t
 tsk_fs_attr_set_run(TSK_FS_FILE * a_fs_file, TSK_FS_ATTR * a_fs_attr,
     TSK_FS_ATTR_RUN * a_data_run_new, const char *name,
-    TSK_FS_ATTR_TYPE_ENUM type, uint16_t id, TSK_OFF_T size,
+    TSK_FS_ATTR_TYPE_ENUM type, int16_t id, TSK_OFF_T size,
     TSK_OFF_T init_size, TSK_OFF_T alloc_size,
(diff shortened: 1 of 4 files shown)


tsk/fs/fs_attr.c
```diff
@@ -275,7 +275,7 @@ tsk_fs_attr_set_str(TSK_FS_FILE * a_fs_file, TSK_FS_ATTR * a_fs_attr,
 uint8_t
 tsk_fs_attr_set_run(TSK_FS_FILE * a_fs_file, TSK_FS_ATTR * a_fs_attr,
     TSK_FS_ATTR_RUN * a_data_run_new, const char *name,
-    TSK_FS_ATTR_TYPE_ENUM type, uint16_t id, TSK_OFF_T size,
+    TSK_FS_ATTR_TYPE_ENUM type, int16_t id, TSK_OFF_T size,
```
karlmortensen: Does this actually help? The underlying bits should still be the same. Are we comparing it's size to 0 at some point?

I thought the issue was some things thought it was a 32-bit value and some thought it was 16.... Am I mis-remembering? :-)

eugene7646: No. The issue is that in C++ the field is stored as uint16 so it can hold values up to 64K but in Java it is stored as short so it can only store +/-32K. This may all be moot though. See story 1669 for latest comments made after this commit.

karlmortensen: Does this mean we'd only support 64k of attributes? While large, that seems potentially limiting...

eugene7646: Actually, according to Brian: "I don't think I've ever seen anything over 10".

karlmortensen: Is that per artifact, or total number of attributes for the case?

eugene7646: It's an id of the file system attribute, not blackboard attribute. Anyway, just look in the story for the latest ;)

PR #655, opened 2016-05-23
JNI Function for Bitlocker Support

mvetsch:
To allow autopsy the implementation of a Bitlocker-Module, we need to know where the volumes start. 

+ added getVolOffset JNI function

### DIFF
--- a/bindings/java/jni/dataModel_SleuthkitJNI.cpp
+++ b/bindings/java/jni/dataModel_SleuthkitJNI.cpp
@@ -1284,6 +1284,31 @@ JNIEXPORT jlong JNICALL Java_org_sleuthkit_datamodel_SleuthkitJNI_openFsNat
     return (jlong) fs_info;
 }
 
+/*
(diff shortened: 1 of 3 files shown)


bcarrier:
Can't you get the needed info from Volume.getStart().

mvetsch:
Volume.getStart() is available after the datasource was added into the sleuthkit case. 
The bitlocker feature needs to detect bitlocker encrypted volumes before the datasource is added to sleuthkit. We need a way to find the beginning of a volume before the information is added to the case database.

mecinformatica:
Would it be possible to implement BitLocker support? Specifically on the library used to decode the filesystem.

simsong:
Is this still relevant?

simsong:
Hi. Please rebase, resolve conflicts, and make sure it passes all check. Thanks.

simsong:
@jayaramcs do you care about this?

PR #659, opened 2016-05-31
Skip blank entries when removing vol_ and img_ from unique path names

karlmortensen:


rcordovano:
Please target branch 251 for this work.

PR #663, opened 2016-06-06
ffind: remove control caracters from output

noxdafox:
Files containing special characters result in corrupted output.

If a file contains control caracters such as newline in its name or
path, they will be replaced by the '^' character.

This reproduce the same behaviour of fls tool.

Signed-off-by: Matteo Cafasso <noxdafox@gmail.com>


noxdafox:
The pull request contains several whitespace fixed automatically by my editor. To ignore them, append the string "?w=1" to the diff URL.

PR #664, opened 2016-06-06
Handle JFIF .tec files as JPEG with a leading EOI marker

karlmortensen:


karlmortensen:
Submitting to develop instead.

PR #665, opened 2016-06-06
Allow offset into Content

karlmortensen:


bindings/java/src/org/sleuthkit/datamodel/ReadContentInputStream.java
```diff
@@ -39,6 +39,18 @@ public ReadContentInputStream(Content content) {
...
+	 *
+	 * @param content The Content object to read from
+	 * @param offset  The initial offset to start reading from
+	 */
+	public ReadContentInputStream(Content content, int offset) {
+		this.content = content;
+		this.currentOffset = offset;
+		this.contentSize = content.getSize();
+	}
+
```
rcordovano: Could we satisfy the use case using the seek method instead of adding a new constructor?

karlmortensen: Yes. Even better. I'll do that.

karlmortensen:
Using seek() instead of a new constructor.

PR #669, opened 2016-06-11
Replace custom path printing with tsk_print_sanitized

noxdafox:


tsk/fs/fs_name.c
```diff
@@ -213,11 +213,12 @@ tsk_fs_name_copy(TSK_FS_NAME * a_fs_name_to,
...
- * Print the path removing control characters.
+ * \ingroup fslib
+ * Prints the path removing control characters.
  *
- * Return 0 on success and 1 on error.
+ * Returns 0 on success and 1 on error.
  */
-static int
+int
 tsk_print_path(FILE * hFile, const char *path)
```
bcarrier: If this now becomes non-static and is in the fs code, then we should rename it to something like 'tsk_fs_print_path'.  Or we could put it into 'base' with a name such as tsk_print_escaped().  I think I like the latter better so that it is not just for file system code.

noxdafox: Makes perfect sense, I'll update the PR.

noxdafox:
Pull request updated.

The pull request contains few whitespace fixed automatically by my editor. To ignore them, append the string "?w=1" to the diff URL.

bcarrier:
This was merged into the release-4.3.0 branch. 1e81b23..105cf42 

Thanks

PR #681, opened 2016-06-16
Fixed memory leaks

uckelman-sf:
* Moved initialization of TSK_FS_INFO* in yaffs2_open to occur immediately after the tsk_fs_malloc, in order that fs is never null when "goto on_error" happens. This ensures that yaffsfs_close(fs) frees the memory rather than leaking a YAFFSFS_INFO every time a non-YAFFS file system is autodetected.
* fiwalk: md and mdctx were leaked in the case where the hash is destroyed without being initialized.
*  Fixed many C++11 "invalid suffix on literal" warnings.
* Fixed leak of temporary buffer.
* Fixed leak of TSK_FS_ATTR_RUNs on error in fatfs_make_data_runs.


uckelman-sf:
Fixed three more memory leaks, this time in hfs.c.

tools/fiwalk/src/hash_t.h
```diff
@@ -245,12 +245,12 @@ class hash_generator__:T { 			/* generates the hash */
 	hashed_bytes += bufsize;
     }
     void release(){			/* free allocated memory */
-	if(initialized){
```
bcarrier: @simsong. Comments on if this removal is correct? I don't know this code.

bcarrier:
Thanks Joel.  I'm good with the core C fixes, but would like an OK from a fiwalk person.

tools/fiwalk/src/hash_t.h
simsong: Yes, it's correct. The `md` and `mdctx` get allocated whether the object is initialized or not. The problem came from my thinking that initialized referred both to the object and to the hash machine.

PR #684, opened 2016-06-17
Fix NTFS attrlist bounds checks, memory leaks

zweger:
- Fixes bounds checks in `ntfs_proc_attrlist()`. Note that `ntfs_attrlist` contains the first byte of its name, but that the name's length is possibly 0. In addition to checking that the attribute struct itself fits inside the current buffer, check the attribute's name length and and declared size.
- Fixes two memory leaks in `tsk_fs_dir_find_orphans()`. if the fs_dir's name's used is decremented, the names must be freed, since `tsk_fs_dir_close()` won't know to free them. Also frees the `orphan_subdir_list` in certain error conditions.

### DIFF
--- a/tsk/fs/fs_dir.c
+++ b/tsk/fs/fs_dir.c
@@ -1220,6 +1220,15 @@ tsk_fs_dir_find_orphans(TSK_FS_INFO * a_fs, TSK_FS_DIR * a_fs_dir)
                     &a_fs_dir->names[a_fs_dir->names_used - 1]);
             }
             a_fs_dir->names_used--;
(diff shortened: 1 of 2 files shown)


bcarrier:
Hi Zweger,

I was curious if you had data that was causing problems that this fixes or if this was proactive bounds checking?  

thanks,
brian

zweger:
I do. I noticed valgrind complaining about some reads inside the loop, which this change prevents. Unfortunately I'm unable to share that data.

zweger:
@bcarrier ping.

I've updated this PR to include to fix two unrelated memory leaks. Also, I realized my initial fix of the `ntfs_proc_attrlist()` bounds checks was not correct - this is now fixed.

PR #688, opened 2016-07-14
add artifact and attribtes

millmanorama:
added TSK_ACCOUNT artifact and credit card attributes


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -218,7 +218,9 @@ public int hashCode() {
 		TSK_REMOTE_DRIVE(37, "TSK_REMOTE_DRIVE", //NON-NLS
 				bundle.getString("BlackboardArtifact.tskRemoteDrive.text")),
 		TSK_FACE_DETECTED(38, "TSK_FACE_DETECTED", //NON-NLS
-				bundle.getString("BlackboardArtifact.tskFaceDetected.text"));
+				bundle.getString("BlackboardArtifact.tskFaceDetected.text")),
+		TSK_ACCOUNT(39, "TSK_ACCOUNT", //NON-NLS
```
noxdafox: This would probably collide with what intended to be changed here: https://github.com/sleuthkit/sleuthkit/blob/develop/API-CHANGES.txt#L2

millmanorama: you are right.  I didn't know about that plan...

noxdafox: not sure if still valid tho

rcordovano: API-CHANGES.txt is definitely out of date (last updated in 2013). We have plans to change the way accounts are handled, so we will evaluate the assertion about the deprecation of TSK_SERVICES_ACCOUNT as the design work for Autopsy proceeds. I am adding updating API-CHANGES.txt to our TODO list.

PR #689, opened 2016-07-17
Feature: NTFS Update Sequence Number support

noxdafox:
This pull request adds support to TSK for parsing NTFS Update Sequence Number Journals.

More information about it here: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365722%28v=vs.85%29.aspx

The feature was discussed in the mailing list: https://sourceforge.net/p/sleuthkit/mailman/message/35188151/

The USN is a unique feature of NTFS File System which allows to build pretty detailed timeline of changes affecting a FS.

Brian Carrier suggested to implement this feature within TSK Journal infrastructure. Even if I used very similar structures and naming, there are few reasons why I didn't follow such path.
1. TSK jls command (and implementation) refers to generic FS journaling. Even if implementations are different, all of them share the same end goal (avoid blocks corruption and recover from power loss). 
2. Most of the FS journaling implementations share the same level of abstraction (blocks).  
3. The Update Sequence Number Journal is NTFS specific, no other FS implement such feature.
4. The USN journal goal is helping FS backup and restore applications which is fairly different from a FS journal.
5. The USN journal refers to a differet level of abstraction. Rather than FS blocks, it keeps information of FS changes such as metadata changes and file content changes.

For these reasons, I thought a separate namespace would be less confusing for the end user.

Therefore a new command line tool has been added: usnjls. The tool checks whether the disk under analysis is a NTFS disk and contains a valid journal file.

At the moment, it reports the raw content of the journal records and a detailed representation of it. 
Example of current output on a Windows 7 disk.
http://pastebin.com/YpfEbSCS
Long format.
http://pastebin.com/RcgrTDNP
Mactime format
http://pastebin.com/2Tf7XVMW

The APIs are accessible as they might be useful for developers willing to include USN journal information within their timeline creation tools.

API documentation is missing, I coulnd't find its location within the code.

PS: several line changes are whitespaces removed automatically by my editor. To ignore them, append the string "?w=1" to the diff URL.


noxdafox:
Further steps if/once this set of changes gets merged.

- Add support for Version 3 records (need to find examples): https://msdn.microsoft.com/en-us/library/windows/desktop/hh802708%28v=vs.85%29.aspx
- Add support for Version 4 records (need to find examples): https://msdn.microsoft.com/en-us/library/windows/desktop/mt684964%28v=vs.85%29.aspx
- Import changes definitions and add fields unpacking.
- Add "pretty print" flag to usnjls tool unpacking fields to readable flags. Example:
http://pastebin.com/qN2JiJm9

noxdafox:
Added constant mapping from MSDN documentation.

Added long information printing format: http://pastebin.com/wsGRxGAu

noxdafox:
Added mactime report format: http://pastebin.com/2Tf7XVMW

noxdafox:
@bcarrier any feedback on this PR?

noxdafox:
Any comment regarding this feature?

bcarrier:
To give a quick update on this, we did an internal review but got side tracked before we completed it. I'll make some time next week to make sure we get this out. There were some code suggestions and output format questions we had in terms of making it most usable.

noxdafox:
Indeed that was more than expected, I'll wait for your comments.

APriestman:
Thanks for waiting!

Here are some specific comments:
* In the usage statement, inode should be in brackets since it is optional
* There should be better error messages for non-NTFS systems. Something about USN journals being an NTFS thing seems more informative than  printing "Unable to open USN Journal file $Extend/$UsnJrnl"
* usnjls.cpp line 191 free(jrnl_file) should be tsk_fs_file_close(jrnl_file)
* usn_journal.c line 313 free(ntfs->usnjinfo) needs to also close tsk_fs_file_close on ntfs->usnjinfo->fs_file
* usn_journal.c line 248 This appears to be wrong: ntfs->usnjinfo = tsk_malloc(sizeof ntfs->usnjinfo); That gives a size of 8 but the struct contains a TSK_FS_FILE, TSK_INUM_T, an a uint32_t
* usn_journal.c parse_fname the UTF16 to UTF8 conversion is wrong - uses the length (in characters) of the UTF16 string as the length of the UTF8 string (which I believe should be in bytes). May also need to put a null at the end of the new string.
* usn_journal.c could generally use better comment blocks, particularly on what the return values are.
* usnjls_lib.c tsk_fs_usnjls seems like it would be nicer if there was a test for ntfs before calling tsk_ntfs_usnjopen (that method does do the check). If we ever had similar functionality for another fs then we could add it there.

We'd also like to see the more NTFS-specific enums and related code moved into ntfs.h, where possible. Feel free to ask about this.

The final thing is that we'd like the default output to be a bit more user friendly. The idea being that someone looking at it for the first time would have a reasonable guess about what the fields are, and we would only display the ones that seem the most interesting (we believe these would be name, inum, timestamp, and reason).

noxdafox:
Thanks for the comments, already started fixing the simple ones.

Few questions about the last points.

> We'd also like to see the more NTFS-specific enums and related code moved into ntfs.h, where possible. Feel free to ask about this.

Could you please point out in the code what shall be moved? 
I believe I implemented it so because I wanted to make sure the APIs were available for developers willing to integrate TSK in their software. 

> The final thing is that we'd like the default output to be a bit more user friendly. The idea being that someone looking at it for the first time would have a reasonable guess about what the fields are, and we would only display the ones that seem the most interesting (we believe these would be name, inum, timestamp, and reason).

Shall the reason be decoded like the long format or is it OK to leave it encoded? I would also keep the parent inum as it's very useful to track directory tree deletes as you can simply traverse the results back in time. This is something where `fls` usually falls short in case some inode (file ref number) in the directory tree is recycled by the OS. 

What about something like:
```
inode parent_inode timestamp decoded_reason name

59387-1 59379-1 1454616080.339843700 DATA_EXTEND,FILE_CREATE readme.txt
```

noxdafox:
Added first set of fixes according to the review comments.

I reworked the default output format as suggested using tabs instead of spaces for fields separation to comply with other tools.

Example:
https://pastebin.com/W2svJuMj

Only remaining point is the move of the enums to the ntfs header file for which I'm waiting for further info.

noxdafox:
@APriestman I guess I understood what you meant.

I moved the USN specific enums to the NTFS header file. The User will still have access to the `tsk_ntfs_usn*` APIs.

I believe the set of requested fixes is complete.

Please let me know if there are other issues with the PR.

bcarrier:
Hi @noxdafox, sorry for the delay on this.  I've been meaning to provide more guidance on this, but keep on forgetting to add it.  My thoughts on this are that:
- Historically, the TSK APIs have been file system-type agnostic.  
- The current API passes a data structure in the callback that the callback must parse and know the structure of.  I'd rather that more of the parsing was done outside of the callback and the callback just had a new structure with the parsed values.   that may take some time to figure out what the callback structure shoudl be.
- I was going to suggest that your changes could be placed in tsk_fs_i.h instead of tsk_fs.h.  tsk_fs_i.h is not part of the "public API" so we can change it more easily in the future. 

Can you do that so that there are no tsk_fs.h changes?  This can then change as you get user feedback.  We'll then get this into the next release (which is at the end of this month).

noxdafox:
@bcarrier I agree with the idea of moving the APIs to tsk_fs_i.h. I'll do it during this WE.

Regarding the second point, what you ask might be a bit tricky. Microsoft developers love to pack information in structure fields to save space. If dealing with bitfield manipulation is fairly easy, the same cannot be said when it comes to "expanding" them.

What shall the field `reason`, `source_info` and `attribute` expand into? 
Should they be arrays of constant values (example `TSK_FS_USN_REASON_FILE_CREATE`)? 
Shall we terminate them with NULL values or shall we add the array size within the structure?
Maybe arrays of strings? I personally dislike this one.

Another question I had in these regards is the maintenance cost of defining our own constants. Unpacking the values at the API level means we are promising the Users we'll maintain the table of references ourselves.

With the current solution the users of the `usnjls` tool would notice an `UNKNOWN` field but the API users would still get the full information.

bcarrier:
Those are all great questions about what to pass in, which is why I'd rather not make it the public API. 

I'd rather not introduce tsk_fs_i.h into tsk_tools_i.h. The tool you created is unique in that it needs access to structures in tsk_fs_i.h, so I'd rather that you directly include that file from the new tool cpp file.

noxdafox:
Made as requested: usnjls tool includes tsk_fs_i.h directly.

I have the following proposal: we merge these changes before the PR gets too messy so we can release the `usnjls` tool (as long as there are no further changes to be made).

Then we can figure out how we can bring FS specific APIs to the user in the mailing list. I have some ideas in merit.

bcarrier:
Great.  I just merged this into 4.4.2 branch. Thanks!

PR #699, opened 2016-08-11
831 status column

millmanorama:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -768,7 +799,51 @@ private int updateFromSchema3toSchema4(int schemaVersionNumber, CaseDbConnection
...
+		if (schemaVersionNumber != 4) {
+			return schemaVersionNumber;
+		}
+
+		Statement statement = null;
+		try {
+			statement = connection.createStatement();
+			// Add the review_statuses lookup table.
+			if (this.dbType.equals(DbType.SQLITE)) {
+				statement.execute("CREATE TABLE review_statuses (review_status_id INTEGER PRIMARY KEY, review_status_name TEXT NOT NULL, display_name TEXT NOT NULL)");
```
millmanorama: If this can be INTEGER in both PostgreSQL and SQLite than we don't need this if.

tsk/auto/db_postgresql.cpp
```diff
@@ -544,8 +544,18 @@ int TskDbPostgreSQL::initialize() {
         ||
         attempt_exec("CREATE TABLE blackboard_attribute_types (attribute_type_id BIGSERIAL PRIMARY KEY, type_name TEXT NOT NULL, display_name TEXT, value_type INTEGER NOT NULL)","Error creating blackboard_attribute_types table: %s\n")
         ||
-        attempt_exec("CREATE TABLE blackboard_artifacts (artifact_id BIGSERIAL PRIMARY KEY, obj_id BIGINT NOT NULL, artifact_type_id BIGINT NOT NULL, "
-        "FOREIGN KEY(obj_id) REFERENCES tsk_objects(obj_id), FOREIGN KEY(artifact_type_id) REFERENCES blackboard_artifact_types(artifact_type_id))",
+		attempt_exec("CREATE TABLE review_statuses (review_status_id BIGSERIAL PRIMARY KEY, "
```
millmanorama: Does it matter if the id column is BIGSERIAL or INTEGER?  see comment above

tsk/auto/db_postgresql.cpp
```diff
@@ -544,8 +544,18 @@ int TskDbPostgreSQL::initialize() {
...
+		"display_name TEXT NOT NULL)",
+		"Error creating review_statuses table: %s\n")
+        ||
+        attempt_exec("CREATE TABLE blackboard_artifacts (artifact_id BIGSERIAL PRIMARY KEY, "
+		"obj_id BIGINT NOT NULL, "
+		"artifact_type_id BIGINT NOT NULL, "
+		"review_status_id INTEGER NOT NULL, "
+        "FOREIGN KEY(obj_id) REFERENCES tsk_objects(obj_id), "
+		"FOREIGN KEY(artifact_type_id) REFERENCES blackboard_artifact_types(artifact_type_id)), "
+		"FOREIGN KEY(review_status_id) REFERENCES review_statuses(review_status_id)",
```
millmanorama: It would be nice to put the default value here, but we don't know what it is because the review statuses are defined and  this lookup table is populated only in the java bindings (SleuthkitCase.

tsk/auto/db_sqlite.cpp
```diff
@@ -303,9 +303,19 @@ int
...
         attempt_exec
-        ("CREATE TABLE blackboard_artifacts (artifact_id INTEGER PRIMARY KEY, obj_id INTEGER NOT NULL, artifact_type_id INTEGER NOT NULL, "
-        "FOREIGN KEY(obj_id) REFERENCES tsk_objects(obj_id), FOREIGN KEY(artifact_type_id) REFERENCES blackboard_artifact_types(artifact_type_id))",
+        ("CREATE TABLE blackboard_artifacts (artifact_id INTEGER PRIMARY KEY, "
+		"obj_id INTEGER NOT NULL, "
+		"artifact_type_id INTEGER NOT NULL, "
+		"review_status_id INTEGER NOT NULL, "
+        "FOREIGN KEY(obj_id) REFERENCES tsk_objects(obj_id), "
+		"FOREIGN KEY(artifact_type_id) REFERENCES blackboard_artifact_types(artifact_type_id), "
+		"FOREIGN KEY(review_status_id) REFERENCES review_statuses(review_status_id))",
```
millmanorama: It would be nice to put the default value here, but we don't know what it is because the review statuses are defined and  this lookup table is populated only in the java bindings (SleuthkitCase.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -768,7 +799,51 @@ private int updateFromSchema3toSchema4(int schemaVersionNumber, CaseDbConnection
...
+			} else {
+				statement.execute("CREATE TABLE review_statuses (review_status_id BIGSERIAL PRIMARY KEY, review_status_name TEXT NOT NULL, display_name TEXT NOT NULL)");
+			}
+
```
millmanorama: If the id column can be INTEGER in both PostgreSQL and SQLite than we don't need this if.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -6246,10 +6321,10 @@ public String getContextString() {
...
 				+ "AND tsk_files.type = ? )"), //NON-NLS
 		SELECT_FILE_BY_ID("SELECT * FROM tsk_files WHERE obj_id = ? LIMIT 1"), //NON-NLS
-		INSERT_ARTIFACT("INSERT INTO blackboard_artifacts (artifact_id, obj_id, artifact_type_id) " //NON-NLS
-				+ "VALUES (?, ?, ?)"), //NON-NLS
-		POSTGRESQL_INSERT_ARTIFACT("INSERT INTO blackboard_artifacts (artifact_id, obj_id, artifact_type_id) " //NON-NLS
-				+ "VALUES (DEFAULT, ?, ?)"), //NON-NLS
+		INSERT_ARTIFACT("INSERT INTO blackboard_artifacts (artifact_id, obj_id, artifact_type_id, review_status_id) " //NON-NLS
+				+ "VALUES (?, ?, ?," + ReviewStatus.UNDECIDED.getID() + ")"), //NON-NLS
+		POSTGRESQL_INSERT_ARTIFACT("INSERT INTO blackboard_artifacts (artifact_id, obj_id, artifact_type_id, review_status_id) " //NON-NLS
+				+ "VALUES (DEFAULT, ?, ?," + ReviewStatus.UNDECIDED.getID() + ")"), //NON-NLS
```
millmanorama: if we can define the default value in the  c++ code then we don't need it here.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -768,7 +799,51 @@ private int updateFromSchema3toSchema4(int schemaVersionNumber, CaseDbConnection
...
+
+			/*
+			 * Add review_status_id column to artifacts table.
+			 *
+			 * NOTE: For DBs created with schema 5 we define a foreign key
+			 * constraint on the review_status_column. We don't bother with this
+			 * for DBs updated to schema 5 because of limitations of the SQLite
+			 * ALTER TABLE command.
+			 */
+			statement.execute("ALTER TABLE blackboard_artifacts ADD COLUMN review_status_id INTEGER NOT NULL DEFAULT " + ReviewStatus.UNDECIDED.getID());
```
millmanorama: here we know the default value cause we are in JAVA land.

bindings/java/src/org/sleuthkit/datamodel/ReviewStatus.java
```diff
@@ -0,0 +1,68 @@
...
+		this.name = name;
+		this.displayName = ResourceBundle.getBundle("org.sleuthkit.datamodel.Bundle").getString(displayNameKey);
+	}
+
+	/**
+	 * Get the ID of this review status.
+	 *
+	 * @return the ID of this review status.
+	 */
+	public Integer getID() {
```
rcordovano: Why do we need this method instead of just calling Enum.ordinal? Is the idea that it is more clear this way?

bindings/java/src/org/sleuthkit/datamodel/ReviewStatus.java
```diff
@@ -0,0 +1,68 @@
...
+		return id;
+	}
+
+	/**
+	 * Get the name of this review status.
+	 *
+	 * @return the name of this review status.
+	 */
+	String getName() {
```
rcordovano: Why do we need this method instead of just calling Enum.name? Is the idea that it is more clear this way?

bindings/java/src/org/sleuthkit/datamodel/ReviewStatus.java
millmanorama: I am not sure.  This is what we do for other enums that map to values in the database, ie artifact and attribute types.

I thought we did it to insulate ourselves from accidental changes due to the fact that both ordinal and name are dependent on specifics of the source code .

bindings/java/src/org/sleuthkit/datamodel/ReviewStatus.java
millmanorama: see above

bindings/java/src/org/sleuthkit/datamodel/ReviewStatus.java
rcordovano: Good point, let's go with it as written.

bindings/java/src/org/sleuthkit/datamodel/ReviewStatus.java
rcordovano: Good point, let's go with it as written.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
rcordovano: I think it can be INTEGER in both places, since we are supplying it instead of doing an auto-increment.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
rcordovano: It sounds like it is worth a try.

tsk/auto/db_postgresql.cpp
rcordovano: See response above.

tsk/auto/db_postgresql.cpp
rcordovano: This actually begs the question, why don't we populate these enum tables in the C++ code? That seems like the way to go!

tsk/auto/db_sqlite.cpp
rcordovano: See above.

millmanorama:
I updated this with an INTEGER primary key instead of the BIGSERIAL key.  I will look into populating the table in C++, but I will make any changes in a new pull request.

PR #701, opened 2016-08-19
Apply MFT update sequence entries at the end of every 512 bytes…

### LINKED ISSUE #700: NTFS parsing doesn't work properly when the disk sector size is > 512
MFT Update Sequence Entries are currently being applied at the end of every sector instead of at the end of every 512 bytes.  Because of this, NTFS parsing fails for disks with sector sizes that are greater than 512.

The symptoms of this problem are exhibited as an error of `Cannot determine file system type` in `fls` if you don't pass the `-f ntfs` flag and `Error in metadata structure (dinode_lookup: More Update Sequence Entries than MFT size)` if you do.

jtsylve:
…instead of every sector. This fixes #700.

### DIFF
--- a/tsk/fs/ntfs.c
+++ b/tsk/fs/ntfs.c
@@ -358,7 +358,7 @@ ntfs_dinode_lookup(NTFS_INFO * a_ntfs, char *a_buf, TSK_INUM_T a_mftnum)
     mft = (ntfs_mft *) a_buf;
     if ((tsk_getu16(fs->endian, mft->upd_cnt) > 0) &&
         (((uint32_t) (tsk_getu16(fs->endian,
-                        mft->upd_cnt) - 1) * 512) >
+                        mft->upd_cnt) - 1) * NTFS_UPDATE_SEQ_STRIDE) >
             a_ntfs->mft_rsize_b)) {
         tsk_error_reset();
         tsk_error_set_errno(TSK_ERR_FS_INODE_COR);
@@ -384,7 +384,7 @@ ntfs_dinode_lookup(NTFS_INFO * a_ntfs, char *a_buf, TSK_INUM_T a_mftnum)
     for (i = 1; i < tsk_getu16(fs->endian, mft->upd_cnt); i++) {
         uint8_t *new_val, *old_val;
         /* The offset into the buffer of the value to analyze */
-        size_t offset = i * 512 - 2;
+        size_t offset = i * NTFS_UPDATE_SEQ_STRIDE - 2;
         /* get the current sequence value */
         uint16_t cur_seq =
             tsk_getu16(fs->endian, (uintptr_t) a_buf + offset);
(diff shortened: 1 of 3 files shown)


noxdafox:
I'd rather use a constant instead of hardcoded values.

jtsylve:
I agree.  Is one already defined?  Suggestions on constant name if not?  I'm not sure of the conventions you guys use.

jtsylve:
Done.  I defined it as NTFS_UPDATE_SEQ_STRIDE as that's the language Microsoft uses in their (seemingly incorrect) documentation of the header.  https://msdn.microsoft.com/en-us/library/bb470212%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

lfcnassif:
Hi,

I've tried the patch above with a 4k sector ntfs image but I am getting the following error:
Error: Error in metadata structure (Error: Index list offsets are invalid on entry: 5) (Error walking directory in file system at offset 8388608)
Any ideas?

Thanks,
Luis

jtsylve:
@lfcnassif That seems like an unrelated issue to me, but without the image I wouldn't know how to dig into it.  Obviously there needs to be a better effort of testing 4KiB sector disks.

lfcnassif:
Thanks jtsylve, I will test again when I have another 4k sector disk, maybe that is an unrelated issue.

jtsylve:
@bcarrier Any feedback on this PR?

bcarrier:
Let me see if I can fire up a 4K ntfs drive as a quick sanity check.  Can you add a 1-liner comment above the #define about what the STRIDE value is used for?  Something like "number of bytes that update sequences are applied to"

bcarrier:
I'm seeing similar things to @lfcnassif.  I have a 1TB external drive that uses 4k sectors.

Before your patch, I can't do anything.  It can't detect a file system for fsstat.  

After your patch, fsstat works.  Great!

But, then I try fls and get the same error as above. Problem hits here:

```
/* Verify the offset pointers */
    if ((tsk_getu32(a_fs->endian, idxelist->seqend_off) <
            tsk_getu32(a_fs->endian, idxelist->begin_off)) ||
        (tsk_getu32(a_fs->endian, idxelist->bufend_off) <
            tsk_getu32(a_fs->endian, idxelist->seqend_off)) ||
        (((uintptr_t) idxe + tsk_getu32(a_fs->endian,
                    idxelist->bufend_off)) >
            ((uintptr_t) fs_attr_root->rd.buf +
                fs_attr_root->rd.buf_size))) {
        tsk_error_reset();
        tsk_error_set_errno(TSK_ERR_FS_INODE_COR);
        tsk_error_set_errstr
            ("Error: Index list offsets are invalid on entry: %" PRIuINUM,
            fs_dir->fs_file->meta->addr);
        return TSK_COR;
    }
```
All of the offsets are way too big (74k for a 4k buffer), so the last check fails. But, relative to each other, the ending location and end of buffer are OK (they are equal). 

The index root structure is also a bit strange.  It says the index alloc size in bytes is 5632 bytes, but that it is also one cluster (which is 4K on this drive). 

@jtsylve, do you have some time to look at an image if I send it to you?  It should compress well .

jtsylve:
I sure can. If it's too big you can snail mail me the drive. Just hit me
offline for an address if it comes to that.

bcarrier:
Found the issue.  It's irrelevant to 4k clusters and has been there for YEARS.  I'll accept this pull request and submit one with just my fix.  Thanks!

lfcnassif:
Hi, just to let you guys know, applying Brian's fix solved the issue with my 4k sector image too. Thank you all.

PR #710, opened 2016-09-12
Removed and moved ResultSetHelper methods

wishdasher:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -2722,6 +2719,7 @@ int getContentChildrenCount(Content content) throws TskCoreException {
 		}
 	}
 
+	//filechildren method
```
rcordovano: Looks like a stray comment here.

PR #716, opened 2016-09-29
Use system sqlite3

kwizart:
This patch use system sqlite3 instead of the bundled copy.


uckelman-sf:
This PR could be closed, since it's subsumed by PR #860, which has already been merged.

uckelman-sf:
This PR could be closed, since it's subsumed in PR #860, which has already been merged.

PR #763, opened 2017-02-08
Remove escaping of strings from prepared statement

hokiemike84:
I haven't tested this, but the change seems simple enough.
It shouldn't be necessary to escape the string when using a PreparedStatement.  PreparedStatments handle escaping internally.  The extra escaping causes any single quotes to be saved as two single quotes in the database, an issue that I've observed.


rcordovano:
hokiemike84, this is a good idea, thanks! This PR does not auto-merge, but we have tested the fix and brought it in through a PR from an Autopsy dev team member, so I am going to close this PR. Again, thanks!

PR #770, opened 2017-02-14
Use system sqlite3 library

uckelman-sf:
The `#ifndef` guards in sqlite3.h changed from `_SQLITE3_H_` to `SQLITE3_H` in commit 5471aca015 to sqlite, and this change first appears in release 3.14, from August 2016. This causes a problem for any program which compiles against both libtsk and the system-installed sqlite3, as the system sqlite3.h header from 3.14 onward is no longer compatible with the sqlite3.h header bundled with sleuthkit.

This PR modifies sleuthkit to use the system sqlite3.


bcarrier:
Does this patch work for the Windows build?  How does it link with a locally installed SQLite? 

So, there seem to be 3 options with the general problem:
1) Upgrade the SQLite that TSK includes so that the correct guard names are used.  This still leaves the challenge that someone may end up with the TSK version of SQLite instead of the system one though.
2) Approach 1 plus only compile the TSK version if there is not a system one.  
3) Not include SQLite in TSK and force the user to have it installed. 

Of the three, my preference is for approach 2. It just needs a bit more autoconf/automake magic to decide to compile the SQLite code or link to it.

uckelman-sf:
> Does this patch work for the Windows build? How does it link with a locally installed SQLite?

For MinGW builds, my patch works the same as for native Linux builds. For Visual Studio, you would specify SQLite as a dependency the same way you specify your other dependencies (e.g., zlib, afflib, libewf).

Are we agreed that Option 1 is a bad option?

There are two problems I see with Option 2:

* How would Option 2 work for Visual Studio builds? There's no autoconf scripting there for selecting which of the bundled or installed lib to use.

* Consider this program:
```c++
#include <tsk/libtsk.h>
#include <iostream>

int main(int argc, char** argv) {
  std::cout << sqlite3_version << std::endl;
  return 0;
}
```
`sqlite3_version` is a string constant from `sqlite3.h`. You might initially object that it won't compile, because we haven't included `sqlite3.h`. But it _does_ compile. This is because `sqlite3.h` is included transitively by `tsk/libtsk.h`---i.e., SQLite3's API is exposed publicly through Sleuthkit's API. This means that anything which compiles against libtsk directly is seeing SQLite3's headers and anything which links against libtsk is seeing libsqlite3's symbols. So long as SQLite3 is exposed publicly by Sleuthkit and has been built against a bundled SQLite3, there is the risk that anyone building software with Sleuthkit will fail to compile due to version skew or fail to link due to symbol collisions (or possibly crash unpredictably, due to the dynamic linker picking the wrong symbol when they're multiply defined).

Option 2 gives the wary a workaround, but preserves the problem for the unwary. Option 3 fixes the problem, and does so in the same way which Sleuthkit already does for other dependent libraries.

bcarrier:
But the build targets in Visual Studio already include libewf and zlib as dependencies.  This PR doesn't include that, so don't the Visual Studio builds now fail or at least don't have the full tsk_loaddb functionality? 

In your example, sqlite3.h should be included via tsk/libtsk.h only if the embedded version was used.

uckelman-sf:
> But the build targets in Visual Studio already include libewf and zlib as dependencies. This PR doesn't include that, so don't the Visual Studio builds now fail or at least don't have the full tsk_loaddb functionality?

Yes, I believe you'd need to update those in Visual Studio. I could try to hand-edit `win32/libtsk/libtsk.vcxproj` but I'd only be guessing as to how to do it. I wouldn't want me to do hand edit those files, were I you.

> In your example, sqlite3.h should be included via tsk/libtsk.h only if the embedded version was used.

The current behavior is that `sqlite3.h` is included by `hashdb/tsk_hashdb.h` which is included by `tsk/libtsk.h`. My patch doesn't change that, and if my patch _did_ change that without changing anything else, it the patched code wouldn't compile. If you want SQLite3 to become a _private_ dependency of Sleuthkit instead of a _public_ dependency---which would mean _not_ including `sqlite3.h` via `tsk/libtsk.h`---then ` TSK_SQLITE_HDB_INFO` and `#include <sqlite3.h>` would need to be moved out of `hashdb/tsk_hashdb.h`. Given that `TSK_SQLITE_HDB_INFO` isn't used outside of `tsk/hashdb/sqlite_hdb.cpp`, I'd strongly recommend moving it there.

uckelman-sf:
I've created PR #848 with a different set of commits which allows the bundled sqlite3 to be used when a system one isn't available.

PR #774, opened 2017-02-23
1883 don't double escape string values of attributes in the DB.

millmanorama:


millmanorama:
this is really the same as PR #763 except it merges.

PR #793, opened 2017-03-25
Added command line option to invoke fls flag TSK_FS_FLS_HASH

amm3:
It seems strange that this flag exists and is supported by all the code behind `tsk_fs_fls()`, yet is not exposed in the fls command line.  In particular because the mactime/bodyfile format has a field for MD5 hash (position 0).

Ref: https://wiki.sleuthkit.org/index.php?title=Body_file

I'm not tied to a particular flag.  -m was taken and -h was not, so I just went with it, although I'd recognize -h is widely use for help/usage statements.


amm3:
Realizing this branch is probably closed... will resubmit under 4.4.1.

PR #796, opened 2017-03-30
Fixed SQL escaping issue for addSource feature

hokiemike84:
I was getting SQL errors while attempting to use the "addSource" feature on a string attribute with a value containing single quotes. I've not tested my solution, but I believe escaping single quotes should resolve the issue.


eugene7646:
So can you please test your solution and confirm that it indeed fixes the problem? ;)

PR #808, opened 2017-04-14
2453 added isCancelled flag to AddImageProcess so canellation won't hang

wschaeferB:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -231,17 +235,18 @@ public void run(String deviceId, String[] imageFilePaths) throws TskCoreExceptio
...
+					}
 				}
 				if (0 == tskAutoDbPointer) {
 					throw new TskCoreException("initAddImgNat returned a NULL TskAutoDb pointer");
 				}
-
-				runAddImgNat(tskAutoDbPointer, deviceId, imageHandle, timeZone, imageWriterPath);
+				if (!isCanceled) {
+					runAddImgNat(tskAutoDbPointer, deviceId, imageHandle, timeZone, imageWriterPath);
+				}
```
rcordovano: @bcarrier, @APriestman and @wschaeferB, I am still finding this code difficult to reason about to convince myself it is thread-safe. I am wondering if by attempting to avoid synchronizing the stop() method, 
we are attempting to slice things rather fine for a class that is already mostly synchronized on "this." The optimization of leaving one method out, if it actually gains us anything (I'm not sure),
may be uneccessary given the infrequent use of this class. I propose the following for safety and clarity:  

1. In the run() method, move lines 235 to 246 into the existing synchronized block.
2. Make the stop() method synchronized, just like the revert(), commit(), and getCurrentDirectory() methods already are.


Then the tskAutoDbPointer field is fully guarded by "this" (and can be so documented with an annotation), and does not need to be volatile. 
The "don't break this" comments can then be removed from stop(). And we can reason thus: 

1. If stop() is called before run(), run() cannot proceed to set tskAutoDbPointer before the isCanceled flag is set, which means that the read on line 247 cannot happen before the write on line 262.
2. If run() is called before stop(), we are guaranteed that run() will see a non-null tskAutoDbPointer and will call stopAddImagNat.

Am I correct? Your thoughts?

APriestman:
openImage can be slow for USB drives. If the stop is coming in on the EDT,
it could get stuck while waiting for openImage to complete.

wschaeferB:
While wrapping the code inside stop in a synchronized block on its own should be fine, however as Ann points out in the case of stop being called during run as soon as we put openImage() in a block synchronized on the same thing as the cancel block then we will always have to wait for open to finish for cancel to be executed, which will probably seem unresponsive at times.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
rcordovano: If we are calling stop from a cancel button in the EDT, I'm thinking it would be reasonable to give a wait cursor until the cancellation completes. Thoughts?

rcordovano:
If we are calling stop from a cancel button in the EDT, I'm thinking it would be reasonable to give a wait cursor until the cancellation completes. Thoughts?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -228,19 +233,19 @@ private AddImageProcess(String timeZone, boolean addUnallocSpace, boolean skipFa
...
+					if (0 != tskAutoDbPointer) {
+						throw new TskCoreException("Add image process already started");
+					}
+					imageHandle = openImage(imageFilePaths, false);
+					if (!isCanceled) {
+						tskAutoDbPointer = initAddImgNat(caseDbPointer, timezoneLongToShort(timeZone), addUnallocSpace, skipFatFsOrphans);
+					}
+					if (0 == tskAutoDbPointer) {
+						throw new TskCoreException("initAddImgNat returned a NULL TskAutoDb pointer");
+					}
```
APriestman: No initAddImageNat should still be fast. It's openImage that is slow for local disks.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -228,19 +233,19 @@ private AddImageProcess(String timeZone, boolean addUnallocSpace, boolean skipFa
...
-					tskAutoDbPointer = initAddImgNat(caseDbPointer, timezoneLongToShort(timeZone), addUnallocSpace, skipFatFsOrphans);
-				}
-				if (0 == tskAutoDbPointer) {
-					throw new TskCoreException("initAddImgNat returned a NULL TskAutoDb pointer");
+				long imageHandle;
+				synchronized (this) {
+					if (0 != tskAutoDbPointer) {
+						throw new TskCoreException("Add image process already started");
+					}
+					imageHandle = openImage(imageFilePaths, false);
```
APriestman: It won't hurt anything to have another check, but it would be nearly impossible for a user to hit cancel before we get to this point.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
rcordovano: Actually, isn't it now true that with the synchronized blocks, we only need one check for cancellation at the start of the block?

PR #809, opened 2017-04-17
New command line option (-c) to invoke MD5 checksum in mactime format

amm3:
It seems strange (to me) that `fls_lib` contains the functionality to include the MD5 hash in the bodyfile output, and yet it is not exposed on the fls command line.  In particular because the mactime/bodyfile format defines a field (position 0) for this value.

Ref: https://wiki.sleuthkit.org/index.php?title=Body_file

I'm not tied to a particular flag.  `-m` was taken and `-c` was not, so I just went with it.


bcarrier:
I'd like to be consistent with tsk_loaddb, which uses '-h'.  Can you change it to that?  thanks!

amm3:
Good point.  Done!

PR #812, opened 2017-04-18
Improved HFS attr data bounds checking

uckelman-sf:
Check that we won't overrun the end of an attr, rather than relying on the attr type for that.


zweger:
I should note that I observed sleuthkit exceeding the bounds of this buffer on a record which had type `HFS_ATTR_RECORD_FORK_DATA`, which sleuthkit doesn't claim to support. That said, the check added in this PR seems more prudent than merely checking the record type.

PR #819, opened 2017-04-26
Fix HFS+ highest inum lookup

zweger:
This changes `hfs_find_highest_inum`'s behavior to first check the HFS+ catalog file for the largest inum before falling back on the current method.

I came across a piece of evidence recently in which the HFS+ volume header `next_cat_id` slightly undershoots the highest inum on an HFS+ volume. This can cause problems for applications which rely on `last_inum` / `inum_count` for allocating arrays and such, although I don't believe this can cause other issues in sleuthkit itself. 

There's a comment in the source indicating my proposed patch is problematic, but [git blame](https://github.com/sleuthkit/sleuthkit/blame/5cd70af01be2bdd174ea70171574f3f07625ce96/tsk/fs/hfs.c) indicates it's been some time since `hfs_find_highest_inum` has been looked at. I'm assuming `hfs_cat_traverse` has been improved since then.

### DIFF
--- a/tsk/fs/hfs.c
+++ b/tsk/fs/hfs.c
@@ -1730,6 +1730,15 @@ hfs_cat_file_lookup(HFS_INFO * hfs, TSK_INUM_T inum, HFS_ENTRY * entry,
 }
 
 
+static uint8_t
+hfs_find_highest_inum_cb(HFS_INFO * hfs, int8_t level_type,
+    const hfs_btree_key_cat * cur_key,
+    TSK_OFF_T key_off, void *ptr)
+{
+  *((TSK_INUM_T*) ptr) = tsk_getu32(hfs->fs_info.endian, cur_key->parent_cnid);
+  return HFS_BTREE_CB_IDX_LT;
+}
+
 /** \internal
 * Returns the largest inode number in file system
 * @param hfs File system being analyzed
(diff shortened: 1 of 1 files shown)


tsk/fs/hfs.c
```diff
@@ -1730,6 +1730,15 @@ hfs_cat_file_lookup(HFS_INFO * hfs, TSK_INUM_T inum, HFS_ENTRY * entry,
 }
 
 
+static uint8_t
+hfs_find_highest_inum_cb(HFS_INFO * hfs, int8_t level_type,
+    const hfs_btree_key_cat * cur_key,
+    TSK_OFF_T key_off, void *ptr)
+{
+  *((TSK_INUM_T*) ptr) = tsk_getu32(hfs->fs_info.endian, cur_key->parent_cnid);
```
bcarrier: Since the goal of this to find the biggest, I'd feel safer if this check first verified that the new value was indeed bigger than what is already there before overwriting ptr. Traverse doesn't really promise what order the callback will be called in.

zweger: I'm fine with adding that check, but I don't understand why it would be necessary. Mustn't the callback be called in tree traversal order? How else would the traversal function know which branch to take?

And if the ordering constraints of the tree are violated, I'd think you'd have a lot bigger problems than the wrong highest inum: you wouldn't be able to find anything.

bcarrier:
i merged this in, but would still feel more comfortable with the check.  I didn't want to make the change myself because I'm not sure I  have the correct test images to verify it doesn't break somewhere.  

My main issue is that the API doc for the traverse method just says that the callback will be called for each record.  It does not guarantee order.  Some of the NTFS code processes trees in a non-sorted order (it processes it in the order that it is stored on disk).  I haven't had a chance to refresh my memory on the traverse method details to know if we can make that guarantee (and should therefore update the docs) or assume that it could be any order and we should add the check into the callback.

PR #823, opened 2017-05-04
Store file name along with inum in the dir map

APriestman:


uckelman-sf:
There are 2^32 = 4,294,967,296 distinct 32-bit hashes. The maximum length of a single filename component on NTFS is 255 characters. If we consider only filenames containing `[A-Za-z0-9._-]` (i.e., a 65-character alphabet), then there are 65^6 = 75,418,890,625 distinct 6-character filenames. (Note that this is a _massive_ underestimate, since NTFS permits Unicode characters in filenames.) By the Pigeonhole Principle, there will be collisions among 6-character filenames using a 32-bit hash. (In fact, about 94% of them will collide...) Using a hash like this

bcarrier:
In more practical terms though, this hash is only applied if two files are in the same folder, have the same MFT entry, and there is a remnant to the file name in the b-tree.   So, yes, there is still a chance that file would incorrectly be reported as an orphan instead of its proper parent, but I'd propose the real-world odds are quite a bit smaller than what you've listed.

uckelman-sf:
Why make getting the correct filename from a walk probabilistic when there's a way of guaranteeing that you get the correct filename? I'm going to dig into this further to see if there is any possibility for a correct optimization.

bcarrier:
The issue is memory. If we don't store the hash, we're storing every file name in memory. For a large image, that can get quite large. 

I'm not even 100% how your image got into this scenario.  We can't recreate it. 

At a higher-level, your image seems to have the following:
- The b-tree structure for the directory has an entry for 3.ini at MFT entry 92753 and sequence # 506.  The sequence number gets incremented each time the MFT entry is reallocated. If I am reading the fiwalk output correctly, it claims (via TSK) that the file is allocated (I'm skeptical of this though because of the sequence number). 
- After processing the b-tree, TSK goes through the MFT entries to look for other traces of files that could exist in that same folder (because the b-tree shuffling often pushes out file names quickly). That is when it finds the 5.tmp reference.  That name is stored in the MFT entry itself and it claims that the MFT sequence is now 563 and it is now unallocated. 

This situation only seems to come into play when the b-tree has an old reference to a file name that used to use an MFT entry and the MFT entry has a different name (but it was already flushed from the b-tree).

bcarrier:
NOTE: This was closed because #835 was created to go to the correct branch.

PR #827, opened 2017-05-10
Update jpeg_extract.cpp

nemermollon:
Incorrect use of pointer should modify variable values.

### DIFF
--- a/tools/fiwalk/plugins/jpeg_extract.cpp
+++ b/tools/fiwalk/plugins/jpeg_extract.cpp
@@ -37,7 +37,7 @@ int main(int argc,char **argv)
 	}
 
 	bool before = true;
-	for(char *cc=linebuf;*cc;*cc++){ // change ' '->'_" and '\t'->':' until we find the \t
+	for(char *cc=linebuf;*cc;cc++){ // change ' '->'_" and '\t'->':' until we find the \t
 	    if(before){
 		switch(*cc){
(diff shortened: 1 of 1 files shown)


bcarrier:
I don't really know this code at all in fiwalk, but, at a quick glance, it looks like this PR has the same behavior as the current code (i.e. it increments by one byte each time).  Did you experience different behavior?

kefir-:
The patch looks correct, but there should be no difference in behaviour. Operator precedence means the pointer is incremented first anyway, and then dereferenced to no effect. So the dereference in `*cc++` is superfluous. If this were rewritten as `(*cc)++` it'd be incrementing the byte value at the pointer, which would be wrong, whereas `cc++` increments the pointer to the next value, which is correct.

Example:
```
#include <stdio.h>
#include <string.h>
#include <stdlib.h>

int main(int argc, char *argv[])
{
   char *cc;
   cc=(char *)malloc(128 * sizeof(char));
   cc=strncpy(cc, "Hello World!", 128 * sizeof(char));
   printf("1: %s\n", cc);

   *cc++;
   printf("2: %s\n", cc);
   
   cc++;
   printf("3: %s\n", cc);
   
   (*cc)++;
   printf("4: %s\n", cc);
   
   return 0;
}
``` 
Output:
```
$ g++ -o inctest inctest.cc
1: Hello World!
2: ello World!
3: llo World!
4: mlo World!
```

PR #837, opened 2017-05-16
Better handle long path names (2)

APriestman:


tsk/auto/db_postgresql.cpp
```diff
@@ -1025,8 +1026,28 @@ int TskDbPostgreSQL::addFile(TSK_FS_FILE * fs_file, const TSK_FS_ATTR * fs_attr,
...
+    // Check if the path may be too long. The rest of the query should take up far less than 500 bytes.
+    if (strlen(name_sql) + strlen(escaped_path_sql) + 500 > bufLen) {
+        // The query may be long to fit in the standard buffer, so create a larger one.
+        // This should be a very rare case and allows us to not use malloc most of the time.
+        // The same buffer will be used for the slack file entry.
+        bufLen = strlen(escaped_path_sql) + strlen(name_sql) + 500;
+        if ((zSQL_dynamic = (char *)tsk_malloc(bufLen)) == NULL) {
+            PQfreemem(escaped_path_sql);
+            PQfreemem(name_sql);
+            return -1;
```
esaunders: Do we need to free name and escaped_path here as we do elsewhere?
Also, shouldn't this be returning 1 instead of -1?

tsk/auto/db_postgresql.cpp
```diff
@@ -1038,22 +1059,38 @@ int TskDbPostgreSQL::addFile(TSK_FS_FILE * fs_file, const TSK_FS_ATTR * fs_attr,
...
+            tsk_error_set_errno(TSK_ERR_AUTO_DB);
+            tsk_error_set_errstr("Error inserting file with object ID for: %" PRId64 , objId);
+            if (zSQL_dynamic != NULL) {
+                free(zSQL_dynamic);
+            }
+            free(name);
+            free(escaped_path);
+            PQfreemem(name_sql);
+            PQfreemem(escaped_path_sql);
+            return -1;
```
esaunders: Should this be "return 1;"?

Issue #393, opened 2015-01-29
ISO9660 is not processed properly. Only a part of files is enumerated by TSK


antonkukoba:
Seem's like there was a fix in a branch for this issue: http://marc.info/?l=sleuthkit-users&m=123182488321201 and it was never merged. So right now TSK tools cannot list the full amount of the files from the CD.

lfcnassif:
I think I have hit this bug too. TSK is ignoring one item per folder, and it can be a whole folder. I have created a test iso. It is attached.
[iso9660-tsk-test.zip](https://github.com/sleuthkit/sleuthkit/files/320687/iso9660-tsk-test.zip)

lfcnassif:
Any updates here? I think it is a critical bug because a whole directory tree can be ignored by tsk. Unfortunatelly I do not know iso9660 fs to propose a fix.

lfcnassif:
Hi @antonkukoba, recently my issue with ISO9660 was fixed on the devlop branch. Could you confirm if your issue was fixed too, so we can close this? Maye they are different issues...

PR #843, opened 2017-05-26
Fix dentry length check

APriestman:


lfcnassif:
Hi guys,
I tested this PR and seems like it solves the issue partially. I am getting strange results with tsk_loaddb (with this fix) related to slacks. Slacks of folders are being inserted into db (is it ok for ISO9660?) and the problem is all allocated files are being put under the deleted slack folders, not under the allocated folders. It can be reproduced with the image I attached on issue #393. Any ideas?

APriestman:
Yes we noticed that problem after making the fix as well. I'm hoping to look at it this week.

lfcnassif:
Thanks Ann!

PR #846, opened 2017-05-31
Fixed a few issues with slack files for directories.

APriestman:


lfcnassif:
Hi Ann,
Just to let you know, tested the PR locally and it fixes the problem with slacks. Thanks for taking a look so fast!

PR #848, opened 2017-06-07
Use system sqlite3 when available

uckelman-sf:
The #ifndef guards in sqlite3.h changed from `_SQLITE3_H_` to `SQLITE3_H` in commit 5471aca015 to sqlite, and this change first appears in release 3.14, from August 2016. This causes a problem for any program which compiles against both libtsk and the system-installed sqlite3, as the system sqlite3.h header from 3.14 onward is no longer compatible with the sqlite3.h header bundled with sleuthkit.

This PR modifies autotools builds to use the system sqlite3 when it is available. There are no changes for VC++ builds.


bcarrier:
Can you redirect this to release-4.4.2?

uckelman-sf:
> Can you redirect this to release-4.4.2?

Done. PR 860.

PR #871, opened 2017-07-17
2827 additional short descriptions

wschaeferB:


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -170,12 +170,25 @@ public String getDisplayName() {
...
+		} else if (artifactTypeId == ARTIFACT_TYPE.TSK_CONTACT.getTypeID()
+				|| artifactTypeId == ARTIFACT_TYPE.TSK_MESSAGE.getTypeID()
+				|| artifactTypeId == ARTIFACT_TYPE.TSK_CALLLOG.getTypeID()) {
+			if (null != getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER))) {
+				attr = getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER));
+			} else if (null != getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM))) {
+				attr = getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_FROM));
+			} else if (null != getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO))) {
+				attr = getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER_TO));
+			}
```
rcordovano: There exists custom module code (see JIRA for details) that uses a different set of phone number attributes for a TSK_CONTACT artifact: TSK_PHONE_NUMBER_HOME,  TSK_PHONE_NUMBER_OFFICE, TSK_PHONE_NUMBER_MOBILE, or TSK_PHONE_NUMBER. The same code potentially also, or only, attaches email address attributes: TSK_EMAIL_HOME, TSK_EMAIL_OFFICE, or TSK_EMAIL.  And in org.sleuthkit.autopsy.modules.iOS.ContactAnalyzer, there will be either TSK_PHONE_NUMBER or TSK_EMAIL, but not both, as is possible for the custom module. 

The custom code also creates TSK_MESSAGE artifacts that may be either phone messages (TSK_PHONE_NUMBER_TO) or email messages (TSK_EMAIL_TO). 

While the central repository/correlation engine (CR/CE) code in EmailArtifactUtil.getTypeFromBlackboardArtifact only deals with  TSK_PHONE_NUMBER_FROM, TSK_PHONE_NUMBER_TO, and TSK_PHONE_NUMBER, I think this method should work for general use, not just the CR/CE use case. @bcarrier, as the designer of the short description feature and as a SME,  should phone nuumbers take precedence over email addresses? Should one type of phone number take precedence over another type of phone number? I am thinking specificfally of TSK_CONTACT artifacts that may contani mulitple phone numbers and email addresses for a single individual.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -169,19 +170,63 @@ public String getDisplayName() {
 	 * @throws TskCoreException if there is a problem creating the description.
 	 */
 	public String getShortDescription() throws TskCoreException {
-		BlackboardAttribute attr = null;
-		if (artifactTypeId == ARTIFACT_TYPE.TSK_WEB_BOOKMARK.getTypeID()) {
-			attr = getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_URL));
+		List<BlackboardAttribute> attrs;
+		attrs = new ArrayList<BlackboardAttribute>();
+		if (artifactTypeId == ARTIFACT_TYPE.TSK_WEB_BOOKMARK.getTypeID()
```
esaunders: Would a switch statement make this easier to read?

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -169,19 +170,63 @@ public String getDisplayName() {
...
+		} else if (artifactTypeId == ARTIFACT_TYPE.TSK_DEVICE_ATTACHED.getTypeID()) {
+			attrs.add(getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DEVICE_ID)));
+		} else if (artifactTypeId == ARTIFACT_TYPE.TSK_CONTACT.getTypeID()
+				|| artifactTypeId == ARTIFACT_TYPE.TSK_MESSAGE.getTypeID()
+				|| artifactTypeId == ARTIFACT_TYPE.TSK_CALLLOG.getTypeID()) {
+			BlackboardAttribute name = getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_NAME));
+			if (name != null && !name.getDisplayString().isEmpty()) {
+				attrs.add(name);
+			}
+			if (null != getAttribute(new BlackboardAttribute.Type(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER))) {
```
esaunders: For each of the following, rather than calling getAttribute() twice would it be cleaner to do something like:
if  (attr = getAttribute(new BlackboardAttribute.Type(TSK_PHONE_NUMBER) != null) {
    attrs.add(attr);
}

I'd also recommend importing the attribute types individually to make it easier to read, i.e. TSK_PHONE_NUMBER instead of BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PHONE_NUMBER.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
wschaeferB: I agree a switch would be easier to read but I am having trouble figuring out how to write it since I can't switch on the artifactTypeId because the getTypeId() method isn't a constant expression.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
wschaeferB: I am getting many nested assignment hints / exclamation points in netbeans but that is indead much easier to read. Thank you for the suggestion.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
millmanorama: You should be able to get the actual ARTIFACT_TYPE enum object corresponding to the Id and switch on that

wschaeferB: Thank you, the fromID method was what I was not seeing before.

PR #872, opened 2017-07-19
Keep sleuthkit build logs

zhhl:


win32/updateBuildLibs.py
```diff
@@ -265,6 +251,9 @@ def main():
         usage()
 
     print('Updating source by %s branch.' % branch)
+    if not os.path.exists(LOG_PATH):
+        os.makedirs(LOG_PATH)
+
```
rcordovano: I think that it is great to get rid of the code duplication for creating the log files directory. However, why are we executing this code after we have already written log files to the directory, which will ostensibly cause it to be created? To me, it seems more logical to create the log directory before we start writing log files; as it is, I expect that unless there was a problem creating the log files above,  the os.path.exists(LOG_PATH) call will always return true, so this code "feels" wrong.

zhhl: We put everyday's build logs in the same log_path. Now I need to make this log_path with time stamp, so we can keep our logs dated and avoid over write the previous day's log. So I make this log_path as global and create it at the start point main() before we do update and build.

rcordovano: Sorry, I misread the code, and did not notice you had moved the log creation to the main method. I needed to expand the code in the diffs.

PR #873, opened 2017-07-20
2689: artifacts are content

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/BlackboardAttribute.java
```diff
@@ -518,8 +521,32 @@ public String getDisplayString() {
...
+					} 
+				} catch (TskException ex) {
+					// contnue with TimeZone.getDefault();
+				}
+				
+				String timeString = "0000-00-00 00:00:00";
+				long epochSeconds = getValueLong();
+				if (epochSeconds != 0) {
+					synchronized (DATE_FORMATTER) {
+						DATE_FORMATTER.setTimeZone(tzone);
```
bcarrier: I'd rather that this go into a new AbstractFile.epochToTime method that takes in a TimeZone argument.   The code in the attribute tries to find the Timezone and then calls either the new epochToTime() with that TimeZone or calls the original one. 

Also, epochToTime() is in AbstractFile, but all of the references are to FsContent. We should update those to make that more obvious. It took me a bit to find it.

raman-bt: Might there be another home for epochToTime() functions?

 AbstractFile isn't the the most intuitive p[lace I'd go looking for time conversion utility methods.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -316,7 +561,7 @@ public int hashCode() {
 	 */
 	@Override
 	public String toString() {
-		return "BlackboardArtifact{" + "artifactID=" + artifactId + ", objID=" + objId + ", artifactTypeID=" + artifactTypeId + ", artifactTypeName=" + artifactTypeName + ", displayName=" + displayName + ", Case=" + sleuthkitCase + '}'; //NON-NLS
+		return "BlackboardArtifact{" + "artifactID=" + artifactId + ", objID=" + getObjectID() + ", artifactObjID=" + artifactObjId + ", artifactTypeID=" + artifactTypeId + ", artifactTypeName=" + artifactTypeName + ", displayName=" + displayName + ", Case=" + getSleuthkitCase() + '}'; //NON-NLS
```
raman-bt: This isn't applicable any more if we BlackboardArtifact isn't extending AbstractContent. Right?

bindings/java/src/org/sleuthkit/datamodel/BlackboardAttribute.java
```diff
@@ -518,8 +521,23 @@ public String getDisplayString() {
...
+					final Content dataSource = getParentArtifact().getDataSource();
+					if ((dataSource != null) && (dataSource instanceof Image )) {
+						// return the date/time string in the timezone associated with the datasource,
+						Image  image = (Image) dataSource;
+						TimeZone tzone = TimeZone.getTimeZone(image.getTimeZone());
+						return epochToTime(getValueLong(), tzone);
+					} 
+				} catch (TskException ex) {
+					// contnue with TimeZone.getDefault();	
+				}
```
raman-bt: I agree with this. But this is in Sleuthkit datamodel and I didn't find any logging in those modules. I didn't think this warranted throwing an exception. Lets discuss alternatives.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
rcordovano: Quite true, sorry, I will delete this comment.

rcordovano:
Closing, because we are going to bring this into the artifacts-are-content branch and do an integration into develop later.

PR #886, opened 2017-08-03
QCOW2 Support

mkp-cse:


joachimmetz:
I would recommend using [libqcow](https://github.com/libyal/libqcow) instead

jtsylve:
I hate to disagree with @joachimmetz, but I like that this is a direct implementation and doesn't add yet another 3rd party dependency.  The LGPL nature of libyal libraries make it difficult to provide static binaries and I hate dealing with "DLL Hell".

joachimmetz:
> The LGPL nature of libyal libraries make it difficult to provide static binaries and I hate dealing with "DLL Hell".

Why? you can still make static libraries. parts of the sleuthkit are GPL2 which is the most strict license you have to comply to when making static binaries.

jtsylve:
AFAIK no part of libtsk is GPL2.  The only mention of any GPL2 I can find in the license information is the `srch_strings` tool.

joachimmetz:
you were talking about binaries, srch_strings is one of the binaries.

Also I'm not 100% about the license integrity of libtsk, see: https://github.com/sleuthkit/sleuthkit/issues/933

jtsylve:
Some projects may wish to only use `libtsk` and not the rest of the binaries.  I believe @bcarrier's intention has always been for `libtsk` to have a less restrictive license.  Whether the code conforms with that assumption may be an open question.

Other than not having a duel effort, what are the benefits of using libqcow over this implementation?

Maybe the solution is to use a compile-time config option that could use libqcow and if it's not found or if `--without-libqcow` is specified then it falls back to this native solution.

mkp-cse:
I can submit a new PR using libyal if that's the consensus solution. But I don't want to dedicate time to coding that up if it's unlikely to be merged due to licensing, dependencies, or other such issues.

joachimmetz:
> Some projects may wish to only use libtsk and not the rest of the binaries.

Some project may wish not to use built in basic qcow2 support.

> Other than not having a duel effort, what are the benefits of using libqcow over this implementation?

This is yet another thing to maintain inside sleuthkit; IMHO the codebase is currently already hard to maintain and could really benefit from a more modular approach.

> I can submit a new PR using libyal if that's the consensus solution. 

Is up to the maintainer, it was mainly a recommendation from my side seeing libtsk already uses libvhdi and libvmdk.

jtsylve:
> Some project may wish not to use built in basic qcow2 support.

I agree this is a possibility as well, which is why I think the best-of-both-worlds option would be to accept this PR as written (assuming there are no problems with it) and then another PR can be written to favor libqcow over the TSK implementation as a compile time configuration option.

This would give the option of not adding yet another 3rd party dependency while still supporting qcow2 or favoring the libyal implementation.  More choice is better, IMHO.

I agree that the ultimate decision is going to have to be whatever @bcarrier, et al. decide.  This is just my $0.02

joachimmetz:
> More choice is better, IMHO.

@tsylve tl;dr is more choice is more maintenance, more testing, more chances for serious errors

I also recommend reading this book https://en.wikipedia.org/wiki/The_Paradox_of_Choice

bcarrier:
Lots of topics to address here. Let's see if I can get them in one post.

For me, when we've integrated functionality in the past, I evaluate it based on:
- Overall quality of the code and how often it is being maintained.
- How many of the TSK OSes it also supports.
- How does it complicate the TSK build process.

My preference is often to keep the build process simple, which is why we've historically included SQLite as a single .h file and we wrote our own basic VHD writer this past year instead of trying to incorporate QEMU. 

As Joachim points, out, we already include a lot of libyal and another would not hurt.  Though, we have been maintaining a separate instance for 64-bit builds, but it sounds like there maybe a better solution to that now. 

As a side note, I think we had problems in the past making static libraries when we had multiple libyal projects because some had different versions of their own internal static libraries and we got conflicts. I'll need to dig up some internal tickets for the details.

In terms of TSK's licenses, srch_strings is the only GPL code in there and it doesn't use any part of 'libtsk'.  It is a self-contained executable.  It's only there for convenience from back in the day because not all strings programs extract unicode.  As another side note, that program is REALLY old.  It's been there since the CVS repo (pre-2008).  

The #933 issue was dead code that called into a GPL library that we prevented from being submitted. A public domain version of CRC was then used and the method below the dead code was calling that instead.  The dead code from #933 was not GPL itself. 

I'm not in a favor of having two options for supporting each type.  It adds too much confusion for users, debugging, testing, etc.

But, in terms of this specific PR. I don't have a strong opinion on using what is in the PR versus libyal.  On the one hand, there is value in having everyone use libyal and have it be the solid library, but on the other hand it adds some build complexity for what looks like a fairly simple format.  

@mkp-cse, did you look at libqcow before writing your own?  Were there reasons that you decided to write your own?

mkp-cse:
> did you look at libqcow before writing your own? Were there reasons that you decided to write your own?

I actually didn't know about libqcow before writing this patch. I did search for libraries but for whatever reason libyal didn't show up. The format is quite simple, so when I couldn't immediately find a library I didn't hesitate to create my own.

simsong:
In addition to creating a new QCOW2 implementation that would need to be supported, this patch now has conflicts that would need to be addressed. Given no discussion on this topic for 7 years, I'm closing this PR. Please feel free to resolve the conflicts, adopt libyal, and resubmit.

PR #893, opened 2017-08-15
Added getVisibleContentChildrenCount method

APriestman:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -3172,6 +3172,50 @@ int getContentChildrenCount(Content content) throws TskCoreException {
 			releaseSharedLock();
 		}
 	}
+	
+	/**
+	 * Counts if the content object children that will be shown in the tree. 
+	 * The children may include both abstract files and blackboard artifacts.
+	 * Note: this is generally more
```
millmanorama: I understand why this is in the SleuthkitCase, but it seems like it is introducing concepts form Autopsy.  Maybe there is a better name and/or way to describe it in the JavaDoc?

rcordovano:
Going to refactor this code into Autopsy.

PR #895, opened 2017-08-16
tsk/img/aff: loop initial declarations are not allowed in C89

gktrk:
Declaring an integer inside a for loop as in for(int i;;) is not
allowed in C89 and causes a build failure. Fix it by declaring the
variable just before the for loop.

### DIFF
--- a/tsk/img/aff.c
+++ b/tsk/img/aff.c
@@ -214,9 +214,10 @@ aff_imgstat(TSK_IMG_INFO * img_info, FILE * hFile)
 static void
 aff_close(TSK_IMG_INFO * img_info)
 {
+    int i;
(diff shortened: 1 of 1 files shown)


tsk/img/aff.c
```diff
@@ -216,7 +216,8 @@ aff_close(TSK_IMG_INFO * img_info)
 {
     IMG_AFF_INFO *aff_info = (IMG_AFF_INFO *) img_info;
     af_close(aff_info->af_file);
-	for (int i = 0; i < img_info->num_img; i++) {
+	int i;
```
bcarrier: Can you move this up to line 217 because other compilers are going to complain that it isn't defined at the top block.

gktrk: Fixed

gktrk:
Updated.

PR #901, opened 2017-08-23
2926 max size for unallocated files

APriestman:


APriestman:
@lfcnassif The PR wasn't quite finished before. The unalloc files should only be broken up from Autopsy, which also has an option to export them as single files.

lfcnassif:
OK thank you @APriestman !
We have a forensic application that uses TSK Java Bindings directly and we group non contiguous unalloc blocks up to 500MB (the java binding default), I've experienced slowdowns with 1GB. For sequential blocks, we have a separate param to break them, like you are doing here, and our default value is 10GB, but it is configurable and is a different param.
Currently looks like m_maxChunkSize depends on m_minChunkSize if it is > 0. We think they should be independent of each other and both configurable. The user shoud be able to group non adjacent blocks up to X and to not break sequential blocks or to break sequential blocks to a higher value Y >> X

bcarrier:
@lfcnassif , are you saying that you'd like another setAddUnallocSpace() method that allows the caller to specify the min and max (instead of just specifying the min)?

lfcnassif:
@bcarrier , yes! That would be great!

APriestman:
@lfcnassif Can you elaborate on exactly how you're using TSK? I can add another setAddUnallocSpace() method but how are you calling it? Are you using Java_org_sleuthkit_datamodel_SleuthkitJNI_initializeAddImgNat ?

lfcnassif:
Hi @APriestman of course. Yes, in the past I have used java AddImageProcess.run(). But I needed to store and access file layout info and there is no option for that in the java api. So now I am using tsk_loaddb to populate the db with file layout info and after that I use the java api to access TskFileRange.

APriestman:
@lfcnassif I need to close this and switch to a new PR. I accidentally merged in the wrong branch and I'm not sure how to get rid of it. The new PR is here:
https://github.com/sleuthkit/sleuthkit/pull/930

tsk_loaddb.exe should be back to how it was before - no chunking. The JNI code has our Autopsy defaults hard-coded in. I like the idea of being able to set both parameters from tsk_loaddb but don't have time to do it right now.

PR #902, opened 2017-08-24
Skip call to hfs_lookup_hard_link

APriestman:


bcarrier:
@uckelman and @jonstewart, can you take a look at this fix on some of your images?  We get significant speedups on images with large numbers of hard links with this and we don't see any changes on our images.  We aren't sure why the original code was doing this resolving (it was contributed code).

sf-jonstewart:
What's the intent of the issue in a nutshell?

But heck yeah, we will test, first on our regression test images, then on some real world junk.

bcarrier:
This is motivated by Autopsy issue [2855](https://github.com/sleuthkit/autopsy/issues/2855) and similar reports.  The problem is that if you add an HFS image with lots of hard links (such as time machine backups), it takes days just to add the files into the DB before ingest can start.

What we found was that most of the time was spent looking up a file named 'iNode35' in the folder of hard links only to find out that its corresponding "inode" number is 35.  This fix doesn't do the lookup.  It just assumes that 'iNode35' is at inode 35.   It works great for our test data, but I'm a bit reluctant since someone wrote that code at some point and I'm not sure if there is an edge case that we missed.

tsk/fs/hfs.c
```diff
@@ -1469,9 +1469,10 @@ hfs_follow_hard_link(HFS_INFO * hfs, hfs_file * cat,
...
 
-            TSK_INUM_T target_cnid;     //  This is the real CNID of the file.
-
-            target_cnid = hfs_lookup_hard_link(hfs, linkNum, FALSE);
+            // It seems like the linkNum is always the same as the CNID
+            // returned by hfs_lookup_hard_link, which is very inefficient when a 
+            // datasource contains large numbers of linked files.
+            TSK_INUM_T target_cnid = linkNum;
 
             if (target_cnid != 0) {
```
jtsylve: This check should likely be removed now.

tsk/fs/hfs.c
```diff
@@ -1532,9 +1533,10 @@ hfs_follow_hard_link(HFS_INFO * hfs, hfs_file * cat,
...
 
-            TSK_INUM_T target_cnid;     //  This is the real CNID of the file.
-
-            target_cnid = hfs_lookup_hard_link(hfs, linkNum, TRUE);
+            // It seems like the linkNum is always the same as the CNID
+            // returned by hfs_lookup_hard_link, which is very inefficient when a 
+            // datasource contains large numbers of linked files.
+            TSK_INUM_T target_cnid = linkNum;
 
             if (target_cnid != 0) {
```
jtsylve: This check should likely be removed now

tsk/fs/hfs.c
uckelman-sf: Is what you're saying is that `linkNum` should simply be returned here unchecked, as there's no possibility for having an error?

jtsylve:
Yes.  I'd suggest something along the lines of this.

```diff --git a/tsk/fs/hfs.c b/tsk/fs/hfs.c
index 9b05f2cc..87ab39b5 100644
--- a/tsk/fs/hfs.c
+++ b/tsk/fs/hfs.c
@@ -1651,22 +1651,10 @@ hfs_follow_hard_link(HFS_INFO * hfs, hfs_file * cat,
             uint32_t linkNum =
                 tsk_getu32(fs->endian, cat->std.perm.special.inum);
 
-            TSK_INUM_T target_cnid;     //  This is the real CNID of the file.
-
-            target_cnid = hfs_lookup_hard_link(hfs, linkNum, FALSE);
-
-            if (target_cnid != 0) {
-                // Succeeded in finding that target_cnid in the Catalog file
-                return target_cnid;
-            }
-            else {
-                // This should be a hard link, BUT...
-                // Did not find the target_cnid in the Catalog file.
-                error_returned
-                    ("hfs_follow_hard_link: got an error looking up the target of a file link");
-                *is_error = 2;
-                return 0;
-            }
+            // It seems like the linkNum is always the same as the CNID
+            // returned by hfs_lookup_hard_link, which is very inefficient when a
+            // datasource contains large numbers of linked files.
+            return linkNum;
         }
     }
     else if (file_type == HFS_LINKDIR_FILE_TYPE
@@ -1713,22 +1701,10 @@ hfs_follow_hard_link(HFS_INFO * hfs, hfs_file * cat,
             uint32_t linkNum =
                 tsk_getu32(fs->endian, cat->std.perm.special.inum);
 
-            TSK_INUM_T target_cnid;     //  This is the real CNID of the file.
-
-            target_cnid = hfs_lookup_hard_link(hfs, linkNum, TRUE);
-
-            if (target_cnid != 0) {
-                // Succeeded in finding that target_cnid in the Catalog file
-                return target_cnid;
-            }
-            else {
-                // This should be a hard link, BUT...
-                // Did not find the target_cnid in the Catalog file.
-                error_returned
-                    ("hfs_follow_hard_link: got an error looking up the target of a dir link");
-                *is_error = 2;
-                return 0;
-            }
+            // It seems like the linkNum is always the same as the CNID
+            // returned by hfs_lookup_hard_link, which is very inefficient when a
+            // datasource contains large numbers of linked files.
+            return linkNum;
         }
```

uckelman-sf:
I'm giving the patch a test on some of our OS X images now.

uckelman-sf:
This change produces the same results as 4.4.1 on the one smallish (8.6GB) OS X image I have handy right now.

rcordovano:
I am integrating this pull request as functional, with the understanding that we will circle back and improve the implementation.

Issue #882, opened 2017-08-01
Make and keep C code in libtsk C89 compatible


joachimmetz:
Python 2.7 on Windows depends on VS2008 see: https://wiki.python.org/moin/WindowsCompilers

This version of VS requires C code to be C89 compatible, please add a test to check if C code is C89 compatible and fix libtsk where needed

bcarrier:
Looks like your specific failure is from #831.  

Are you proposing that we supply -std=c89 to gcc to ensure that everything is C89?  Is it just that one line in fs_open that is the issue?

joachimmetz:
> Is it just that one line in fs_open that is the issue?

There are only a few lines that cause issues. While getting pytsk to build I found issues in:
* sleuthkit/tsk/fs/fs_name.c
* sleuthkit/tsk/fs/fs_open.c
* sleuthkit/tsk/img/raw.c

Also see: https://github.com/py4n6/pytsk/blob/master/setup.py#L263

> Are you proposing that we supply -std=c89 to gcc to ensure that everything is C89? 

You could use this for testing, though VC 9 for Python on Windows is more strict about C89

bcarrier:
I spent a few minutes on this tonight and added this to configure.ac:

`+CFLAGS="$CFLAGS -std=c89"`

It did not generate any errors for the files mentioned, so this won't help ensure that travis fails if we check in non-compliant code...

joachimmetz:
maybe add appveyor CI testing e.g. https://ci.appveyor.com/project/joachimmetz/libvhdi you see that one of the build targets is VS2008. As indicated that is the most strict one.

Also see AppVeyor config: https://github.com/libyal/libvhdi/blob/master/appveyor.yml

Issue #881, opened 2017-07-31
sys/types.h, sys/stat.h, unistd.h and fcntl.h missing from sleuthkit/tsk/img/raw.c


joachimmetz:
```
sleuthkit/tsk/img/raw.c:423:28: note: each undeclared identifier is reported only once for each function it appears in
sleuthkit/tsk/img/raw.c:423:39: error: ‘S_IFDIR’ undeclared (first use in this function)
     else if ((sb.st_mode & S_IFMT) == S_IFDIR) {
```

bcarrier:
What compiler / platform are you using?

joachimmetz:
Stock Ubuntu 14.04 with gcc and Python build flags

Issue #880, opened 2017-07-31
time.h include missing from sleuthkit/tsk/fs/fs_name.c


joachimmetz:
```
sleuthkit/tsk/base/tsk_os.h:201:16: error: ‘tzname’ undeclared (first use in this function)
 #define TZNAME tzname
                ^
sleuthkit/tsk/fs/fs_name.c:308:13: note: in expansion of macro ‘TZNAME’
             TZNAME[(tmTime->tm_isdst == 0) ? 0 : 1]);
````

**Comment by joachimmetz** [comment]
from `man tzname`

```
TZSET(3)                                                                                  Linux Programmer's Manual                                                                                  TZSET(3)

NAME
       tzset, tzname, timezone, daylight - initialize time conversion information

SYNOPSIS
       #include <time.h>

       void tzset (void);

       extern char *tzname[2];
       extern long timezone;
       extern int daylight;

   Feature Test Macro Requirements for glibc (see feature_test_macros(7)):

       tzset(): _POSIX_C_SOURCE
       tzname: _POSIX_C_SOURCE
       timezone, daylight: _XOPEN_SOURCE
           || /* Glibc since 2.19: */ _DEFAULT_SOURCE
           || /* Glibc versions <= 2.19: */ _SVID_SOURCE
```

also the presence of time.h also does not appear to be checked in configure.ac

**Comment by joachimmetz** [comment]
Also the source does not check for `__USE_POSIX` which if not set means tzname can be not available but `__tzname` might

**Comment by bcarrier** [comment]
So, if the code was something like as follows it would probably work on your system:

```
#if __USE_POSIX
#define TZNAME tzname
#else 
#define TZNAME __tzname
#endif 

```

**Comment by joachimmetz** [comment]
Not entirely, from time.h
```
/* Defined in localtime.c.  */
extern char *__tzname[2];       /* Current timezone names.  */
```
...
```
#ifdef  __USE_POSIX
/* Same as above.  */
extern char *tzname[2];
```

> on your system
Not my system but an Ubuntu 14.04 build VM of TravisCI

The hack/work-around in pytsk for now: https://github.com/py4n6/pytsk/blob/master/setup.py#L263

PR #919, opened 2017-08-26
Changes to make libtsk C-code C89 compatible #882 and fixes for #881 and #880

joachimmetz:


joachimmetz:
Also addresses https://github.com/sleuthkit/sleuthkit/issues/881

Issue #921, opened 2017-08-27
configure.ac cppunit check not portable - fails to detect cppunit on Fedora 26


joachimmetz:
In the latest version of the cppunit package for fedora, cppunit-config is removed in favor for pkg-config based solutions

Also see:
* https://apps.fedoraproject.org/packages/cppunit/changelog/
* https://github.com/OpenLightingProject/ola/issues/1043

joachimmetz:
While working on #941

The issue is surfacing on mingw: https://ci.appveyor.com/project/bcarrier/sleuthkit/build/1.0.317/job/y2iayum259v81edj#L213

having a look

joachimmetz:
I have a work-around but the bootstrap script is not sufficient see: https://github.com/sleuthkit/sleuthkit/issues/965

PR #922, opened 2017-08-27
Run unittests on Travis and added more portable cppunit check #921

joachimmetz:


bcarrier:
@joachimmetz, this is giving me an error in Cygwin:

```
checking if more special flags are required for pthreads... no
./configure: line 18679: syntax error near unexpected token `CPPUNIT,cppunit'
./configure: line 18679: `PKG_CHECK_MODULES(CPPUNIT,cppunit >= 1.12.1,ac_cv_cppunit=yes,ac_cv_cppunit=no)'

```
Thoughts?

joachimmetz:
@bcarrier 

1. make sure you have pkg-config and the latest autotools installed
2. Maybe this needs a bit of tweaking for the version of autotools on cygwin `PKG_CHECK_MODULES([CPPUNIT],[cppunit >= 1.12.1],[ac_cv_cppunit=yes],[ac_cv_cppunit=no])`

bcarrier:
I'm getting it on my mac too.   I didn't have pkg-config installed, but I just did via brew.  Same error though.   Note that I'm not sure if these systems have cppunit on them....

What was wrong with the original code?

joachimmetz:
> What was wrong with the original code?

Not supported on Fedora 26

bcarrier:
I updated pkg-config and it is now working. I'll update the docs to list the new build dependency.  

Is there a better way to detect that pkg-config isn't installed / correct instead of the obscure syntax error message?

bcarrier:
Does this mean that every person who downloads a tar ball of TSK source and runs './configure' needs to have the most up to date version of pkg-config?   That seems to be a bit much just for unit tests (that don't really do much right now).

joachimmetz:
While working on https://github.com/sleuthkit/sleuthkit/pull/941

The issue is surfacing on mingw: https://ci.appveyor.com/project/bcarrier/sleuthkit/build/1.0.317/job/y2iayum259v81edj#L213 

having a look

PR #924, opened 2017-08-27
timezone option to tsk_loaddb

lfcnassif:
This PR adds -z option to tsk_loaddb, like in other command line tools and to be consistent with the java api. It is important for FAT32 file systems.


lfcnassif:
Should I change autoDb->setAddUnallocSpace(true) to autoDb->setAddUnallocSpace(true, 500*1024*1024) to be consistent with the java api? Where is the best place to declare the value to be used for both jni and loaddb?

PR #925, opened 2017-08-27
Do not abort with corrupted data

lfcnassif:
In the past the processing was aborted with a corrupted file system. The changes in this PR solved the issue by ignoring the corrupted data instead of aborting the full ingesting.


bcarrier:
In theory, those methods to add/set attr runs could have more fine grained error reporting about corruption versus errors (some are because of size issues, some are because memory could not be allocated), but I agree this is probably the better solution.  If we ignore a memory issue in this pass, it will get caught soon there after.

lfcnassif:
Actually, I did not know about possible memory alloc issues, our case was a corrupted attr one. If memory issues are reported after that, I think the fix is ok too.

PR #930, opened 2017-08-29
Create max unallocated space file size

APriestman:


lfcnassif:
Hi @APriestman, I did not tested, but looks very good to me, thank you very much!

A suggestion would be to expose minChunkSize and maxChunkSize by JNI, so Autopsy could set them to its defaults, and others apps too. Currently we use tsk_loaddb because there is no option to store file layouts with the java api, but that should change in the future.

tsk/auto/auto_db.cpp
```diff
@@ -107,10 +108,16 @@ void TskAutoDb::setAddUnallocSpace(bool addUnallocSpace)
     setAddUnallocSpace(addUnallocSpace, -1);
 }
 
-void TskAutoDb::setAddUnallocSpace(bool addUnallocSpace, int64_t chunkSize)
+void TskAutoDb::setAddUnallocSpace(bool addUnallocSpace, int64_t minChunkSize)
+{
+    setAddUnallocSpace(addUnallocSpace, minChunkSize, -1);
+}
+
+void TskAutoDb::setAddUnallocSpace(bool addUnallocSpace, int64_t minChunkSize, int64_t maxChunkSize)
```
bcarrier: Just realized we can remove the bool here.  If they want to disable it, they should call the method that takes no size arguments.  It's too late to change the existing one (that just takes one size).

Issue #909, opened 2017-08-25
fls segfaults in hfs_UTF16toUTF8.


adambuchbinder:
To reproduce:

    $ unzip segfault.zip
    Archive:  segfault.zip
      inflating: segfault.img
    $ fls segfault.img
    Segmentation fault

This bug was found using [american fuzzy lop](http://lcamtuf.coredump.cx/afl/) and input files ultimately from [files.fuzzing-project.org](https://files.fuzzing-project.org/filesystems/).

Backtrace:

```
#0  __memcpy_sse2_unaligned ()
    at ../sysdeps/x86_64/multiarch/memcpy-sse2-unaligned.S:36
#1  0x0000000000469c44 in memcpy (__len=131070, __src=0x7d0f2e, 
    __dest=0x7ffff7f6e010) at /usr/include/x86_64-linux-gnu/bits/string3.h:51
#2  hfs_UTF16toUTF8 (fs=fs@entry=0x7cecd0, 
    uni=uni@entry=0x7d0f2e '0' <repeats 200 times>..., ulen=65535, 
    asc=asc@entry=0x7fffffffd450 "1+", alen=alen@entry=382, 
    flags=flags@entry=0) at hfs_dent.c:115
#3  0x0000000000455ecb in hfs_load_extended_attrs (
    fs_file=fs_file@entry=0x7cec50, 
    isCompressed=isCompressed@entry=0x7fffffffd650 "", 
    compDataInRSRC=compDataInRSRC@entry=0x7fffffffd660 "", 
    uncompressedSize=uncompressedSize@entry=0x7fffffffd670) at hfs.c:3822
#4  0x0000000000463e3b in hfs_make_catalog (fs_file=0x7cec50, hfs=0x7cecd0)
    at hfs.c:1940
#5  hfs_inode_lookup (fs=0x7cecd0, a_fs_file=0x7cec50, inum=<optimized out>)
    at hfs.c:2573
#6  0x0000000000428821 in tsk_fs_file_open_meta (a_fs=a_fs@entry=0x7cecd0, 
    a_fs_file=a_fs_file@entry=0x0, a_addr=a_addr@entry=4) at fs_file.c:128
#7  0x00000000004410fb in hfs_open (img_info=<optimized out>, 
    offset=<optimized out>, ftype=<optimized out>, test=<optimized out>)
    at hfs.c:6299
#8  0x0000000000435998 in tsk_fs_open_img (a_img_info=0x7ffff7f8f010, 
[...]

PR #931, opened 2017-08-31
Validate HFS attribute name length (Fixes: #909)

jtsylve:


jtsylve:
Sorry about the false starts... I've apparently been awake for too long.

jtsylve:
actually this fix isn't entirely correct at all.  The allocation isn't of the structure itself, but is based on a node size, which could mean that the entire structure we're reading from can be truncated.  This fix by itself is not sufficient to fix the issue.

Issue #878, opened 2017-07-29
error: 'vasprintf' was not declared in this scope


vanhauser-thc:
(on Cygwin x64)

g++ -DHAVE_CONFIG_H -I. -I../../../tsk  -I../../.. -Wall -I/usr/local/include  -g -O2 -MT dfxml.o -MD -MP -MF .deps/dfxml.Tpo -c -o dfxml.o dfxml.cpp
dfxml.cpp: In member function 'void xml::printf(const char*, ...)':
dfxml.cpp:267:29: error: 'vasprintf' was not declared in this scope
     if(vasprintf(&ret,fmt,ap) < 0){
                             ^
dfxml.cpp: In member function 'void xml::xmlprintf(const string&, const string&, const char*, ...)':
dfxml.cpp:433:29: error: 'vasprintf' was not declared in this scope
     if(vasprintf(&ret,fmt,ap) < 0){
                             ^
make[3]: *** [Makefile:477: dfxml.o] Error 1

I fixed it by inserting into tools/fiwalk/src/dfxml.cpp : 

#define _GNU_SOURCE 

But not sure if that is the best way.

joachimmetz:
I've reproduced the issue on appveyor https://ci.appveyor.com/project/bcarrier/sleuthkit/build/1.0.199/job/c3p1q5ig9qqepadh#L1317 while working on https://github.com/sleuthkit/sleuthkit/pull/941

joachimmetz:
Pending fix as part of https://github.com/sleuthkit/sleuthkit/pull/941

simsong:
Has this been addressed?

simsong:
Looks like it was fixed, from this section of code:
```C++
#if (!defined(HAVE_VASPRINTF)) || defined(_WIN32)
#ifndef _WIN32
#define ms_printf __print
#define __MINGW_ATTRIB_NONNULL(x)
#endif

extern "C" {
    /**
     * We do not have vasprintf.
     * We have determined that vsnprintf() does not perform properly on windows.
     * So we just allocate a huge buffer and then strdup() and hope!
     */
#ifdef _GNUC_
    int vasprintf(char **ret,const char *fmt,va_list ap)
        __attribute__((__format__(ms_printf, 2, 0)))
        __MINGW_ATTRIB_NONNULL(2) ;
#endif
    int vasprintf(char **ret,const char *fmt,va_list ap)
    {
        /* Figure out how long the result will be */
        char buf[65536];
[...]

PR #941, opened 2017-09-04
Added AppVeyor configuration with multiple build targets

joachimmetz:
Have a look at: https://ci.appveyor.com/project/bcarrier/sleuthkit/build/1.0.195

This adds the following build targets
* [x] VS 2015 Release Win32 build
  * ~~libvhdi and libvmdk building, libewf unable to find zlib.lib - swiched libewf_64bit to libewf-legacy and vstools~~
  * ~~LINK : fatal error LNK1181: cannot open input file 'libvhdi.lib' - building only the DLL does not put the files in the location libtsk is looking for them~~
  * ~~LINK : fatal error LNK1181: cannot open input file 'zlib.lib'~~
* [x] VS 2015 Release x64  build
  * ~~LINK : fatal error LNK1181: cannot open input file 'kernel32.lib' - issue with PlatformTools version~~
* [x] VS 2015 Debug build
  * LINK : fatal error LNK1181: cannot open input file 'kernel32.lib'
* [x] cygwin 32-bit and 64-bit
  * ~~running into https://github.com/sleuthkit/sleuthkit/issues/878~~
* [x] MinGW 32-bit MSYS
  * ~~missing regex, look for a package that provides it or change source~~
  * failing build - disabled for now requires more changes - https://github.com/sleuthkit/sleuthkit/issues/966
* [x] MinGW 64-bit MSYS2

I don't have visibility into the current appveyor configuration so this will need some tweaking to get right


bcarrier:
@joachimmetz, in our internal regression test setup for Autopsy, we've been using the win32/updateBuildLibs.py Python script to update the TSK dependent libraries.  To reduce what we need to maintain in AppVeyor config files and this Python script, I'd like to adapt that Python script as much as possible to work in AppVeyor.  The current AppVeyor integration was the bare minimum to get it working. 

I was thinking a design something like:
- AppVeyor config are responsible for:
 - Cloning the various repos 
 - Calling the TSK script that builds the various projects and the various targets
 - Eventually calling regression / unit tests

joachimmetz:
@bcarrier SG, I'll have a look at the python script. As indicated still in the process of flushing out this change.

joachimmetz:
@bcarrier qq. what is the license on https://github.com/sleuthkit/sleuthkit/blob/develop/win32/updateBuildLibs.py "Common Public License Version 1.0" ?

joachimmetz:
@bcarrier so it looks like I need tweak updateBuildLibs.py a bit to split it in different phases to match the appveyor and travis flows.

joachimmetz:
@bcarrier I assumed the license was CC a couple of questions:

* am I correct in assuming the libyal dependencies for the TSK build are uploaded to https://github.com/sleuthkit/libvmdk_64bit, https://github.com/sleuthkit/libvhdi_64bit, https://github.com/sleuthkit/libewf_64bit ?
* what purpose then do https://github.com/sleuthkit/libvhdi and https://github.com/sleuthkit/libvmdk have?
* do you have specific permissions set on the sleuthkit source directory? I'm running into some issues under cygwin and mingw not being able to change into a copy of the directory

bcarrier:
@joachimmetz:
- Yes, the script is CC.  I updated that.
- Yes, we build off of the _64bit repos.  The others were likely left overs from previous tests to update the VS solution files in place based on your latest branches.  I think it got messy though if they were updated in place and then you made changes and they conflicted with ours.   I just deleted them.  Though, perhaps the more direct fork is the easier solution when the vstools-based solution is used. 
- I don't intentionally have special permissions set.

joachimmetz:
@bcarrier thx, I'll see what I can find regarding the permissions, I have a work around

joachimmetz:
@bcarrier this is ready for review now

* major overhaul of win32/updateBuildLibs.py
  * added multiple build modes: build (will only build libtsk), prepare (sets up dependencies: libyal and zlib) and the default (not set) will provide the previous behavior
  * for appveyor this now uses libyal vstools
    * this depends on powershell, for libyal Windows build scripts
    * I've put the code of legacy libewf version here https://github.com/libyal/libewf-legacy
    * this will use the last tagged version from a libyal git repo, which is equivalent to the release
      * this prevent breakage due to working from HEAD
    * no need for sleuthkit specfic libyal_64bit repos for appveyor build
  * backwards compatibility for the previous approach is still there when you add the `--use-vstools` flag
    * however I was unable to test this
* on appveyor the MinGW-MSYS is broken more changes are needed, but I rather do that in separate PRs
* this fixes #878 for cygwin

bcarrier:
Hi @joachimmetz , sorry it is taken so long to look at this. Thanks for all of this! 

The current AppVeyor setup just runs Release_noLibs using VS2015. That's all. 

How does this enforce that the code stays with the older C styles?  Or will that be a later addition? 

My main reservation is that I'm not sure we are ready to change our Autopsy development process to always just grab the latest tag of the various libyal versions. But, we can add some additional targets later to cover that scenario.

joachimmetz:
> How does this enforce that the code stays with the older C styles? Or will that be a later addition?

So I did not wanted to put everything in a single PR. One thing that appveyor supports is multiple versions of Visual Studio, e.g. https://ci.appveyor.com/project/joachimmetz/libvmdk

Adding a VS 2008 build or Python 2.7 VS build target can catch compatibility issues. This might need some more work on vstools, but I want to know first if that is something the Sleuthkit project is willing to commit to.

> My main reservation is that I'm not sure we are ready to change our Autopsy development process to always just grab the latest tag of the various libyal versions.

For libyal tagged versions are dist packaged releases. If there is a need to stay with specific versions or other reasons not to use latest releases let's chat.

joachimmetz:
No response from project

bcarrier:
Hi @joachimmetz , sorry for the lack of responses to this.  The major overhaul of updateBuildLibs.py is what took so long.  Everytime I have accepted changes to the build system, it usually comes back to bite me and I needed to find time to fully understand those changes.

If the branch still exists, I'd like to cherry pick some of the features  in.  But, it looks like you deleted it.  Sorry.

PR #946, opened 2017-09-11
2855 more unique short desc

wschaeferB:


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -193,65 +194,61 @@ public String getDisplayName() {
...
 			case TSK_CALLLOG:
-				BlackboardAttribute name;
-				if (((name = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_NAME))) != null) && !name.getDisplayString().isEmpty()) {
-					attrs.add(name);
-				}
-				BlackboardAttribute attr;
-				if ((attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_PHONE_NUMBER))) != null) {
-					attrs.add(attr);
+				//get the first of these attributes which exists and is non null
+				if (((attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_NAME))) != null) && !attr.getDisplayString().isEmpty()) {
```
millmanorama: The amount of duplicate code( each if branch is identical except for the attribute type) here makes me nervous,  I feel like a different control structure might be better suited.  What about something like:
```JAVA
<Stream of types> types = { TSK_NAME, TSK_PHONE_NUMBER...}
for (type t: types){
atttr = getAttribute(t);
if (attr != null){
break;
}
}
```

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -193,65 +194,61 @@ public String getDisplayName() {
...
+		if (attr != null) {
+			shortDescription.append(attr.getAttributeType().getDisplayName()).append(": ").append(attr.getDisplayString());
+		} else {
+			shortDescription.append(getDisplayName());
+		}
+		BlackboardAttribute date;
+		//get the first of these date attributes which exists and is non null
+		if ((date = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_DATETIME))) != null) {
+		} else if ((date = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_DATETIME_SENT))) != null) {
+		} else if ((date = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_DATETIME_RCVD))) != null) {
```
millmanorama: same comment as above

Issue #752, opened 2017-01-19
Block size not honoured


kefir-:
I recently worked on an SxS memory card, and it had a 2kb sector size. Autopsy 4.3.0 does not handle this, and that leads me to believe that the latest sleuthkit doesn't either, though I haven't tested it.

I tested with an older version of mmls, and although the man page says that sector size is honoured, it doesn't appear to be. Here's a test disk as fdisk sees it:
```$ sudo fdisk -l /dev/sdb
Disk /dev/sdb: 512 MiB, 536870912 bytes, 262144 sectors
Units: sectors of 1 * 2048 = 2048 bytes
Sector size (logical/physical): 2048 bytes / 4096 bytes
I/O size (minimum/optimal): 4096 bytes / 4096 bytes
Disklabel type: dos
Disk identifier: 0x0cdfa81c

Device     Boot Start    End Sectors  Size Id Type
/dev/sdb1         512 262143  261632  511M  c W95 FAT32 (LBA)
```
Here's mmls' view of the same drive:
```
$ sudo mmls /dev/sdb
DOS Partition Table
Offset Sector: 0
Units are in 512-byte sectors

      Slot      Start        End          Length       Description
000:  Meta      0000000000   0000000000   0000000001   Primary Table (#0)
001:  -------   0000000000   0000000511   0000000512   Unallocated
002:  000:000   0000000512   0000262143   0000261632   Win95 FAT32 (0x0c)
003:  -------   0000262144   0001048575   0000786432   Unallocated
```
If I specify the sector size with -b 2048, things look much better. The same happens with an E01 image of the disk above, but the E01 format contains the correct sector size:

```
$ ewfinfo 2k-image.E01 | grep sector
	Bytes per sector:	2048
	Number of sectors:	262144
```
Here's a sample E01 image with 2kb sector size:

[2k-image.zip](https://github.com/sleuthkit/sleuthkit/files/718062/2k-image.zip)

bcarrier:
Thanks kefir.  If the E01 file has the sector size in it, we should be using that. Is that the case? Does it look OK if you give the E01? I can't tell from your message.   A quick glance at the code suggests it should work. 

`vs->block_size = img_info->sector_size;`


If the same file came in as a raw image though, I'm not sure we can detect that it's bad.  I think the only way to do that would be to open the FAT partition (in this case) and see how big it thinks it is and see if that is consistent with a 512-byte sector, 2k, or 4k.  

[...]

PR #951, opened 2017-09-14
Use sector size in E01.  Fix for issue with 4096 sector size with DOS…

bcarrier:
… partition table


bcarrier:
Addresses part of #752.

PR #961, opened 2017-09-15
2978 db schema version

millmanorama:


bindings/java/src/org/sleuthkit/datamodel/TskUnsupportedSchemaVersionException.java
```diff
@@ -0,0 +1,63 @@
...
+
+	/**
+	 * The current version of the schema that the code uses.
+	 *
+	 * @return The current version of the schema that the code uses.
+	 */
+	public DBSchemaVersion getCurrentVersion() {
+		return currentVersion;
+	}
+}
```
rcordovano: If we are going to narrow this class by providing state (as opposed to just supporting the base class constructors), I think we would do well to use that state to provide overrides of getMessage, getLocalizedMessage (can we?), and toString. The typical client of an Exception class would be very unlikely to try to use more specialized methods, since exception handling code is usually pretty generic. So I vote we remove the getters altogether as too specific to the Autopsy use case. I am certainly comfortable with the "user-friendly" message generated for this exception in Autopsy just stating incompatibility, as I don't see users caring about the version number details. In fact, it seems we might be best served to by doing way with the "non-standard" constructor and allowing the client of the class to put in whatever message they see fit, with the name of the class conveying the general idea of the exception type to the recipient. I think that is the usual thing to do when defining Exception subclasses (e.g., see https://docs.oracle.com/javase/tutorial/essential/exceptions/creating.html, https://stackoverflow.com/questions/1754315/how-to-create-custom-exceptions-in-java, etc.).

millmanorama:
I have no ideological problem making these classes more general, however I think there are a couple of things that need to be thought through:
-  I think if the version  class is going to be more general, we shouldn't even limit it to SchemaVersion. why not just Version, or VersionNumber or even SemanticVersion? The last suggestion is in line with my next point:
- If the version class is going to be more general, I think it should include a patch number.  Then it is useful for the representation of Semantic Versioning according to the SemVer spec.  Further I think many user may expect/want it to have a patch number in accordance with SemVer.
- I swear I have seen Exception classes with getters for more details (of course I can't find any examples now...) , but if you are okay with a generic "user friendly" message, and prefer to only implement the standard constructors I will do that.

bindings/java/src/org/sleuthkit/datamodel/DBSchemaVersion.java
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * The minor version is incremented for compatible changes that are use-able by
+ * older versions of TSK, although the new schema may not be fully taken
+ * advantage. For example, adding an index should be backwards compatible: the
+ * old version will still be able to open and use the db, but query performance
+ * may or may not be affected. Also, adding a column to a table should be
+ * backwards compatible as old versions should simply ignore it.
+ */
+public class DBSchemaVersion implements Comparable<DBSchemaVersion> {
+
```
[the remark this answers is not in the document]
millmanorama: Can easily do these changes to the docs

bindings/java/src/org/sleuthkit/datamodel/DBSchemaVersion.java
```diff
@@ -0,0 +1,105 @@
...
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
+			return false;
+		}
+		final DBSchemaVersion other = (DBSchemaVersion) obj;
+		return this.compareTo(other) == 0;
+	}
+}
```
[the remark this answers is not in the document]
millmanorama: I purposely left this out because I think the semantics are ambiguous:  When invoking isCompatibleWith() which version number do you invoke it on, and which is the method parameter?  Maybe we can come up with a better/less ambiguous method name?

bindings/java/src/org/sleuthkit/datamodel/TskUnsupportedSchemaVersionException.java
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.datamodel;
+
+/**
+ * Subtype of TskCoreException that is thrown when attempting to open a database
+ * with an unsupported schema version.
+ */
```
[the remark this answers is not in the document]
millmanorama: Ok, I will expand this java doc

bindings/java/src/org/sleuthkit/datamodel/DBSchemaVersion.java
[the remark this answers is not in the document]
millmanorama: Further, the logic described in the class docs does not map exactly to the logic we implement in SleuthkitCase because we provide schema upgrade paths, which are not even mentioned in the docs for DBSchemaVersion. Other clients of a more general SchemaVersionNumber might not have upgrade paths and so any  isCompatibleWith() method we write for use in SleuthkitCase might not be valid for other clients.  We are string to get into the GNU Libtools approach if we want a general compatibility check I think.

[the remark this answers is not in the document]
rcordovano: I'm good with the idea we discussed in person: making this a base class (with all three version parts) and then creating a CaseDbSchemaVersion subclass that has the case specific isConmpatible method.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -5273,10 +5326,7 @@ public Image getImageById(long id) throws TskCoreException {
 				long type = rs1.getLong("type"); //NON-NLS
 				long ssize = rs1.getLong("ssize"); //NON-NLS
 				String tzone = rs1.getString("tzone"); //NON-NLS
-				String md5 = "";
-				if (getSchemaVersion() > 2) {
-					md5 = rs1.getString("md5"); //NON-NLS
-				}
+				String md5 = rs1.getString("md5"); //NON-NLS
 				String name = rs1.getString("display_name");
```
millmanorama: @rcordovano we discussed this, but I just thought I would point it out for review again

rcordovano: Looks good to me.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -475,45 +483,70 @@ private void updateDatabaseSchema(String dbPath) throws Exception {
...
-				schemaVersionNumber = updateFromSchema5toSchema6(schemaVersionNumber, connection);
-				schemaVersionNumber = updateFromSchema6toSchema7(schemaVersionNumber, connection);
+				// Each method should examine the schema version passed to it and either:
+				//    a. do nothing and return the schema version unchanged, or
+				//    b. upgrade the database and return the schema version that the db was upgraded to.
+				dbSchemaVersion = updateFromSchema2toSchema3(dbSchemaVersion, connection);
+				dbSchemaVersion = updateFromSchema3toSchema4(dbSchemaVersion, connection);
+				dbSchemaVersion = updateFromSchema4toSchema5(dbSchemaVersion, connection);
+				dbSchemaVersion = updateFromSchema5toSchema6(dbSchemaVersion, connection);
+				dbSchemaVersion = updateFromSchema6toSchema7dot1(dbSchemaVersion, connection);
```
rcordovano: Why 7.1 instead of 7.0?

millmanorama: Gah, I can't remember the details, but there was a reason.  It had something to do with wanting to be able to upgrade existing "dev" cases that already had a schema 7 in them.  Maybe we could stay at 7.0 if tat matters.

rcordovano: As you have already seen, I had Ann restore the original 6 to 7 upgrade, and leave your new code in your 7 to 7.1 upgrade, which makes it easy to both upgrade our plentiful supply of "dev" cases and "real world" cases when we release.  It thinking this was what you intended, and got side-tracked. Stuff happens., it's all good now.

PR #987, opened 2017-10-17
863: Add and populate data_source_obj_id column to blackboard_artifac…

raman-bt:
…ts table

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
@@ -1637,7 +1637,7 @@ public class SleuthkitCase {
 	 * @throws TskCoreException if there is a problem getting the data source
 	 *                          obj ids.
 	 */
-	public List<Long> getDataSourceObjIds(String deviceId) throws TskCoreException {
+	List<Long> getDataSourceObjIds(String deviceId) throws TskCoreException {
 
 		// check cached map first
 		synchronized (deviceIdToDatasourceObjIdMap) {
@@ -1652,7 +1652,7 @@ public class SleuthkitCase {
 		ResultSet rs = null;
 		try {
 			s = connection.createStatement();
-			rs = connection.executeQuery(s, "SELECT obj_id FROM data_source_info WHERE device_id = " + deviceId); //NON-NLS
+			rs = connection.executeQuery(s, "SELECT obj_id FROM data_source_info WHERE device_id = '" + deviceId + "'"); //NON-NLS
 			List<Long> dataSourceObjIds = new ArrayList<Long>();
 			while (rs.next()) {
 				dataSourceObjIds.add(rs.getLong("obj_id"));
--- a/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
@@ -631,7 +631,7 @@ public class SleuthkitCase {
 				dbSchemaVersion = updateFromSchema5toSchema6(dbSchemaVersion, connection);
 				dbSchemaVersion = updateFromSchema6toSchema7(dbSchemaVersion, connection);
 				dbSchemaVersion = updateFromSchema7toSchema7dot1(dbSchemaVersion, connection);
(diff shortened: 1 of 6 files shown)


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1162,6 +1165,61 @@ private CaseDbSchemaVersionNumber updateFromSchema7toSchema7dot1(CaseDbSchemaVer
...
+		if (schemaVersion.getMajor() != 7) {
+			return schemaVersion;
+		}
+
+		if (schemaVersion.getMinor() != 1) {
+			return schemaVersion;
+		}
+
+		/*
+		 * This upgrade adds a minor version number column.
```
millmanorama: this comment is wrong/out of date

millmanorama:
I think accounts_relationships doesn't build since it is missing one of the classes added in this PR

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1571,6 +1627,54 @@ public AddImageProcess makeAddImageProcess(String timeZone, boolean addUnallocSp
...
+	 * @return A list of the data source object_id for the given device_id for
+	 *         the case.
+	 *
+	 * @throws TskCoreException if there is a problem getting the data source
+	 *                          obj ids.
+	 */
+	public List<Long> getDataSourceObjIds(String deviceId) throws TskCoreException {
+
+		// check cached map first
+		if (deviceIdToDatasourceObjId.containsKey(deviceId)) {
```
rcordovano: This is not thread-safe because the map is not guarded by any lock, nor is it a concurrent map class. Rather than making this thread safe, let's just not do this optimization for now. There will be very few rows in the data_source_info table, so it might actually be more costly to control access to the collection than to just execute the query.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1571,6 +1627,54 @@ public AddImageProcess makeAddImageProcess(String timeZone, boolean addUnallocSp
...
+	 *
+	 * @param deviceId device_id
+	 *
+	 * @return A list of the data source object_id for the given device_id for
+	 *         the case.
+	 *
+	 * @throws TskCoreException if there is a problem getting the data source
+	 *                          obj ids.
+	 */
+	public List<Long> getDataSourceObjIds(String deviceId) throws TskCoreException {
```
rcordovano: I don't think this needs to be a public API. I believe the only client is the CommunicationsManager, which is in the same package.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1162,6 +1165,59 @@ private CaseDbSchemaVersionNumber updateFromSchema7toSchema7dot1(CaseDbSchemaVer
...
+			// populate data_source_obj_id for each artifact
+			resultSet = connection.executeQuery(statement, "SELECT artifact_id, obj_id FROM blackboard_artifacts"); //NON-NLS
+			while (resultSet.next()) {
+				long artifact_id = resultSet.getLong("artifact_id");
+				long obj_id = resultSet.getLong("obj_id");
+				long data_source_obj_id = getDataSourceObjectId(connection, obj_id);
+
+				updstatement.executeUpdate("UPDATE blackboard_artifacts SET data_source_obj_id = " + data_source_obj_id + " "
+						+ "WHERE artifact_id = " + artifact_id);
+			}
```
rcordovano: We need to ask @bcarrier if he wants to do the data upgrade as well as the schema upgrade. I am generally in favor of always doing a data upgrade to eliminate the possibility of  issues going forward,  and we have a precedent for doing the data upgrade when we did the same denormalization to the tsk_files table (see updateFromSchema3toSchema4), but Brian did not want to do a data upgrade in updateFromSchema6toSchema7, so we should let him make the call.

rcordovano: I am commenting out the call to this code so we can proceed. We will resolve this question later.

PR #999, opened 2017-10-26
API Filters

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -1409,16 +1379,102 @@ private String normalizePhoneNum(String phoneNum) {
...
 		for (T val : values) {
 			result.append(val);
 			result.append(",");
 		}
+		
 		return result.substring(0, result.length() - 1);
 	}
 	
-	/*
+	/**
```
rcordovano: It looks like all of these SQL helper methods can be static.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -1409,16 +1379,102 @@ private String normalizePhoneNum(String phoneNum) {
...
+	 * Returns the SQL for a subfilter.
+	 */
+	private String getSubFilterSQL(SubFilter subFilter) {
+		String result = "";
+		
+        if (subFilter == null) {
+            return "";
+        } else if (subFilter instanceof DeviceFilter) {
+            result = getSQL((DeviceFilter) subFilter);
+        } else if (subFilter instanceof AccountTypeFilter) {
```
rcordovano: Given that it looks like the getSQL methods can be static, is it possible to add a getSQL method to the SubFilter class and thereby eliminate the need for casting here?

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -1409,16 +1379,102 @@ private String normalizePhoneNum(String phoneNum) {
 	 * Utility method to convert a list to an CSV string
 	 */
 	private static <T> String buildCSVString(Collection<T> values) {
-		if (values==null || values.isEmpty()) return "";
+		if (values==null || values.isEmpty()) 
+			return "";
+
```
rcordovano: I think you can use org.apache.commons.lang3.StringUtils.join to make a CSV list instead of rolling your own. See https://stackoverflow.com/questions/15815587/best-way-to-convert-list-to-comma-separated-string-in-java.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
raman-bt: Further downstream, both the getSQL() methods need to access non static methods (to convert device_ids to ds_object_ids, and to convert Account>type to account_type_id, respectively) 

That's why these cant be static.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
raman-bt: Are we good with adding a new dependency to the DataModel?  I was trying to avoid that, that's why  I rolled my own.

rcordovano: Ok, I forgot I was looking at a SleuthKit PR, not an Autopsy PR.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
rcordovano: Ok

bindings/java/src/org/sleuthkit/datamodel/AccountTypeFilter.java
```diff
@@ -18,30 +18,66 @@
...
 	/**
 	 * Get the list of account types.
-	 * 
-	 * @return list of account types
+	 *
+	 * @return list of account types.
 	 */
 	public List<Account.Type> getAccountTypes() {
-		return accountTypes;
+		return new ArrayList<Account.Type>(accountTypes);
```
millmanorama: If we are using sets internally I think we should return a set here.

bindings/java/src/org/sleuthkit/datamodel/AccountTypeFilter.java
```diff
@@ -18,30 +18,66 @@
...
+	 *
+	 * @return SQL String for the filter.
+	 */
+	@Override
+	public String getSQL(CommunicationsManager commsManager) {
+		if (accountTypes.isEmpty()) {
+			return "";
+		}
+
+		String sql = "";
```
millmanorama: I think this variable could be inlined

bindings/java/src/org/sleuthkit/datamodel/DeviceFilter.java
```diff
@@ -18,32 +18,71 @@
...
 	/**
 	 * Get the list of device id.
-	 * 
-	 * @return list of device Ids 
+	 *
+	 * @return list of device Ids.
 	 */
 	public List<String> getdeviceIds() {
-		return deviceIds;
+		return new ArrayList<String>(deviceIds);
```
millmanorama: Again.  I think this should be returned as a set for consistency.

PR #1008, opened 2017-11-09
3184 get data source size

dgrove727:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -5376,14 +5374,21 @@ public Image getImageById(long id) throws TskCoreException {
 					}
 				}
 				long size = rs1.getLong("size"); //NON-NLS
-				return new Image(this, obj_id, type, ssize, name,
+				
+				s3 = connection.createStatement();
```
rcordovano: I believe you can do this in just one round trip top the case database by doing a JOIN with the data_source_info table, instead of two separate queries.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -5356,10 +5352,17 @@ public Image getImageById(long id) throws TskCoreException {
...
+						+ "AND tsk_image_names.obj_id = " + rs1.getLong("obj_id")); //NON-NLS
 				List<String> imagePaths = new ArrayList<String>();
-				while (rs2.next()) {
-					imagePaths.add(rs2.getString("name"));
+				String device_id = null;
+				if (rs2.next()) {
+					device_id = rs2.getString("device_id");
+					do {
+						imagePaths.add(rs2.getString("name"));
+					} while (rs2.next());
```
rcordovano: The logic here could be simplified, and therefore made more maintainable, while still reducing the number of database round trips from three to two by doing the join between tsk_image_info and data_source_info instead of between tsk_image_names and data_source_info (note that only device_id is selected from data_source_info):

SELECT tsk_image_info.*, data_source_info.device_id FROM tsk_image_info, data_source_info WHERE tsk_image_info.obj_id = [object id] AND tsk_image_info.obj_id = data_source_info.obj_id; 

or even better, using the preferred syntax:

SELECT tsk_image_info.*, data_source_info.device_id FROM tsk_image_info INNER JOIN data_source_info ON tsk_image_info.obj_id = data_source_info.obj_id WHERE tsk_image_info.obj_id =  [object id]; 

Then the query for the image file names can be very simple and the if with the nested do-while in the code can be just a simple for loop again (note that just name is selected from tsk_image_names):

SELECT name FROM tsk_image_names WHERE tsk_image_names.obj_id = [object id];

PR #1014, opened 2017-11-16
896: Add columns to relationships table

raman-bt:
- Partially done


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -162,6 +163,15 @@ public long getObjectID() {
 		return this.sourceObjId;
 	}
 
+	/**
+	 * Gets the object id of the data source for this artifact.
+	 *
+	 * @return The data source object id.
+	 */
+	public long getDataSourceObjectID() {
```
rcordovano: I'm wondering if this needs to be a public API.

PR #1018, opened 2017-11-20
Acct refactor

bcarrier:


bindings/java/src/org/sleuthkit/datamodel/Account.java
```diff
@@ -116,40 +118,25 @@ public int hashCode() {
...
-
 		this.accountType = accountType;
 		this.accountUniqueID = accountUniqueID;
-
 	}
 
 	/**
-	 * Gets unique string identifier for the account
+	 * Gets unique identifier (assigned by a provider) for the account.
+	 * Example includes an email address.
```
millmanorama: I think the notion of provider should be elaborated, if we are going to introduce it.

bindings/java/src/org/sleuthkit/datamodel/AccountFileInstance.java
```diff
@@ -1,7 +1,7 @@
 /*
  * Sleuth Kit Data Model
  *
- * Copyright 2011-2017 Basis Technology Corp.
+ * Copyright 2017 Basis Technology Corp.
```
millmanorama: I thought we had been using 2011-current year for all copyright headers?

bindings/java/src/org/sleuthkit/datamodel/Account.java
bcarrier: Agreed.  That's why I made the jira to decide to elaborate or pick something else.

bindings/java/src/org/sleuthkit/datamodel/AccountFileInstance.java
bcarrier: For new files, it doesn't make sense to have dates that are before the file existed.

rcordovano: What has happened here is that I tried to standardize on the policy of using a copyright for the software for every file. Brian disagrees with this policy. It may, in fact, be incorrect. Unfortunately, we have lots of "my style" copyright notices in place now.

PR #1020, opened 2017-11-22
896 get account device instances with communications refactor

millmanorama:
refactored getAccountDeviceInstancesWithCommunications to use relationship.data_source_obj_id and to eliminate joins against the artifacts and account_to_instances_map tables.

updated unit tests after manual analysis of expected results.  

I also updated the relationship table name in this PR rather than doing another that will almost certainly have merge conflicts

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
@@ -698,32 +698,29 @@ public class CommunicationsManager {
 		try {
 			s = connection.createStatement();
 
-			// set up applicable filters
-			Set<String> applicableFilters1 = new HashSet<String>();
-			applicableFilters1.add(DateRangeFilter.class.getName());
-			applicableFilters1.add(DeviceFilter.class.getName());
+			//TODO: this could be static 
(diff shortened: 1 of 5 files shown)


bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -701,57 +699,62 @@ private void addAccountsRelationship(long account1_id, long account2_id, long re
...
-
-			// Date range filters are applied to the two inner queries
-			String innerQueryfilterSQL = getCommunicationsFilterSQL(filter, applicableFilters1);
-			if (!innerQueryfilterSQL.isEmpty()) {
-				innerQuery1 += " AND " + innerQueryfilterSQL;
-			}
-			if (!innerQueryfilterSQL.isEmpty()) {
-				innerQuery2 += " AND " + innerQueryfilterSQL;
-			}
+			//TODO: this could be static
```
rcordovano: I'm not going to hold up integration for this, but @millmanorama, please either act on the TODO, remove the comment, or make a story and reference it. Otherwise, the TODO will likely live on forever....thanks!

PR #1021, opened 2017-11-27
889 easy api changes

millmanorama:


rcordovano:
Does not auto-merge, please address and re-submit.

PR #1022, opened 2017-11-27
sanity check relationships that everything  is on the same device

millmanorama:


rcordovano:
Does not auto-merge, please address and re-submit.

PR #1034, opened 2017-12-04
3173 metrics sql

dgrove727:


bindings/java/src/org/sleuthkit/datamodel/LocalFilesDataSource.java
```diff
@@ -59,23 +68,34 @@
...
+	/**
+	 * Gets the object id of this data source.
+	 *
+	 * @return The object id.
+	 */
+	@Override
+	public long getId() {
+		return objectId;
+	}
+
```
rcordovano: There should already be an inherited implementation of this method.

bindings/java/src/org/sleuthkit/datamodel/Image.java
```diff
@@ -377,6 +382,61 @@ public String getDeviceId() {
...
+				closeResultSet(resultSet);
+				closeStatement(statement);
+				connection.close();
+			}
+		} catch (TskCoreException ex) {
+			LOGGER.log(Level.SEVERE, "An error occurred attempting to establish a database connection.", ex); //NON-NLS
+		}
+
+		return contentSize;
+	}
```
rcordovano: Looking at this method causes me to realize I made a mistake when specifying the getContentSize API. It should throw TskCoreException, including wrapping any SQLException in a TskCoreExcepton.  Sorry about the oversight, please update the code throughout as required, thanks!

dgrove727:
This PR should be merged alongside https://github.com/sleuthkit/autopsy/pull/3271 to avoid breaking Autopsy.

PR #1052, opened 2017-12-18
Fix libewf V2 API support

### LINKED ISSUE #642: Sleuthkit build fails with current libewf-experimental
After successfully compiling the current libewf-experimental source in Ubuntu 16.04 with GCC 5.3.1, the sleuthkit build (from GitHub master) fails out as follows:

/tmp/sleuthkit/tsk/img/ewf.c:67: undefined reference to `libewf_handle_read_random'

This is because "libewf_handle_read_random" (itself a v2 libewf API patch vs "libewf_read_random") no longer exists in libewf (libewf_handle_read_buffer_at_offset may be an appropriate alternative).

I understand that TSK only "officially" supports a specific 2013 build of libewf, but that build is increasingly difficult to find, and the link at http://www.sleuthkit.org/sleuthkit/download.php actually goes to a SourceForge site that has been more or less abandoned (it now forwards the user to the GitHub, which includes only libewf-experimental builds as releases, and some 2014 builds via Google).

It would be nice to have ewf.c updated to accommodate the current libewf conventions.

### LINKED ISSUE #780: ld: symbol(s) not found for architecture x86_64
i've installed the newest version of libewf from source and the binares from there works fine

./configure:
`Building:
   afflib support:                        yes
   libewf support:                        yes
   zlib support:                          yes

Features:
   Java/JNI support:                      no
   Multithreading:                        yes
`
sudo make:
```
Undefined symbols for architecture x86_64:
  "_libewf_handle_read_random", referenced from:
      _ewf_image_read in libtsk.a(ewf.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [img_cat] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all-recursive] Error 1
```

what is my problem here?

//Alexander

h3xx:
The deprecated function `libewf_handle_read_random` was removed in libyal/libewf@54f8d8ffb67ea3808336204d5a072db358066ec5, but at that
point it was just a wrapper for `libewf_handle_read_buffer_at_offset`.[1]

Solution: Replace deprecated libewf_handle_read_random call with the
recommended[2] replacement.

Also replace the the method detecting the presence of the V1 API. While I was unable to get `libewf-20160608` itself to compile anymore, I did confirm the presence of the `LIBEWF_OPEN_READ_WRITE` in the old source[3], and its absence in the new source.

References:

1. https://github.com/libyal/libewf/commit/54f8d8ffb67ea3808336204d5a072db358066ec5#diff-b9e76700775f6f9aa3efde388cb3bba9L74
2. https://github.com/libyal/libewf/commit/54f8d8ffb67ea3808336204d5a072db358066ec5#diff-fedcd7a7973ead791242e3a6d86e1a15L802
3. https://github.com/libyal/legacy/blob/master/libewf/libewf-20140608.tar.gz

Closes #642 
Closes #780

### DIFF
--- a/tsk/img/ewf.cpp
+++ b/tsk/img/ewf.cpp
@@ -67,7 +67,7 @@ ewf_image_read(TSK_IMG_INFO * img_info, TSK_OFF_T offset, char *buf,
 
     tsk_take_lock(&(ewf_info->read_lock));
 #if defined( HAVE_LIBEWF_V2_API )
-    cnt = libewf_handle_read_random(ewf_info->handle,
+    cnt = libewf_handle_read_buffer_at_offset(ewf_info->handle,
         buf, len, offset, &ewf_error);
     if (cnt < 0) {
         char *errmsg = NULL;
--- a/tsk/img/ewf.h
+++ b/tsk/img/ewf.h
@@ -21,9 +21,9 @@
 #include <libewf.h>
 #include <string>
 
-// libewf version 2 no longer defines LIBEWF_HANDLE
+// libewf version 2 no longer defines LIBEWF_OPEN_READ_WRITE
 #undef HAVE_LIBEWF_V2_API
-#if !defined( LIBEWF_HANDLE )
+#if !defined( LIBEWF_OPEN_READ_WRITE )
 #define HAVE_LIBEWF_V2_API
 #endif
 


h3xx:
I'd love to know from @joachimmetz if there's a better way to detect the V1 API than testing for the `LIBEWF_OPEN_READ_WRITE` macro.

joachimmetz:
V2 is the the current API version https://github.com/libyal/libewf-legacy. It introduces the libewf_handle functions. 

V3 is work in progress in the experimental version https://github.com/libyal/libewf and not finished yet

h3xx:
I'm not sure how anyone would know that the project marked "legacy" is the one that's supposed to be used, and not libyal/libewf.

"Legacy" to me means that it's old, unmaintained, and you shouldn't use it if you can avoid it.

joachimmetz:
It is but probably less unmaintained han libtsk

On Jan 12, 2018 22:02, "Dan Church" <notifications@github.com> wrote:

I'm not sure how anyone would know that the project marked "legacy" is the
one that's supposed to be used, and not libyal/libewf.

"Legacy" to me means that it's old, unmaintained, and you shouldn't use it
if you can avoid it.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/sleuthkit/sleuthkit/pull/1052#issuecomment-357349238>,
or mute the thread

joachimmetz:
Also legacy does not mean unmaintained
https://en.m.wikipedia.org/wiki/Legacy_system

dndusdndus12:
THX @h3xx you are my hero!!

I finally build with below dependencies (in Windows10 with VS2015)
Those are VERY NEW. (at least I believed)

libewf-experimental-20171104(https://github.com/libyal/libewf/releases)
libvhdi-alpha-20170223(https://github.com/libyal/libvhdi/releases)
libvmdk-alpha-20170226(https://github.com/libyal/libvmdk/releases)

zlib-1.2.11(https://zlib.net/)
dokan(https://github.com/joachimmetz/dokan)
bzip2-1.0.5(http://gnuwin32.sourceforge.net/packages/bzip2.htm)

If you want to build in the same way then you need to configure 
"additional library directories" for ALL but not libtsk
&
"additional include directories" only for libtsk

please note that Legacy is not a good choice. trust me I failed with it.
GOOD LUCK at your build!!

joachimmetz:
@dndusdndus12 if you run into known issue with the experimental version, don't say you've not been warned. 

> please note that Legacy is not a good choice. trust me I failed with it.

Please back this up with fact. I know the code and I do not trust your claims.

dndusdndus12:
@joachimmetz Yes. I could have a problem with that part.
But some of the links were dead and the edits I wanted to make were inevitable because of the limitations of the version.
Thank you for letting me(and to the others) know what I missed.

joachimmetz:
> But some of the links were dead and the edits I wanted to make were inevitable because of the limitations of the version.

which links, can you be specific so broken things can be fixed?

joachimmetz:
This out of scope of the original issue.

> But the zlib and gzip library homepages in these two links are incorrect.

In the future I recommend to directly reach out to the libewf project to let the maintainers know that links are outdated. I'll address the zlib link and older Dokan support is a known (and more complicated) issue: https://github.com/libyal/libewf/issues/76

> Also, BUILDING.txt in the win32 folder of TSK 4.6.2 (http://www.sleuthkit.org/sleuthkit/download.php) was not up to date. (Eg the "https://github.com/libyal/libvhdi" link on vmdk step)

I opt to file a separate issue with sleuthkit.

simsong:
@joachimmetz & @h3xx  - Is this still relevant? I'm trying to address the old PRs.

simsong:
If it's still relevant, can you bring the PR up to the current version of develop?

joachimmetz:
FYI v3 api is still under development

simsong:
So should I accept this or drop it?

bcarrier:
Some context on this. 

We've historically tried to prioritize @joachimmetz "stable" releases.  Some libraries have no "stable" release, such as libvmdk, and therefore we will link to a recent version we have done some basic sanity checks on. 

If the goal is to accept the v3 libewf line, which is labelled as "experimental", then I think TSK needs to still support people who want to stay on the stable v2. It's never been quite clear to me if people want v3 because it's simply the most recent or if it has features (such as ex01) that are required for them.

We've never done any analysis to understand if the v3 API version of libewf has any regressions on E01 from v2. Joachim is very thorough, so I'd expect there are none. But, I also trust his labelling of these as "experimental".  

FYI: On the Windows side, there is a bunch of work around NuGet packages and such that would also need to be updated to ensure it still compiles if the default is changed to v3.

simsong:
@bcarrier  — thank you for the context. I'm in complete agreement on your desire to stay compatible with v2 libewf.  I personally think that we should detect v2 or v3 and work with either.

However, this specific PR only changes 3 lines of code:

image

The question is — are these APIs supported in the legacy version of libewf that we typically use?

The GitHub actions that I've been developing do not yet have proper code coverage — I'm getting there, but there's been a delay because of the whole cppunit vs. catch2 issue. (And catch2 becoming catch 3. ugh! Now I know one of the reasons that @bcarrier has been loath to update the code base...) But I'm getting there. I just don't know if we have libewf in the full tests, and I don't know how to force GitHub to run the checks on a 6 year old PR.

simsong (changes requested):
Please bring this PR up to the current code base so that the unit tests will run.

bcarrier:
And looking at this PR a bit more, it is probably an abuse of existing #defines. It's using a v3 function within the #if of v2. So, a full solution would be properly detecting v2 vs v3 and then calling the appropriate method for each.

simsong:
So should we just kill it?

lfcnassif:
Just my 2 cents on libewf v3. We upgraded to it because it brings partial Ex01 support and mainly because of this:
https://github.com/sepinf-inc/IPED/issues/993

We have been using it for more than 2 years (20201230 snapshot), thousands of E01 images were processed without issues.

But since it is tagged as experimental, I agree libewf v2 should be still supported.

simsong:
My personal feeling is that our CI/CD pipeline should test and build with both and actually certify both.

h3xx:
Original author here. Wow, this PR does not take into consideration the things mentioned in this thread.

- Compatibility with libewf v1 API
- CI testing builds against both v1 and v2

Since this PR is so simple, I'd just as soon close it, especially if this does in fact break compatibility with libewf v1, which it looks like it might. I don't know if it can be improved, only completely re-written.

I'm not usually known for my C. I really only just patch and maintain C projects. That being said, I'm sure someone else more familiar with C could do a better job here and hit the missing points I mentioned above.

Sorry to waste everyone's time with this trivial and none too well-thought-out patch.

simsong:
Thanks. I will close.

PR #1065, opened 2017-12-28
3208 delay writing to the database until the end of the ingest pipeline

APriestman:


bindings/java/src/org/sleuthkit/datamodel/HashUtility.java
```diff
@@ -42,6 +42,19 @@
...
+	/**
+	 * Calculate the MD5 hash for the given FsContent and optionally store it in the
+	 * database. The hash will always be written to the AbstractFile object.
+	 *
+	 * @param file file object whose md5 hash we want to calculate
+	 * @param saveToDatabase true to save the hash to the database immediately
+	 *
+	 * @return md5 of the given FsContent object
+	 */
+	static public String calculateMd5(AbstractFile file, boolean saveToDatabase) throws IOException {
```
rcordovano: I would like to handle this API change a little bit differently. Let's create a new calculateMD5Hash method that takes a Content instead of an AbstractFile, calculates the hash, and returns it. Then let's deprecate the existing calculateMd5 method, which should call the new method and then save the result. Thanks!

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -6891,6 +6891,9 @@ public synchronized void close() {
 	/**
 	 * Store the known status for the FsContent in the database Note: will not
 	 * update status if content is already 'Known Bad'
+	 * 
+	 * This method is no longer called during normal ingest, as we save writing
```
rcordovano: These comments probably don't belong at the SleuthKit level, where we have no concept of ingest, since that is an Autopsy concept.

bindings/java/src/org/sleuthkit/datamodel/HashUtility.java
APriestman: Content doesn't have a getSleuthkitCase() method

rcordovano: Hmmm, of course not. Let's cast to AbstractContent and throw an exception if instanceof fails.

rcordovano: Okay, after in-person discussion, going with Content is more work than I expected. So we are going to stick with AbstractFile.

PR #1081, opened 2018-01-10
Fixed incorrect bounds check for HFS+ attr names

uckelman-sf:
Fixed incorrect bounds check for HFS+ attr names. Attr name length should be checked against max attr name length, not attr data length should be checked against max attr name length, not attr data length.


tsk/fs/hfs.c
```diff
@@ -4060,10 +4061,12 @@ hfs_load_extended_attrs(TSK_FS_FILE * fs_file,
                     goto on_error;
                 }
 
-                // name_len is in UTF_16 chars
-                if ((uint32_t)2 * tsk_getu16(endian, keyB->attr_name_len)  > attributeLength) {
```
bcarrier: @uckelman-sf, I think the original code should actually be comparing against keyLength.  The 'key' contains the name. 

Your updated check is also good, but I think we need to also tie it back to keyLength (and could keep yours just in case).  I think it would be something like:

`if ((uint32_t)2 * tsk_getu16(endian, keyB->attr_name_len)  > (keyLength-12)) {`

But, I could be off by a byte or so. I'm not sure if I have test data with this.  Do you have one to test it with?  We are wrapping up a release, so if you can't get to this for a while, we can take your change (which is better than what is currently there) and then incorporate the more strict test later.

(keyLength)

uckelman-sf: The name is 14 bytes from the start of the structure. The end of the structure is at `keyLength + 2`. So, the length of the name ought to be `keyLength + 2 - 14 == keyLength - 12`. I'll give that a try on our test images when I have a chance. Maybe in the next few days, but I cant make any promises.

bcarrier: OK. We merged this in since your fix was better than what was there. We'll get the even better fix into the next release if you can try it out.

Thanks.

uckelman-sf: PR #1162 has the second bounds check.

uckelman-sf: I checked some HFS+ images I have. None of them failed this additional check, which is what I expected. I don't have anything which is corrupt in the right way trigger the new error message.

PR #1092, opened 2018-01-16
rename getRelationships(account1, account1, filter) to getRelationshi…

millmanorama:
…pSources and have it return content insead of BlackboardArtifacts, to line up with the rest of the API


millmanorama:
updates with some fixes

PR #1100, opened 2018-01-17
Add offline condition for downloading ivy deps

bcarrier:


rcordovano:
@bcarrier, should I be concerned about the travis job failures here?

Issue #1088, opened 2018-01-15
Timeout (71896132)


Google-Autofuzz:
Hello sleuthkit team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
sleuthkit (tested with revision * develop c7ad6ad91b7bbac2a643912ee6d24e49f5ef8c68).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

TL;DR instructions:
[artifacts_71896132.zip](https://github.com/sleuthkit/sleuthkit/files/1631315/artifacts_71896132.zip)

* `mkdir project`
* `cp Dockerfile.sleuthkit /path/to/project/Dockerfile`
* `docker build --no-cache /path/to/project`
* `docker run -it image_id_from_docker_build`

From another terminal, outside the container:
`docker cp /path/to/attached/reproducer running_container_hostname:/fuzzing/reproducer`
(reference: https://docs.docker.com/engine/reference/commandline/cp/)

And, back inside the container:
`/fuzzing/repro.sh /fuzzing/reproducer`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
uthkit/tsk/vs/dos.c:774:22
    #133 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
    #134 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
    #135 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
    #136 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
    #137 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
    #138 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
    #139 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
    #140 0x5164d7 in dos_load_ext_table /fuzzing/sleuthkit/tsk/vs/dos.c:774:22
[...]

PR #1102, opened 2018-01-18
Add check that the extended partition has not already been added

APriestman:
#1088

### DIFF
--- a/tsk/vs/dos.c
+++ b/tsk/vs/dos.c
@@ -765,6 +765,10 @@ dos_load_ext_table(TSK_VS_INFO * vs, TSK_DADDR_T sect_cur,
                             "Starting sector %" PRIuDADDR
                             " of extended partition has already been used\n",
                             (TSK_DADDR_T)(sect_ext_base + part_start));
+                    tsk_error_reset();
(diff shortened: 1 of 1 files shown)


tsk/vs/dos.c
```diff
@@ -751,6 +751,25 @@ dos_load_ext_table(TSK_VS_INFO * vs, TSK_DADDR_T sect_cur,
...
+            TSK_VS_PART_INFO *part_info;
+
+            /* Sanity check to prevent infinite recursion in dos_load_ext_table.
+             * If we already have a partition with this starting address then
+             * return an error. This will prevent any more partitions from being
+             * added but will leave any existing partitions alone. */
+            part_info = vs->part_list;
+            while (part_info != NULL) {
+                if (part_info->start == (TSK_DADDR_T)(sect_ext_base + part_start)) {
+                    if (tsk_verbose)
```
bcarrier: We should also set an error code for this.  Something like:

```
            tsk_error_reset();
            tsk_error_set_errno(TSK_ERR_VS_BLK_NUM);
            tsk_error_set_errstr
                ("dos_load_ext_table: Loop in partition table detected");
```

PR #1113, opened 2018-01-29
Update schema and turn Report into Content

esaunders:


bindings/java/src/org/sleuthkit/datamodel/Report.java
```diff
@@ -18,49 +18,74 @@
...
 
 	/**
-	 * Construct a data transfer object for a row in the reports table.
+	 * Create a Report instance.
 	 *
 	 * @param id          Primary key from associated row in the case database.
 	 * @param path        Absolute path to report.
 	 * @param createdTime Created time of report (in UNIX epoch time).
 	 * @param reportName  May be empty
+	 * @param source	  The source from which the Report was created, if
```
rcordovano: I think that it would not be crazy to require a report name. In fact it would be good for the output of the collections search, where the the Source Name for reports is the report name. All we need to change is the comment. I am not sure why I allowed the empty string.

bindings/java/src/org/sleuthkit/datamodel/Report.java
```diff
@@ -18,49 +18,74 @@
...
- * This is a data transfer object (DTO) class that models reports.
+ * This is a class that models reports.
  */
-public class Report {
+public class Report implements Content {
 
 	static long ID_NOT_SET = -1;
 	private long id = ID_NOT_SET;
-	private final String path;
+	private final Path path;
```
rcordovano: Given that the id field is now a Sleuthkit object id, it would be nice to rename it objId or objectId and also deprecate getId(), replacing it with getObjectId(). I recall that when I wrote this class originally, I was trying to make it clear that this property was not an object id, which, of course, it now is.  We could make the column in the reports table be obj_id during the upgrade, perhaps.

bindings/java/src/org/sleuthkit/datamodel/Report.java
```diff
@@ -90,4 +115,186 @@ public String getSourceModuleName() {
...
+	@Override
+	public long getArtifactsCount(String artifactTypeName) throws TskCoreException {
+		return db.getBlackboardArtifactsCount(artifactTypeName, id);
+	}
+
+	@Override
+	public long getArtifactsCount(int artifactTypeID) throws TskCoreException {
+		if (artifactTypeID != BlackboardArtifact.ARTIFACT_TYPE.TSK_KEYWORD_HIT.getTypeID()) {
+			throw new TskCoreException("Reports can only have keyword hit artifacts.");
+		}
```
rcordovano: Why throw for this getArtifactsCount overload, but not for the others? It seems to me that we should be consistent.

bindings/java/src/org/sleuthkit/datamodel/Report.java
```diff
@@ -90,4 +115,186 @@ public String getSourceModuleName() {
...
+			LOGGER.log(Level.SEVERE, "Failed to get size of report.", ex);
+			// If we cannot determine the size of the report, return zero
+			// to prevent attempts to read content.
+			return 0;
+		}
+	}
+
+	@Override
+	public <T> T accept(ContentVisitor<T> v) {
+		throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
```
rcordovano: What does it mean that we support SleuthkitItemVistitor but not ContentVisitor? Is this just an interim implementation detail?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1237,6 +1237,43 @@ private CaseDbSchemaVersionNumber updateFromSchema7dot1toSchema7dot2(CaseDbSchem
...
+		if (schemaVersion.getMajor() != 7) {
+			return schemaVersion;
+		}
+
+		if (schemaVersion.getMinor() != 2) {
+			return schemaVersion;
+		}
+
+		Statement statement = null;
+		Statement updstatement = null;
```
rcordovano: Sort of an unusual abbreviation, is it intentional or a typo?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -7877,6 +7916,24 @@ public void updateImagePath(String newPath, long objectId) throws TskCoreExcepti
...
+	}
+
+	/**
+	 * Inserts a row into the reports table in the case database.
+	 *
+	 * @param localPath        The path of the report file, must be in the
+	 *                         database directory (case directory in Autopsy) or
+	 *                         one of its subdirectories.
+	 * @param sourceModuleName The name of the module that created the report.
+	 * @param reportName       The report name, may be empty.
```
rcordovano: See comment above - I'm not sure why I thought empty string report names were a good idea.

bindings/java/src/org/sleuthkit/datamodel/Report.java
esaunders: I can make the change to the Report class but in a previous conversation we decided not to change the name of the column in the reports table for backwards compatibility purposes (old Autopsy versions will be able to open upgraded cases).

bindings/java/src/org/sleuthkit/datamodel/Report.java
esaunders: Yes, ContentVisitor support is coming along in the "search" related code changes.

bindings/java/src/org/sleuthkit/datamodel/Report.java
esaunders: I thought I had covered all the others...where is it missing. Note that all the others end up calling this method.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
esaunders: No idea...copy and paste from previous update code. I'll take care of it when I get to filling out this method.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
esaunders: Are you simply looking for a comment update here too or do you want us to enforce it?
If we choose to enforce it, do we want to change the schema to make in a non null column?

bindings/java/src/org/sleuthkit/datamodel/Report.java
esaunders: This is an implementation of Content.getId() so creating new method and deprecating old doesn't seem like a good idea.

rcordovano: Sorry, I forgot about that decision. Please let it be, then.

bindings/java/src/org/sleuthkit/datamodel/Report.java
rcordovano: Okay, thanks, just checking.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
rcordovano: Okay, thanks.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
rcordovano: The schema already specifies "report_name TEXT NOT NULL." The "may be empty" in the parameter comments is literally true - a NULL argument is an error, but the empty string is legal. I don't think we should necessarily get into generally more rigorous method parameter checking as a policy at this point in the game, but I would rather not openly suggest in the comments in this class that the empty string is okay as a report name.

bindings/java/src/org/sleuthkit/datamodel/Report.java
rcordovano: I figured that was your intent, but it looks like 

public long getArtifactsCount(String artifactTypeName) throws TskCoreException

goes straight to the database without calling this overload.

esaunders: Ok, looks like I missed that one. I'll take care of this in a subsequent pull request.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
esaunders: Ok, the updated comment will be taken care of in a subsequent pull request. If that's everything can you go ahead and take this pull request (and the related Autopsy PR) and the requested changes will be included in the "search" PR?

rcordovano: Okay.

PR #1114, opened 2018-01-30
classpath and offline mode

rishwanth1995:
Modified build.xml to work in offline mode i.e the build process search for dependencies in /usr/share/java and added the required dependencies to the Readme.txt

### DIFF
--- a/bindings/java/build-unix.xml
+++ b/bindings/java/build-unix.xml
@@ -2,7 +2,7 @@
 <project name="TSKTestTargets">
 	<property name="dlls" value="../../win32/x64/Release_PostgreSQL"/>
     <property environment="env"/>
-
+    <property name="default-jar-install" location="/usr/share/java"/>
 	<target name="test" 
 			description="Performs regression tests."
 			depends="compile-test, copyTSKLibs">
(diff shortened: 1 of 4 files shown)


bindings/java/build-unix.xml
```diff
@@ -27,6 +27,42 @@
 		</java>
 	</target>
 
+	<condition property="online">
```
bcarrier: This should probably go into build.xml and we should not do the check if someone specifies online=false as an ant argument.

bindings/java/build-unix.xml
```diff
@@ -27,6 +27,42 @@
...
+	</target>
+
+	<target name="set-library-path-online" description="set this library path when the user is online" if="online">
+		<path id="libraries">
+		<fileset dir="${lib}">
+			<include name="*.jar"/>
+		</fileset>
+		<pathelement path="${build}"/>
+		</path>
+		<echo message="${online}"/>
```
bcarrier: Is this needed?

bindings/java/build-unix.xml
```diff
@@ -27,6 +27,42 @@
...
+		<http url="https://www.google.com/"/>
+	</condition>
+
+	<target name="set-library-path"  description="sets the path of the libraries" depends="set-library-path-online,set-library-path-offline">
+
+	</target>
+
+	<target name="set-library-path-online" description="set this library path when the user is online" if="online">
+		<path id="libraries">
+		<fileset dir="${lib}">
```
bcarrier: Can you look into making the indentation consistent for both this and the offline target?

bindings/java/build-unix.xml
```diff
@@ -2,7 +2,7 @@
 <project name="TSKTestTargets">
 	<property name="dlls" value="../../win32/x64/Release_PostgreSQL"/>
     <property environment="env"/>
-
+    <property name="default-jar-install" location="/usr/share/java"/>
```
bcarrier: We should also have a way to support them being in /usr/local/share/java.

bindings/java/build-unix.xml
rishwanth1995: I used that for testing I will remove and make another request

bindings/java/build-windows.xml
```diff
@@ -30,6 +30,14 @@
 	</target>
 
 	<target name="check-native-build" depends="check-build-32,check-build-64"/>
+	<target name="set-library-path" description="set-library-path">
```
bcarrier: We now have 3 copies of this code. We should redesign this so that all builds will get the libraries out of 'lib' and only offline will go to the other location.

bcarrier:
#1115 is the new PR for this.

PR #1115, opened 2018-01-31
classpath and offline mode added

rishwanth1995:
changed the classpath to a default location during offline mode and now the indentation looks fine.

### DIFF
--- a/bindings/java/build.xml
+++ b/bindings/java/build.xml
@@ -1,14 +1,14 @@
 <project xmlns:ivy="antlib:org.apache.ivy.ant" name="DataModel" default="dist" basedir=".">
 	<description>
-		Sleuthkit Java DataModel 
+		Sleuthkit Java DataModel
 	</description>
-	<condition property="os.family" value="unix">  
-		<os family="unix"/>  
-	</condition>  
-	<condition property="os.family" value="windows">  
-		<os family="windows"/>  
-	</condition>  
-	<import file="build-${os.family}.xml"/> 
+	<condition property="os.family" value="unix">
+		<os family="unix"/>
+	</condition>
+	<condition property="os.family" value="windows">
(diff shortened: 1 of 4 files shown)


bindings/java/build.xml
```diff
@@ -46,72 +39,99 @@
...
+		<mkdir dir="${i586}/linux"/>
 		<mkdir dir="${i686}"/>
 		<mkdir dir="${i686}/win"/>
 		<mkdir dir="${i686}/linux"/>
 	</target>
-  
-	<property name="ivy.install.version" value="2.3.0-rc2" />
+
+	<!-- set classpath for dependencies-->
+	<condition property="online">
```
bcarrier: We should skip this check if someone specifies online=false as an ant argument.

bindings/java/build.xml
```diff
@@ -46,72 +39,99 @@
...
+				<then>
+					<pathelement path="${default-jar-location}"/>
+				</then>
+			</if>
+			<fileset dir="${lib}">
+				<include name="*.jar"/>
+			</fileset>
+			<pathelement path="${build}"/>
+		</path>
+		<echo message="${online}"/>
```
bcarrier: Can be removed.

bindings/java/build.xml
```diff
@@ -46,72 +39,99 @@
...
+			<pathelement path="${build}"/>
+		</path>
+	</target>
+
+	<target name="set-library-path-offline" description="set the library path when the user is offline" unless="online">
+		<path id="libraries">
+			<if>
+				<equals arg1="${os.family}" arg2="unix"/>
+				<then>
+					<pathelement path="${default-jar-location}"/>
```
bcarrier: I don't see this being set anywhere.

PR #1122, opened 2018-02-01
3284 tsk disk image block size

dgrove727:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -395,7 +396,29 @@ public synchronized String currentDirectory() {
...
+			 *                       (e.g., a UUID).
+			 * @param imageFilePaths Full path(s) to the image file(s).
+			 *
+			 * @throws TskCoreException if a critical error occurs within the
+			 *                          SleuthKit.
+			 * @throws TskDataException if a non-critical error occurs within
+			 *                          the SleuthKit (should be OK to continue
+			 *                          the process)
+			 */
+			@Deprecated
```
rcordovano: Why deprecate this method? I don't think we should deprecate this method, it seems helpful to treat auto-detect as the default vs. requiring a user to enter the somewhat artificial sector size of zero argument. Do you have a counterargument?

dgrove727:
You make a sound point about the deprecation of the two argument 'run()' method (line 419) of the SleuthkitJNI class. I was deprecating it since it was doing nothing except calling the new one that has the third argument. But considering 0 is a default value, it would make sense to not necessarily discourage users from using the old method.

With that in mind, I have updated the pull request to remove the deprecation.

PR #1136, opened 2018-02-21
930 get relationships count

millmanorama:
WIP do not merge!


bindings/java/src/org/sleuthkit/datamodel/Relationship.java
```diff
@@ -36,6 +36,54 @@
  */
 public class Relationship {
 
+	public static class RelationshipKey {
```
rcordovano: Is public access required here?

PR #1160, opened 2018-03-12
Fixed compile error

jtsylve:
Compiling sleuthkit 4.6.0 while using a newer version of the C++ standard causes a compile error.

### DIFF
--- a/tools/srchtools/sigfind.cpp
+++ b/tools/srchtools/sigfind.cpp
@@ -302,7 +302,7 @@ main(int argc, char **argv)
             break;
         }
         else if (retval == -1) {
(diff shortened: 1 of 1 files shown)


bcarrier:
You got errors because there weren't spaces?  We have other code that doesn't have spaces too...

bcarrier:
Though, it looks like nearly all have the space and the ones that do not are in ".c" files.  Thanks.

jtsylve:
Correct, the issue here is that this code was in cpp file.  Modern versions of C++ support user-defined literals at the end of strings and this was getting confused with that on some compilers.  Putting the spaces resolves the compile issues and should serve the same purpose.

http://en.cppreference.com/w/cpp/language/user_literal

PR #1173, opened 2018-03-22
Let fileFiles() can find a partial file name.

zhhl:


rcordovano:
It turns out that this is a documentation error, rather than a coding error.

PR #1174, opened 2018-03-22
added support for postgresql on linux

rishwanth1995:


configure.ac
```diff
@@ -362,6 +421,8 @@ Building:
    afflib support:                        $ax_afflib
    libewf support:                        $ax_libewf
    zlib support:                          $ax_zlib
+   libvhdi support:                       $ax_libvhdi
+   libvmdk support:                       $ax_libvmdk
```
bcarrier: Add postgresql

tsk/auto/db_postgresql.cpp
```diff
@@ -125,11 +118,16 @@ PGconn* TskDbPostgreSQL::connectToDatabase(TSK_TCHAR *dbName) {
     PQescapeString(&userName_sql[0], userName, strlen(userName));
     PQescapeString(&password_sql[0], password, strlen(password));
     PQescapeString(&hostName_sql[0], hostNameOrIpAddr, strlen(hostNameOrIpAddr));
+#ifdef TSK_WIN32
+	snprintf(connectionString, 1024, "user=%s password=%s dbname=%S host=%s port=%s", userName_sql, password_sql, dbName, hostName_sql, hostPort);
```
bcarrier: Can't we use the %PRI.... for this?

tsk/auto/db_postgresql.cpp
```diff
@@ -148,25 +146,28 @@ TSK_RETVAL_ENUM TskDbPostgreSQL::createDatabase(){
...
-    // If you use single quotes, PostgreSQL will convert db name to lower case. If database was created using double quotes 
+    // If you use single quotes, PostgreSQL will convert db name to lower case. If database was created using double quotes
     // you now need to always use double quotes when referring to it (e.dg when deleting database).
     char createDbString[1024];
+#ifdef TSK_WIN32
     snprintf(createDbString, 1024, "CREATE DATABASE \"%S\" WITH ENCODING='UTF8';", m_dBName);
-
-    PGresult *res = PQexec(serverConn, createDbString); 
+#else
+	snprintf(createDbString, 1024, "CREATE DATABASE \"%s\" WITH ENCODING='UTF8';", m_dBName);
```
bcarrier: Same as above.  Can't we use PRIttocTSK?

configure.ac
rishwanth1995: sure

tsk/auto/db_postgresql.cpp
rishwanth1995: It works fine on linux and then it complains on windows. So, only I added the ifdef logic.

PR #1183, opened 2018-03-27
Fix several NTFS decompression bugs

uckelman-sf:
This patch fixes decompression errors in NTFS compressed files that have an initialized size which differs from the logical size, and also allows decompression of an NTFS compressed file to continue in the event that a corrupted compression unit is encountered.

### DIFF
--- a/tsk/fs/ntfs.c
+++ b/tsk/fs/ntfs.c
@@ -869,7 +869,7 @@ static uint8_t
 ntfs_uncompress_compunit(NTFS_COMP_INFO * comp)
 {
     size_t cl_index;
-    uint8_t recover_data = 0;
+    uint8_t recover_data = 0; // set to 1 when we proceed even with potentially corrupt data
 
     tsk_error_reset();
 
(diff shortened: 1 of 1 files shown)


bcarrier:
I had to revert this because it caused a regression in one of our test images.  The file ended up getting a few chunks of 0s were it used to have valid data.  The file was 100% initialized, so no special handling was required.

tsk/fs/ntfs.c
```diff
@@ -1360,18 +1367,29 @@ ntfs_attr_walk_special(const TSK_FS_ATTR * fs_attr,
...
+                                    flags & TSK_FS_META_FLAG_ALLOC) ?
+                                "Allocated" : "Deleted");
+                            free(comp_unit);
+                            ntfs_uncompress_done(&comp);
+                            return 1;
+                        }
+
+                        // handle the initialized size
+                        int64_t remanining_init_size = fs_attr->nrd.initsize - off;
+                        if (remanining_init_size < comp.buf_size_b) {
```
bcarrier: I don't understand this check.  Is your goal to provide zeros for everything past the initialized area?

kastonework: Yes - the intention is to zero-fill after the init size is reached, to match Windows' behavior for the file.

kastonework:
@bcarrier Is there any way to get a hold of the test image that's displaying the regression?

uckelman-sf:
@bcarrier Is the test image where you spotted the problem one you could give us? @kastonework is investigating the problem, but it would help her if she had that.

uckelman-sf:
PR #1276 addresses the bug causing the regression.

PR #1193, opened 2018-04-04
Fix many compiler warnings

uckelman-sf:
This pull request contains fixes for a number of compiler warnings.


bcarrier:
Actually, I was able to edit the PR and add the const values.  Waiting for Travis to run it.

PR #1198, opened 2018-04-10
added CI support for sleuthkit windows build

rishwanth1995:


setupLibs.py
```diff
@@ -0,0 +1,53 @@
...
+    print('Usage: python3 setupLibs.py [library directory]')
+    sys.stdout.flush()
+    sys.exit(1)
+
+def main():
+
+    base_Library_path = os.getcwd() #setting the base directory as current directory if no argument is passed
+
+    if len(sys.argv) == 2:
+        base_Library_path = sys.argv[1]
```
bcarrier: Let's change this script so that it puts the repos where the environment variables are set to.  So, it won't need to be passed in the root directory.  For each repo, it gets the HOME value and clones into there.

setupLibs.py
```diff
@@ -0,0 +1,53 @@
+# Copyright (c) 2017 Basis Technology.
+#
+# This software is distributed under the Common Public License 1.0
+
```
bcarrier: Add a comment to the effect of "This script makes the repositories needed to compile The Sleuth Kit and its dependencies."

win32/updateBuildLibs.py
```diff
@@ -19,7 +19,7 @@
 CURRENT_PATH = os.getcwd()
 # save the build log in the output directory
 LOG_PATH = os.path.join(CURRENT_PATH, 'output', time.strftime("%Y.%m.%d-%H.%M.%S"))
-
+APPVEYOR = os.getenv("APPVEYOR",False)
```
bcarrier: I'd rather not have any APPVEYOR specific stuff in here.   I think we should change this such that there is an argument that people can pass in so that only 64-bit Release_PostgreSQL is built.

win32/updateBuildLibs.py
```diff
@@ -34,8 +34,10 @@ def pullAndBuildAllDependencies(branch):
     # get the LIBEWF_HOME, LIBVHDI_HOME, LIBVMDH_HOME
     ewfHome = os.getenv("LIBEWF_HOME", "C:\\libewf_64bit")
     vhdiHome = os.getenv("LIBVHDI_HOME", "C:\\libvhdi_64bit")
-    vmdkHome = os.getenv("LIBVMDK_HOME", "C:\\libvmdk_64bit")
-
+    if(APPVEYOR):
+        vmdkHome = os.getenv("LIBVMDK_HOME", "C:\\libvmdk_64bit\\libvmdk")
```
bcarrier: I think the original code was a bug.  I think you can just replace the original one-liner with this one.

win32/updateBuildLibs.py
```diff
@@ -224,7 +232,10 @@ def buildTSK(wPlatform, target):
 
     outputFile = os.path.join(LOG_PATH, "TSKOutput.txt")
     VSout = open(outputFile, 'w')
-    ret = subprocess.call(vs, stdout=VSout)
+    if(APPVEYOR):
+        ret = subprocess.call(vs, stdout=sys.stdout)
```
bcarrier: What if we always just use sys.stdout instead of this if statement?

win32/updateBuildLibs.py
```diff
@@ -205,6 +210,9 @@ def buildTSK(wPlatform, target):
 
     print ("Building TSK " + str(wPlatform) + "-bit " + target + " build.")
     sys.stdout.flush()
+    if(APPVEYOR):
+        TSK_HOME = os.getenv("APPVEYOR_BUILD_FOLDER")
```
bcarrier: appveyor.yml should set TSK_HOME.

PR #1200, opened 2018-04-10
added CI support for sleuthkit windows build

rishwanth1995:


win32/updateBuildLibs.py
```diff
@@ -205,6 +217,9 @@ def buildTSK(wPlatform, target):
 
     print ("Building TSK " + str(wPlatform) + "-bit " + target + " build.")
     sys.stdout.flush()
+    if(APPVEYOR):
+        TSK_HOME = os.getenv("APPVEYOR_BUILD_FOLDER")
```
bcarrier: This should be in AppVeyor.

win32/updateBuildLibs.py
```diff
@@ -239,25 +255,58 @@ def usage():
...
+    for o,a in opts:
+        if o in ("-p","--platform"):
+            if a == '64':
+                Build_32 = False
+                Both =False
+            elif a == '32':
+                Build_64 = False
+                Both = False
+            elif a == '':
+                Both = True
```
bcarrier: This should be an error and we should display usage().

PR #1206, opened 2018-04-13
updated report to support URL paths

bcarrier:


bindings/java/src/org/sleuthkit/datamodel/Report.java
```diff
@@ -39,15 +39,16 @@
...
+	private final Path pathAsPath; // NULL if path is for a URL
 	private final long createdTime;
 	private final String sourceModuleName;
 	private final String reportName;
 
 	private Content parent; // The object from which the report was generated. 
 
 	private final SleuthkitCase db; // A reference to the database instance.
-	private FileChannel fileChannel; // Used to read report content.
+	private FileChannel fileChannel = null; // Used to read report content.
```
rcordovano: This is not necessary. Reference fields are initialized to null by default.

bindings/java/src/org/sleuthkit/datamodel/Report.java
```diff
@@ -39,15 +39,16 @@
 
 	static long ID_NOT_SET = -1;
 	private long objectId = ID_NOT_SET;
-	private final Path path;
+	private final String pathAsString;
+	private final Path pathAsPath; // NULL if path is for a URL
```
rcordovano: I don't see the value in having two fields here. It just complicates the code. A string path will be cleaner. I would not make the implicit isURL flag explicit either. The small cost of a path.startsWith("http") here and there does not justify adding a flag field, in my opinion.

PR #1216, opened 2018-04-18
Ignore VS2015 temporary directory

cilto:
Ignore temporary directory for Visual Studio 2015.
Reference for such meta info: https://github.com/github/gitignore/blob/master/VisualStudio.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ win32/BuildErrors.txt
 win32/BuildErrors-64bit.txt
 win32/.vs
 win32/tsk-win.VC.VC.opendb
+win32/tsk-win.VC.opendb
 win32/tsk-win.VC.db
 framework/msvcpp/framework/Debug/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -46,6 +46,7 @@ win32/BuildErrors.txt
 win32/BuildErrors-64bit.txt
 win32/.vs
 win32/tsk-win.VC.VC.opendb
+win32/tsk-win.VC.opendb
```
rcordovano: Is the preceding line a typo that needs to be removed?

cilto: No, we need both. One of these files is used in Visual Studio as a cache, and it depends on the version. It may cause permission error if uploaded.

(Reference: https://stackoverflow.com/questions/34975423/visual-studio-2015-git-error-opensomefile-vc-opendb-permission-denied-fa)

PR #1220, opened 2018-04-19
code review (UTF32)

sspring:
1. UTF32 was defined as unsigned long.sizeof(unsigned long) is 8bytes on 64bit linux(tested on Ubuntu17.10 with gcc7.2) while 4bytes on other cases.unexpacted behaviour may occurred if put self-increase operation on UTF32 pointer.I suggest define UTF32 as uint32_t becasue it has fixed size in any case.  
2. add tsk_UTF8toUTF32 function.

### DIFF
--- a/tsk/base/tsk_base_i.h
+++ b/tsk/base/tsk_base_i.h
@@ -28,7 +28,6 @@
 // most of the local files need this, so we include it here
 #include <string.h>
 
-
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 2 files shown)


sspring:
ok. the first problem is more important.just ignore the second one if you don't like it. I would create a new pr if my boss permits.

lfcnassif:
Hi sspring, any progress on your awesome work of adding apfs support to TSK?

sspring:
Well,I have done most of the work,including filevault decompression.I'm very sorry that I can't push any code here.and you should know the reason.  
https://github.com/sgan81/apfs-fuse  This may be useful for you.

lfcnassif:
That is sad, thank you for you best intentions.

PR #1226, opened 2018-04-26
275 events population infrastructure

millmanorama:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -131,8 +132,23 @@
...
+		eventBus.register(listener);
+	}
+
+	public void unregisterForEvents(Object listener) {
+		eventBus.unregister(listener);
+	}
+
+	void postTSKEvent(Object event) {
+		eventBus.post(event);
+	}
```
rcordovano: @bcarrier and @millmanorama I believe we should be moving towards simplifying SleuthkitCase so that one day it provides database primitives to "manager" classes that partition its functionality, i.e., the direction we are heading with CommunicationsManager, TimelineManager, Blackboard, etc. Putting a SleuthKit layer event bus in this class therefore seems like a step in the wrong direction. I think that we need a separate SleuthkitEventBus abstraction for the event bus. A mirroring API consisting of register, unregister and post would make sense to me. Perhaps, if we need to tie the EventBus to given case, the reference could be obtained through SleuthkitCase like the "managers" are, but I would not make it part of SleuthkitCase.

bindings/java/src/org/sleuthkit/datamodel/timeline/ArtifactEventType.java
```diff
@@ -50,64 +42,56 @@
...
-	 *
-	 * @return the ID of the the artifact type that this EventType is derived
-	 *         from.
+	 * @return a function from an artifact to a String to use as part of the
+	 *         full event description
 	 */
-	default int getArtifactTypeID() {
-		return getArtifactType().getTypeID();
-	}
+	String extractFullDescription(BlackboardArtifact artf) throws TskCoreException;
```
rcordovano: Do APIs such as this belong as pubic APIs at the SleuthKit level? They seem really specific to the UI, a criticism I had for many of the CommunicationsManager APIs. I am also a bit confused by the use case, given that we are passing artifacts into these methods, where it seems that the artifact would be part of the state of the event. I would like to have some time to discuss where all this is going; perhaps I am hampered by limited familiarity with the Autopsy timeline code.

By the way, artf seems like an awkward and unpronounceable abbreviation. Why not just spell out artifact?

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -0,0 +1,181 @@
...
+	 * Constructs a representation of the blackboard, a place where artifacts
+	 * and their attributes are posted.
+	 *
+	 * @param casedb The case database.
+	 */
+	Blackboard(SleuthkitCase casedb) {
+		this.caseDb = casedb;
+	}
+
+	public SleuthkitCase getSlethkitCase() {
```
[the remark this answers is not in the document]
millmanorama: sure.

millmanorama:
Looking back at the way development has progressed.  Maybe I should close this PR and the corresponding Autopsy ones, and replace them with new PRs with more "evolved"code.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
millmanorama: I am fine with this if this is way we want to go.  It should not be that hard.

PR #1258, opened 2018-05-16
added basic TSK tests for E01, vhd and vmdk

rishwanth1995:


.gitignore
```diff
@@ -178,3 +178,4 @@ unit_tests/base/test_base
 *.img
 *.vhd
 *.E01
+*.vmdk
```
bcarrier: Does this work?  Looks like we downloading files to the tests/data folder, not unit_tests/base/

PR #1261, opened 2018-05-23
3802:  Group tree by data source

raman-bt:
Sleuthkit changes to support this feature.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
@@ -153,22 +153,22 @@ public final class Blackboard implements Closeable {
 				
 		SleuthkitCase.CaseDbConnection connection = caseDb.getConnection();
 		caseDb.acquireSingleUserCaseReadLock();
-		Statement s = null;
-		ResultSet rs = null;
+		Statement statement = null;
+		ResultSet resultSet = null;
 		
 		try {
-			s = connection.createStatement();
-			rs = connection.executeQuery(s, queryString); //NON-NLS	
+			statement = connection.createStatement();
+			resultSet = connection.executeQuery(statement, queryString); //NON-NLS	
 			long count = 0;
-			if (rs.next()) {
-				count = rs.getLong("count");
+			if (resultSet.next()) {
+				count = resultSet.getLong("count");
 			}
 			return count;
 		} catch (SQLException ex) {
 			throw new TskCoreException("Error getting artifact types is use for data source." + ex.getMessage(), ex);
 		} finally {
-			SleuthkitCase.closeResultSet(rs);
-			SleuthkitCase.closeStatement(s);
+			SleuthkitCase.closeResultSet(resultSet);
+			SleuthkitCase.closeStatement(statement);
 			connection.close();
 			caseDb.releaseSingleUserCaseReadLock();
 		}
--- a/bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
@@ -24,7 +24,6 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.logging.Logger;
 
 /**
  * A representation of the blackboard, a place where artifacts and their
@@ -32,7 +31,6 @@ import java.util.logging.Logger;
  *
  */
 public final class Blackboard implements Closeable {
-	private static final Logger LOGGER = Logger.getLogger(Blackboard.class.getName());
 	private SleuthkitCase caseDb;
 
 	/**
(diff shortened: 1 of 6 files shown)


bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -0,0 +1,176 @@
...
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A representation of the blackboard, a place where artifacts and their
+ * attributes are posted.
+ *
+ */
+public final class Blackboard implements Closeable {
```
rcordovano: This class should not implement Closeable, i.e., it should not have a public close method. This is a mistake in the timeline-event-mgr branch.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -0,0 +1,176 @@
...
+	private SleuthkitCase caseDb;
+
+	/**
+	 * Constructs a representation of the blackboard, a place where artifacts
+	 * and their attributes are posted.
+	 *
+	 * @param casedb The case database.
+	 */
+	public Blackboard(SleuthkitCase casedb) {
```
rcordovano: This class should not have a public constructor. It should instead be a per case singleton available through a SleuthkitCase method, as in the timeline-event-mgr branch:

	public synchronized Blackboard getBlackboard() {
		if (null == blackboardInstance) {
			blackboardInstance = new Blackboard(this);
		}
		return blackboardInstance;
	}

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	/**
+	 * Closes the blackboard.
+	 *
+	 */
+	@Override
+	public synchronized void close() {
+		caseDb = null;
+	}
```
rcordovano: The close method should be removed, along with the implementation of Closeable, see prior comment.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -0,0 +1,176 @@
...
+	 * Gets both static and dynamic IDs.
+	 *
+	 * @param ds_obj_id  data source obj id 
+	 * 
+	 * @return The list of unique IDs
+	 *
+	 * @throws TskCoreException exception thrown if a critical error occurred
+	 *                          within tsk core
+	 */
+	public List<BlackboardArtifact.Type> getArtifactTypesInUseByDataSource(long ds_obj_id) throws TskCoreException {
```
rcordovano: This method returns type objects, not ids, so the doc comment is incorrect. Also, I think "static and dynamic ids" is confusing; what is meant is standard and custom types.

We should use Java naming conventions for method parameters: ds_obj_id => dataSourceObjId. This applies to several methods on this class.

We could probably shorten the method name, allowing the dataSourceObjId parameter to distinguish overloads of a getArtifactTypesInUse when we move SleuthkitCase.getArtifactTypesInUse into this class.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -0,0 +1,176 @@
...
+	 *
+	 * @param artifactTypeID artifact type id (must exist in database)
+	 * @param ds_obj_id         data source object id
+	 *
+	 * @return count of blackboard artifacts
+	 *
+	 * @throws TskCoreException exception thrown if a critical error occurs
+	 *                          within TSK core
+	 */
+	public long getBlackboardArtifactsCountByDataSource(int artifactTypeID, long ds_obj_id) throws TskCoreException {
```
rcordovano: This method name can be shortened by removing "Blackboard" since we are in the Blackboard class. We could probably also shorten the method name by dropping the "ByDataSource" part, allowing the dataSourceObjId to distinguish overloads of a getArtifactsCount when we move SleuthkitCase.getBlackboardArtifactsCount into this class.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -0,0 +1,176 @@
...
+	 *
+	 * @param artifactTypeID artifact type to get 
+	 * @param ds_obj_id data source to look under
+	 * 
+	 * @return list of blackboard artifacts
+	 *
+	 * @throws TskCoreException exception thrown if a critical error occurs
+	 *                          within TSK core
+	 */
+	public List<BlackboardArtifact> getBlackboardArtifactsByDataSource(int artifactTypeID, long ds_obj_id) throws TskCoreException {
```
rcordovano: This method name can be shortened by removing "Blackboard" since we are in the Blackboard class. We could probably also shorten the method name by dropping the "ByDataSource" part, allowing the dataSourceObjId to distinguish overloads of a getArtifacts when we move SleuthkitCase.getBlackboardArtifacts into this class.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifactTag.java
```diff
@@ -26,18 +26,36 @@
...
+	/**
+	 * Constructor
+	 * 
+	 * @param id tag id
+	 * @param artifact artifact tagged
+	 * @param content content from where the artifact was extracted
+	 * @param name tag name
+	 * @param comment comment
+	 */
+	public BlackboardArtifactTag(long id, BlackboardArtifact artifact, Content content, TagName name, String comment) {
```
rcordovano: Why do we need this constructor to be public?

bindings/java/src/org/sleuthkit/datamodel/ContentTag.java
```diff
@@ -28,26 +28,54 @@
...
+	/**
+	 * Constructor 
+	 * @param id tag id
+	 * @param content tagged content
+	 * @param name tag name
+	 * @param comment comment
+	 * @param beginByteOffset starting offset in content
+	 * @param endByteOffset sending offset on content
+	 */
+	public ContentTag(long id, Content content, TagName name, String comment, long beginByteOffset, long endByteOffset) {
```
rcordovano: Why do we need this constructor to be public?

bindings/java/src/org/sleuthkit/datamodel/Tag.java
```diff
@@ -24,7 +24,7 @@
  */
 public abstract class Tag {
 
-	static long ID_NOT_SET = -1;
+	public static long ID_NOT_SET = -1;
```
rcordovano: Why does this need to be public?

bindings/java/src/org/sleuthkit/datamodel/TagName.java
```diff
@@ -80,9 +80,17 @@ public static HTML_COLOR getColorByName(String colorName) {
...
+	/**
+	 * Constructor 
+	 * 
+	 * @param id tag id
+	 * @param displayName display name 
+	 * @param description description
+	 * @param color color to display tagged items in 
+	 * @param knownStatus if the file is known 
+	 */
+	public TagName(long id, String displayName, String description, HTML_COLOR color, TskData.FileKnown knownStatus) {
```
rcordovano: Why does this need to be public?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -9435,16 +9443,13 @@ public String getContextString() {
...
 	 *
 	 * @return A data source object id.
 	 *
-	 * @deprecated This only exists to support deprecated TSK object
-	 * constructors.
 	 */
-	@Deprecated
-	long getDataSourceObjectId(long objectId) {
+	//@Deprecated
+	public long getDataSourceObjectId(long objectId) {
```
rcordovano: Why has this been undeprecated and made public?

PR #1268, opened 2018-06-08
2823 Use map to keep track of which nodes have children

APriestman:
Will break Autopsy until https://github.com/sleuthkit/autopsy/pull/3837 is also merged.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
@@ -88,6 +88,7 @@ import org.sqlite.SQLiteJDBCLoader;
 public class SleuthkitCase {
 
 	private static final int MAX_DB_NAME_LEN_BEFORE_TIMESTAMP = 47;
+	private static int getContentChildrenCountCount = 0;
 
 	/**
 	 * This must be the same as TSK_SCHEMA_VER and TSK_SCHEMA_MINOR_VER in
@@ -121,6 +122,8 @@ public class SleuthkitCase {
 	private Map<Integer, BlackboardAttribute.Type> typeIdToAttributeTypeMap;
 	private Map<String, BlackboardArtifact.Type> typeNameToArtifactTypeMap;
 	private Map<String, BlackboardAttribute.Type> typeNameToAttributeTypeMap;
+	//private Set<Long> hasChildren;
+	private SparseBitSet hasChildrenBitSet;
 
 	private long nextArtifactId; // Used to ensure artifact ids come from the desired range.
 	// This read/write lock is used to implement a layer of locking on top of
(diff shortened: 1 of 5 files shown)


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -121,6 +121,7 @@
 	private Map<Integer, BlackboardAttribute.Type> typeIdToAttributeTypeMap;
 	private Map<String, BlackboardArtifact.Type> typeNameToArtifactTypeMap;
 	private Map<String, BlackboardAttribute.Type> typeNameToAttributeTypeMap;
+	private Map<Long, SparseBitSet> hasChildrenBitSetMap;
```
bcarrier: Can you add a comment about what the key is or name it so that it is obvious.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -531,6 +574,61 @@ private void initEncodingTypes(CaseDbConnection connection) throws SQLException,
...
+			closeStatement(statement);
+			releaseSingleUserCaseWriteLock();
+		}
+	}
+	
+	/**
+	 * Add the object IDs for a new data source to the has children map.
+	 * At present, we simply reload the entire table. 
+	 * TODO: Test whether it's worth doing a compound query to only add
+	 * entries from the new data source
```
rcordovano: We should either remove this comment or create a story and reference it with a "TODO (JIRA-XXXX):" style comment, lest this perpetually slips through the cracks.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -531,6 +574,61 @@ private void initEncodingTypes(CaseDbConnection connection) throws SQLException,
...
+			synchronized (this) {
+				if (hasChildrenBitSetMap == null) {
+					hasChildrenBitSetMap = new HashMap<Long, SparseBitSet>();
+				}
+				while(resultSet.next()) {
+					setHasChildren(resultSet.getLong("par_obj_id"));
+				}
+			}
+			long delay = System.currentTimeMillis() - timestamp;
+			logger.log(Level.INFO, "Time to initialize parent node cache: {0} ms", delay); //NON-NLS
```
rcordovano: Should we be retaining this performance data for the release, or is this more along the lines of debug info that can now be removed?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -274,9 +277,49 @@ private void init() throws Exception {
...
 	}
-
+	
+	/**
+	 * Use the internal map to determine whether the content
+	 * object has children (of any type).
+	 * @param c
+	 * @return true if the content has children, false otherwise
+	 */
+	public boolean getHasChildren(Content c) {
```
rcordovano: Does this need to be public?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
APriestman: I figured leave it in at least for this release so we can make sure this isn't causing a massive slowdown for opening cases.

rcordovano: Okay.

PR #1273, opened 2018-06-25
3949: View File in Directory broken for logical file sets when Group …

raman-bt:
…by Data Source is enabled

 - Name of data source is incorrect for LocalFilesDataSource, deviceid was being used instead of the name.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
@@ -2009,8 +2009,11 @@ public class SleuthkitCase {
 		acquireSingleUserCaseReadLock();
 		Statement statement = null;
 		ResultSet resultSet = null;
+		Statement statement2 = null;
+		ResultSet resultSet2 = null;
 		try {
 			statement = connection.createStatement();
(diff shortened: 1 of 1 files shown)


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -2027,13 +2030,18 @@ public DataSource getDataSource(long objectId) throws TskDataException, TskCoreE
 					 * No data found in 'tsk_image_info', so we build an
 					 * LocalFilesDataSource.
 					 */
+					
+					resultSet2 = connection.executeQuery(statement2, "SELECT name FROM tsk_files WHERE tsk_files.obj_id = " + objectId); //NON-NLS
+					String dsName = (resultSet2.next()) ? resultSet2.getString("name") : "";
+					resultSet2.close();
```
wschaeferB: Is there a benefit to closing resultSet2 here? I know we need to close it in the finally block, as we are doing, however that close leads me to believe this close might be unnecessary.

raman-bt:
You're right, this close is not necessary.  I'll fix it.

PR #1279, opened 2018-07-11
Add locking to JNI calls

APriestman:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -101,7 +103,8 @@ private SleuthkitJNI() {
 		 * @param fsHandle   The file system handle in which the file lives.
 		 */
 		private static void addFileHandle(long fileHandle, long fsHandle) {
-			synchronized (cacheLock) {
+			cacheLock.writeLock().lock();
```
bcarrier: This can probably be a read lock. I think we need the write block only when we are looking to remove something.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -567,7 +581,8 @@ private static long openImage(String[] imageFiles, int sSize, boolean useCache)
 		}
 		final String imageKey = keyBuilder.toString();
 
-		synchronized (HandleCache.cacheLock) {
+		getWriteLock();
```
bcarrier: Also can probably be read.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -674,7 +704,13 @@ public static long openFile(long fsHandle, long fileId, TSK_FS_ATTR_TYPE_ENUM at
...
 		 */
-		long fileHandle = openFileNat(fsHandle, fileId, attrType.getValue(), convertSignedToUnsigned(attrId));
+		long fileHandle = 0;
+		getReadLock();
+		try {
+			fileHandle = openFileNat(fsHandle, fileId, attrType.getValue(), convertSignedToUnsigned(attrId));			
+		} finally {
+			releaseReadLock();
+		}
 		HandleCache.addFileHandle(fileHandle, fsHandle);
```
bcarrier: Should this be in the lock?  Similar code is in other places.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -1195,13 +1287,51 @@ public static void cancelFinishImage(long imgHandle) {
 	 *                          queried
 	 */
 	public static long findDeviceSize(String devPath) throws TskCoreException {
-		return findDeviceSizeNat(devPath);
+		getReadLock();
```
bcarrier: I'm not sure this needs a lock.  This method is fully self contained and doesn't rely on any other handles.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -1195,13 +1287,51 @@ public static void cancelFinishImage(long imgHandle) {
...
+		try {
+			return findDeviceSizeNat(devPath);
+		} finally {
+			releaseReadLock();
+		}
 	}
 
 	public static boolean isImageSupported(String imagePath) {
-		return isImageSupportedNat(imagePath);
+		getReadLock();
```
bcarrier: Same.  Don't think lock is needed.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
APriestman: Our caches are just maps so I was trying to protect every place that accesses them. We used to have synchronized blocks any time the map was changed or accessed. Although I'm not actually sure the locks will ensure the map is properly visible on different threads. If we changed the maps to concurrent hash maps then we could probably have only the case closing section get the write lock.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
APriestman: This is because addFileHandle currently gets the write lock and we can't hold both read and write at the same time. If we can get rid of most of the write locks this will no longer have to be outside the lock.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
bcarrier: Oh yea.  Worth a try.

PR #1291, opened 2018-07-25
New methods to check for existing artifacts.

dgrove727:


bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -164,4 +165,91 @@ private long getArtifactsCountHelper(int artifactTypeID, String whereClause) thr
...
+     * @param attributesMap  The attribute type/value combinations to check for.
+     *
+     * @return True if all attributes are found; otherwise false.
+     */
+    private static boolean checkIfAttributesMatch(List<BlackboardAttribute> attributesList, Map<BlackboardAttribute.Type, String> attributesMap) {
+        for (Map.Entry<BlackboardAttribute.Type, String> mapEntry : attributesMap.entrySet()) {
+            boolean match = false;
+            for (BlackboardAttribute attribute : attributesList) {
+                BlackboardAttribute.Type attributeType = attribute.getAttributeType();
+                String attributeValue = attribute.getValueString();
```
millmanorama: This seems to be ignoring the fact that not all attributes are Strings.

dgrove727: I overlooked that point. Of course this also means the method signature will need to be changed to use "Object" instead of "String". I'll put together a change for this.

millmanorama:
As Brian said, if the user passes in BlackboardAttribute objects, the value
type is encoded in that and you can do the correct comparisons.

APriestman:
Please go ahead and add the "break" that codacy wants

PR #1294, opened 2018-07-31
1003: Add transactions support in CaseDbAccessManager api.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -154,12 +156,46 @@ public void createIndex(final String indexName, final String tableName, final St
 	 * @throws TskCoreException
 	 */
 	public long insertOrUpdate(final String tableName, final String sql) throws TskCoreException {
+		
+		CaseDbTransaction localTrans = tskDB.beginTransaction();
```
rcordovano: We don't really need to create an explicit transaction in the overrides of this method that just do the operations.

raman-bt: The intent of creating a transaction for a single operation is primarily to be able to call the overloaded method which does take a transaction.  This we way reuse those methods and avoid code duplication. 

I'll leave this as is, unless you really feel very strongly about it.

rcordovano: After further review, I see that this is the established approach, so leaving it alone agrees with precedent.

PR #1296, opened 2018-08-02
1007 ingest hang

raman-bt:


bcarrier:
No longer needed.  Fixed by #1297

PR #1309, opened 2018-08-20
Check that the index line length is not zero

APriestman:
Also close the index file handle if an error occurs while opening it

### DIFF
--- a/tsk/hashdb/binsrch_index.cpp
+++ b/tsk/hashdb/binsrch_index.cpp
@@ -1088,12 +1088,20 @@ int8_t
     if (hdb_binsrch_open_idx(hdb_info_base, htype))
         return -1;
 
-    /* Sanity check */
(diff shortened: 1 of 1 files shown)


bcarrier:
I'm fine with closing it out and I'd be fine going an additional step actually of resetting other variables of TSK_HDB_INFO that get set in that method before it bails because of an error.  I added a comment yesterday (https://github.com/sleuthkit/sleuthkit/commit/b92520b8d806822061a306b3b28d8edaa24842bf) to a branch I was working on to reflect that this method leaves the structure in an inconsistent state and that has been leading to confusion.  Even better if we reset everything, such as idx_size, idx_off, etc.  Then we can get rid of that comment.

APriestman:
I'll make a story for resetting the rest of TSK_HDB_INFO so we can make sure we do it right.

PR #1312, opened 2018-08-23
4114 tsk comment display n ame

wschaeferB:
Required changes regarding display name change of TSK_COMMENT for 4114

Japanese display name has been generated using google translate as the display name is stored in the case database and I do not know why it thinks the entire Japanese bundle file changed.

related autopsy PR: https://github.com/sleuthkit/autopsy/pull/4065


wschaeferB:
No longer required with the comment indicating column being renamed to C.

PR #1314, opened 2018-09-05
Hashdb format fix

bcarrier:


tsk/hashdb/idxonly.c
```diff
@@ -53,7 +50,7 @@ static void
                 fprintf(stderr,
                 "Failed to read name from index; using file name instead");
             hdb_base_db_name_from_path((TSK_HDB_INFO*)hdb_binsrch_info);
-            return;
+            return 0;
```
APriestman: Should this be returning 1?

bcarrier:
updated return code.

PR #1316, opened 2018-09-07
1024: Add a tsk_examiners table.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/Examiner.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Encapsulates an Autopsy examiner
+ * 
+ */
+public final class Examiner {
+	private final long id;
+	private final String loginName;				
+	private final String displayName;	
+	
+	public Examiner(long id, String loginName, String displayName) {
```
bcarrier: Lets make this package scope for now.

PR #1320, opened 2018-09-12
1012 group seen by user

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1520,6 +1563,45 @@ private CaseDbSchemaVersionNumber updateFromSchema7dot2toSchema8dot0(CaseDbSchem
 		}
 	}
 
+	/**
+	 * Updates a schema version 8.0 database to a schema version 8.1 database.
```
millmanorama: this comment looks like a cut and paste error.  8.0 -> 8.1 should be 8.1-> 8.2

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1520,6 +1563,45 @@ private CaseDbSchemaVersionNumber updateFromSchema7dot2toSchema8dot0(CaseDbSchem
...
+		if (schemaVersion.getMinor() != 1) {
+			return schemaVersion;
+		}
+		Statement statement = connection.createStatement();
+		acquireSingleUserCaseWriteLock();
+		try {
+			// create examiners table
+			if (this.dbType.equals(DbType.SQLITE)) {
+				statement.execute("CREATE TABLE tsk_examiners (examiner_id INTEGER PRIMARY KEY, login_name TEXT NOT NULL, display_name TEXT, UNIQUE(login_name) )");
+			}
```
millmanorama: is there ever a chance of different examiners having the same login name (say local accounts on different machines)?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1883,6 +1966,79 @@ private static String createCaseDataBaseName(String candidateDbName) {
...
+		try {
+			PreparedStatement statement = connection.getPreparedStatement(PREPARED_STATEMENT.SELECT_EXAMINER_BY_ID);
+			statement.clearParameters();
+			statement.setLong(1, id);
+			resultSet = connection.executeQuery(statement);
+			if (resultSet.next()) {
+				return new Examiner(resultSet.getLong("examiner_id"), resultSet.getString("login_name"), resultSet.getString("full_name"));
+			}
+			else {
+				throw new TskCoreException("Error getting examaminer for id = " + id);
```
millmanorama: should this throw an exception or return null/empty optional?

tsk/auto/db_sqlite.cpp
```diff
@@ -372,10 +372,15 @@ int
...
-			"Error creating accounts table: %s\n") ||
+			"Error creating accounts table: %s\n") 
+		||
 		attempt_exec
 		("CREATE TABLE account_relationships (relationship_id INTEGER PRIMARY KEY, account1_id INTEGER NOT NULL, account2_id INTEGER NOT NULL, relationship_source_obj_id INTEGER NOT NULL,  date_time INTEGER, relationship_type INTEGER NOT NULL, data_source_obj_id INTEGER NOT NULL, UNIQUE(account1_id, account2_id, relationship_source_obj_id), FOREIGN KEY(account1_id) REFERENCES accounts(account_id), FOREIGN KEY(account2_id) REFERENCES accounts(account_id), FOREIGN KEY(relationship_source_obj_id) REFERENCES tsk_objects(obj_id), FOREIGN KEY(data_source_obj_id) REFERENCES tsk_objects(obj_id))",
-			"Error creating relationships table: %s\n") ) {
+			"Error creating relationships table: %s\n") 
+		||
+		attempt_exec
+		("CREATE TABLE tsk_examiners (examiner_id INTEGER PRIMARY KEY, login_name TEXT NOT NULL, display_name TEXT, UNIQUE(login_name))",
```
millmanorama: does the schema version need to be update in tsk, it will probably be updated anyways when we merge the timeline branch in also

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
raman-bt: It's certainly possible to have same examiner login ids if they are local accounts and not in domain. 
Hopefully, people sharing cases with each other are not using local accounts though.

Otherwise we may need to get more sophisticated and determine the type of account, domain name or machine name, and prepend that. 

@bcarrier: should we open a JIRA to handle this later?

bcarrier: I don't think we need a jira. We'll deal with it in the future if it comes up.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
raman-bt: I think an Exception is reasonable here, the caller would get the id perhaps from the group_seen table, with foreign key constraint. If we can't find the examiner by id, something is definitely wrong.

PR #1325, opened 2018-09-19
Tl schema fix

bcarrier:


bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -732,27 +758,29 @@ private TimelineEvent addEvent(long time, EventType type, long datasourceID, lon
...
-				sleuthkitCase.fireTSKEvent(new EventAddedEvent(singleEvent));
-				return singleEvent;
 			}
 		} catch (SQLException ex) {
 			throw new TskCoreException("Failed to insert event.", ex); // NON-NLS
 		} finally {
 			sleuthkitCase.releaseSingleUserCaseWriteLock();
 		}
+		
+		sleuthkitCase.fireTSKEvent(new EventAddedEvent(singleEvent));
```
millmanorama: this is something I had noticed and thought was probably bad.

PR #1328, opened 2018-09-20
make new version of test_root that is not from other licensed code.  …

### LINKED ISSUE #1251: potential relicensed (L)GPL code in ext2fs.c test_root
From https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ext2fs.c#L67

```
/** \internal
    test_root - tests to see if a is power of b
    Adapted from E2fsprogs sparse.c
    Super blocks are only in block groups that are powers of 3,5, and 7
    @param a the number being investigated
    @param b the root
    @return 1 if a is a power of b, otherwise 0
*/
static uint8_t
test_root(uint32_t a, uint32_t b)
{
    if (a == 0)
        return 1;
    while (1) {
        if (a == 1)
            return 1;
        if (a % b)
            return 0;
        a = a / b;
    }
}
```

Looking at https://git.busybox.net/busybox/plain/e2fsprogs/ext2fs/sparse.c?h=1_3_stable

```
static int test_root(int a, int b)
{
	if (a == 0)
		return 1;
	while (1) {
		if (a == 1)
			return 1;
		if (a % b)
			return 0;
		a = a / b;
	}
}
```

Docstring says code "adapted" from E2fsprogs sparse.c, looks directly copied.
How is this code "adapted", it looks exactly like the original?

bcarrier:
…Closes #1251

### DIFF
--- a/tsk/fs/ext2fs.c
+++ b/tsk/fs/ext2fs.c
@@ -55,8 +55,6 @@ debug_print_buf(unsigned char *buf, int len)
 
 /** \internal
     test_root - tests to see if a is power of b
-    Adapted from E2fsprogs sparse.c
-    Super blocks are only in block groups that are powers of 3,5, and 7
     @param a the number being investigated
     @param b the root
     @return 1 if a is a power of b, otherwise 0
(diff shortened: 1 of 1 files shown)


APriestman:
Did you want to also update the ext2fs_bg_has_super() method description?

tsk/fs/ext2fs.c
```diff
@@ -55,24 +55,32 @@ debug_print_buf(unsigned char *buf, int len)
...
-            return 0;
-        a = a / b;
+    if (a == 0) {
+        return (b == 0);
+    }
+    else if (b == 0) {
+        return 0;
     }
+    else if (a == 1) {
+        return (b == 1);
```
APriestman: Do we care about the fact that any value for b raised to the zero power would be 1?

bcarrier: Updated

PR #1330, opened 2018-09-25
4187 examiner relationship to tags table

wschaeferB:


wschaeferB:
Need to change examiner_id column type from big serial to something that can be null

PR #1338, opened 2018-10-09
Added a new funciton that gives the EncodedFileOutputStream a mutable…

dannysmyda:
… state, as to cope with the memory leaks in the seven zip jbindings library


dannysmyda:
DEPENDS ON: https://github.com/sleuthkit/autopsy/pull/4199

bcarrier:
Closing this in favor of approach that does not impact public API

PR #1343, opened 2018-10-16
1088: case creation hangs

raman-bt:
- Fixed table creation SQL for event_types and events tables.


raman-bt:
@bcarrier: can you please review and merge this?

PR #1354, opened 2018-10-26
Decmpfs reorganization 

jtsylve:
I'm submitting this PR in advance of the APFS patches.  APFS uses the same compression logic as HFS+, so this patch pulls that logic into it's own file and generalizes it a bit.  

Since so much of this code was moved to another file, it has been my major source of pain with merging upstream patches.


jtsylve:
I don't know how to add the new file to the MSVC build, which is causing the check to fail.

jtsylve:
@bcarrier Any insight on this?

PR #1374, opened 2018-11-24
Fix CVE-2018-19497. 

JordyZomer:
An issue was discovered in The Sleuth Kit (TSK) through 4.6.4.
The "tsk_getu16(hfs->fs_info.endian, &rec_buf[rec_off2])" call in hfs_dir_open_meta_cb in
tsk/fs/hfs_dent.c does not properly check boundaries. This results in
a crash (SEGV on unknown address
READ memory access)
when reading too much in the destination buffer.

this is because the boundary check in hfs_traverse_cat wasn't done properly.

Kind Regards,

Jordy Zomer

### DIFF
--- a/tsk/fs/hfs.c
+++ b/tsk/fs/hfs.c
@@ -961,7 +961,7 @@ hfs_cat_traverse(HFS_INFO * hfs,
                     tsk_error_set_errno(TSK_ERR_FS_GENFS);
                     tsk_error_set_errstr
                         ("hfs_cat_traverse: length of key %d in index node %d too large (%d vs %"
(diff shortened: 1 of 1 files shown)


JordyZomer:
Validated the patch, now it returns:

General file system error (hfs_cat_traverse: length of key 3 in leaf node 1 too large (65537 vs 4096))

On the proof of concept.

JordyZomer:
@rcordovano Any updates? The ticket has been open for 5 days (Sorry for directly mentioning you).

bcarrier:
Thanks for the fix!

PR #1407, opened 2019-01-05
1171: Create IG table schema versions

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -104,6 +149,37 @@ public void createTable(final String tableName, final String tableSchema) throws
...
 
+	/**
+	 * Alters a table with the specified name.
+	 * 
+	 * @param tableName name of the table to alter
+	 * @param alterSQL SQL to alter the table
+	 * 
+	 * @throws TskCoreException 
+	 */
+	public void alterTable(final String tableName, final String alterSQL) throws TskCoreException {
```
rcordovano: I think we need to continue the design of allowing users to execute commands within a transaction. An ALTER table statement executed all by itself is probably less common then a reworking of several aspects of a schema at one time. This probably also argues for allowing doesDoesColumnExist to be done in the context of a transaction as well.

raman-bt: Umm, why does "doesColumnExist"  need a transaction?   I mean it can be called in the context of a transaction that alters table, but it the method itself doesn't need to take a transaction as an argument, right?

rcordovano: I am imagining an upgrade flow something like this:

start a transaction
do first part of the upgrade with its own column exists queries, alter table commands, etc.
do second part of the upgrade with its own column exists queries, alter table commands, etc.
do third part of the upgrade with its own column exists queries, alter table commands, etc.
and so on
commit or rollback transaction

All of the above would be done using the same connection, obtained from the CaseDbTransaction object. Otherwise, you would need to do a bunch of columnExists (shorter name?) calls and make a bunch of boolean locals to remember what you found out, each in a call that opened and closed its own connection. So making it possible to use columnExists within the transaction seems cleaner to me.  

You could get the same effect by allowing a connection to be passed in to columnExists. I suppose we could have used that strategy for the other APIs as well...

PR #1417, opened 2019-01-16
Fix issue with random seeking in sparse files

jtsylve:
`tsk_fs_attr_read` doesn't always reset `byteoffset_toread`.  This causes errors when reading randomly from sparse files.

### DIFF
--- a/tsk/fs/fs_attr.c
+++ b/tsk/fs/fs_attr.c
@@ -1288,6 +1288,7 @@ tsk_fs_attr_read(const TSK_FS_ATTR * a_fs_attr, TSK_OFF_T a_offset,
 
             }
             len_remain -= len_inrun;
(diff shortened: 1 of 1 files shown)


bcarrier:
Hey @jtsylve , this fix looks correct, but it is making me look more at that method and I'm not sure if we should be resetting byteoffset_toread at line 1257 because we use the value in the following 'if' statement.  It seems like we could get rid of line 1257 with your fix in place (and maybe move the 1-line comment down to it).  Thoughts?

jtsylve:
To be honest, I'm not sure.  That whole function is a bit of a mess and could use cleanup/rewriting.  Due to the importance of the function, I only made the change that addressed the issue I was seeing.  I didn't want to break anything else.

PR #1430, opened 2019-02-13
New attribute types added.

dgrove727:


rcordovano:
Use custom attribute types instead.

PR #1433, opened 2019-02-13
If reading into the image cache fails, attempt a read without caching

jtsylve:
This PR attempts a read that bypasses the image cache if reading a cache block fails.  This fixes issues I've seen when reading from some macOS block devices that only let you read from specific addresses.


jtsylve:
IMO, `TSK_IMG_INFO.cache_lock` is poorly named.  In the initial version of this fix, I released the lock before calling `tsk_img_read_no_cache` (because that function wasn't modifying the cache.  This caused a race condition because some of the lower-level image readers expect that the cache lock is held.  I suggest changing the name to `TSK_IMG_INFO.read_lock` as that's essentially what it is acting as.

PR #1438, opened 2019-02-21
4761 NPE fix

APriestman:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -6190,7 +6190,11 @@ public LocalFile addLocalFile(String fileName, String localPath,
 			statement.setLong(13, atime);
 			statement.setLong(14, mtime);
 			statement.setString(15, md5);
-			statement.setByte(16, known.getFileKnownValue());
+			if (known != null) {
```
rcordovano: @APriestman, I am going to accept this PR to get us back on track, but we should document in the param doc for known that null is a valid argument.

PR #1439, opened 2019-02-21
4636 populate acquisition details e01

dannysmyda:


tsk/auto/auto_db.cpp
```diff
@@ -208,13 +209,38 @@ uint8_t
...
        if (ewf_info->md5hash_isset) {
            md5 = ewf_info->md5hash;
        }
+	   if (ewf_info->sha1hash_isset) {
+		   sha1 = ewf_info->sha1hash;
+	   }
+
+       //Need 1MB for libewf read and extra 100 bytes for header name and formatting
+       const size_t buffer_size = 1024100;
+       char result[buffer_size];
```
APriestman: Looking at this again, 1024100 might be a bit large for the stack. I'm not having problems but it might be safer to do a malloc here.

PR #1449, opened 2019-02-28
1188 Web cache artifact

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -1189,7 +1189,18 @@ public int hashCode() {
 		 * Indicates an person's address filled in a web form
 		 */
 		TSK_WEB_FORM_ADDRESS (50, "TSK_WEB_FORM_ADDRESSES ",  //NON-NLS
-				bundle.getString("BlackboardArtifact.tskWebFormAddresses.text"));
+				bundle.getString("BlackboardArtifact.tskWebFormAddresses.text")),
```
kellykelly3: Should you be using the @message annotation instead of adding to the .properties file?

raman-bt: Not in this case.  For consistency, we will keep all the artifact definition strings in the properties file.  I discussed it with Richard, and he and agrees.

PR #1450, opened 2019-03-01
4736 associate cache with case

APriestman:


rcordovano:
@APriestman, there are a couple of things Codacy flagged that can be easily fixed, thanks!

PR #1454, opened 2019-03-06
add "User Created" event type

millmanorama:


millmanorama:
I replaced this with https://github.com/sleuthkit/sleuthkit/pull/1457  which has less noise from merging develop

PR #1466, opened 2019-03-19
4819 add ability to find files based on attributes

jkho:


tools/autotools/TskFindFiles.cpp
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Print errors as they are encountered
+ */
+uint8_t TskFindFiles::handleError() {
+    fprintf(stderr, "%s", tsk_error_get());
+    return 0;
+}
+
+/**
+* Process a file. If the file contains an extension is specified in the LogicalImagerConfig,
```
APriestman: Probably missing a "which" after "extension

tools/autotools/TskFindFiles.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    char *extension = PathFindExtensionA(fs_file->name->name);
+    if (extension[0] == '.') {
+        // skip the leading dot
+        extension = &extension[1];
+    }
+    if (m_logicialImagerConfig->hasExtension(extension)) {
+        //fprintf(stderr, "processFile: fs_file->name->name=%s\n", fs_file->name->name);
+        //fprintf(stderr, "processFile: path=%s\n", path);
+        //fprintf(stderr, "extension %s\n", extension);
```
APriestman: Should probably delete the debug statements

tools/autotools/tsk_logical_imager.cpp
```diff
@@ -635,15 +627,31 @@ BOOL getDriveToProcess(string& driveToProcess) {
...
+    tsk_fprintf(stderr, "\t-c configPath: The configuration file\n");
+    tsk_fprintf(stderr, "\t-v: verbose output to stderr\n");
+    tsk_fprintf(stderr, "\t-V: Print version\n");
+    exit(1);
+}
+
+int
+main(int argc, char **argv1)
+{
+	TSK_IMG_INFO *img;
```
APriestman: It looks like there are tabs instead of spaces in this file.

win32/tsk_logical_imager/config.txt
```diff
@@ -0,0 +1,4 @@
+jpg
```
APriestman: Should this file be checked in? I see it's also part of the project. @bcarrier was this intended?

bcarrier: Long-term, no it shoudl not be checked in here.  The config file is going to evolve with each step. I don't have a strong opinion about checking it in for now or not.  I'm assuming that when this is done, we'll have a sample config file though, but not in the win32 folder.

jkho: I'll delete it.

tools/autotools/tsk_logical_imager.cpp
jkho: sorry. Not probably cut & paste. Untabify will be use from now on :-)

PR #1469, opened 2019-03-25
1218 back\forward code for Communications window

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/CommunicationsFilter.java
```diff
@@ -145,21 +145,24 @@ public String getSQL(CommunicationsManager commsManager) {
 		private final long startDate;
 		private final long endDate;
 		private static final long SECS_PER_DAY = 86400;
+		private boolean startDateEnabled = false;
```
millmanorama: I wouldn't initialize this explicitly since the value is just overwritten by the constructor.

PR #1473, opened 2019-03-26
4882: DrawableDB init error.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -144,6 +143,78 @@ public boolean columnExists(String tableName, String columnName, CaseDbTransacti
...
+		boolean tableExists = false;
+        Statement statement = null;
+        try {
+			CaseDbConnection connection = transaction.getConnection();
+			statement = connection.createStatement();
+			if (DbType.SQLITE == tskDB.getDatabaseType()) {
+				ResultSet tableQueryResults = statement.executeQuery("SELECT name FROM sqlite_master WHERE type='table'");  //NON-NLS
+				while (tableQueryResults.next()) {
+					if (tableQueryResults.getString("name").equalsIgnoreCase(tableName)) { //NON-NLS
+						tableExists = true;
```
wschaeferB: I think it would become a bit simpler logic if instead of querying for "SELECT name FROM sqlite_master WHERE type='table'"  we instead queried for ""SELECT name FROM sqlite_master WHERE type='table' and name = '%s'" with %s being the table name this would allow us to remove the looping through results and have the query be the only thing different about the SQLite and Postgres logic, as both would be checking if results exists to determine if the table was present.

But I might be missing something.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -144,6 +143,78 @@ public boolean columnExists(String tableName, String columnName, CaseDbTransacti
...
+						break;
+					}
+				}
+				tableQueryResults.close();
+			}
+			else {
+				String tableInfoQueryTemplate = "SELECT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name='%s')";  //NON-NLS	
+				ResultSet resultSet = statement.executeQuery(String.format(tableInfoQueryTemplate, tableName.toLowerCase()));
+				if (resultSet.next()) {
```
wschaeferB: The SELECT EXISTS I don't think really buys us anything, since if the result set is empty it doesn't exist. So I think we could drop the outer query and just say if (resultSet.next()){ tableExists = true; } which would be a check that worked with both this and the proposed query above.

But I am perhaps missing some detail about the queries that makes this a better solution.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -144,6 +143,78 @@ public boolean columnExists(String tableName, String columnName, CaseDbTransacti
...
+					if (tableQueryResults.getString("name").equalsIgnoreCase(tableName)) { //NON-NLS
+						tableExists = true;
+						break;
+					}
+				}
+				tableQueryResults.close();
+			}
+			else {
+				String tableInfoQueryTemplate = "SELECT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name='%s')";  //NON-NLS	
+				ResultSet resultSet = statement.executeQuery(String.format(tableInfoQueryTemplate, tableName.toLowerCase()));
```
wschaeferB: resultSet doesn't get closed.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -144,6 +143,78 @@ public boolean columnExists(String tableName, String columnName, CaseDbTransacti
...
+				if (resultSet.next()) {
+					tableExists = resultSet.getBoolean(1);
+				}
+			}
+        } 
+		catch (SQLException ex) {
```
wschaeferB: When exceptions are thrown things are not necessary cleaned up correctly.

PR #1477, opened 2019-03-29
"modify ext4 journal descriptor table structure"

hoyoi05:
The structure of the current sleuthkit version of the ext4 journal descriptor table is problematic.

The value of sizeof (ext2fs_journ_dentry) on lines 463 and 468 of the ext2fs_journal.c file is currently 8 bytes, but the size of the dentry(tag) in the ext4 journal descriptor table should be 16 or 32 bytes.

### DIFF
--- a/tsk/fs/tsk_ext2fs.h
+++ b/tsk/fs/tsk_ext2fs.h
@@ -614,6 +614,8 @@ extern "C" {
     typedef struct {
         uint8_t fs_blk[4];
         uint8_t flag[4];
+        uint8_t fs_blk_hi[4];
+        uint8_t checksum[4];
     } ext2fs_journ_dentry;
 
 


simsong:
I would love to get some other eyes on this.

simsong:
I think that we need an `ext4` disk image to test this. Do you have one?

simsong:
> The structure of the current sleuthkit version of the ext4 journal descriptor table is problematic.
> 
> The value of sizeof (ext2fs_journ_dentry) on lines 463 and 468 of the ext2fs_journal.c file is currently 8 bytes, but the size of the dentry(tag) in the ext4 journal descriptor table should be 16 or 32 bytes.

We also need a citation for this.

simsong:
@Taha-Ebrahim  - can you look at this?
- Do we have a test image for ext4? 
- Does this break output? 
- Can you find a reference in the ext4 documentation that our definition for `ext2fs_journ_dentry` is wrong?

Taha-Ebrahim:
> The structure of the current sleuthkit version of the ext4 journal descriptor table is problematic.
> 
> The value of sizeof (ext2fs_journ_dentry) on lines 463 and 468 of the ext2fs_journal.c file is currently 8 bytes, but the size of the dentry(tag) in the ext4 journal descriptor table should be 16 or 32 bytes.

What I could find was that the size of the journal descriptor table depends on whether a checksum is set. See [ext4 documentation](https://docs.kernel.org/filesystems/ext4/journal.html#descriptor-block) section 3.6.5 which states the following: 

> If JBD2_FEATURE_INCOMPAT_CSUM_V3 is set, the journal block tag is defined as struct journal_block_tag3_s, which looks like the following. The size is 16 or 32 bytes.

> If JBD2_FEATURE_INCOMPAT_CSUM_V3 is NOT set, the journal block tag is defined as struct journal_block_tag_s, which looks like the following. The size is 8, 12, 24, or 28 bytes

This is the only documentation that I could find that seems to relate to the sizes journal descriptor block. Generally, it seems that checksums have become standard with ext4 and so the issue arises. 

The PR also seems to make sense based on the documentation in 3.6.5:

Screenshot 2025-08-20 at 15 43 16

simsong:
So it seems to be really need an ext4 disk image. Is the sizeof this
structure used anywhere in our existing source code?

Taha-Ebrahim:
> So it seems to be really need an ext4 disk image. Is the sizeof this structure used anywhere in our existing source code?

It seems it is used in the `tsk/fs/ext2fs_journal.cpp` file, specifically within the `ext2fs_jentry_walk` and `ext2fs_jblk_walk` functions. So far I have not found an issue when just testing with normal CLI functions on a tiny ext4 image that I created locally but I'm sure it's possible that issues will crop up.

simsong:
can you zip the tiny image and attach it to this ticket?

Taha-Ebrahim:
Yes, let me upload it here!

[ext4.img.zip](https://github.com/user-attachments/files/22022739/ext4.img.zip)

simsong:
> Yes, let me upload it here!
> 
> [ext4.img.zip](https://github.com/user-attachments/files/22022739/ext4.img.zip)

Can you make a few changes to this?
```
simsong@Seasons-2 ~ % fls ext4.img
d/d 11:	lost+found
r/r 12:	hello.txt
V/V 2561:	$OrphanFiles
simsong@Seasons-2 ~ %
```
It would be good if this image had the following additions:
* A symlink `hello` that points to `hello.txt`
* A directory `testdir`
* A new file `testdir/hello2.txt` that contains the repeating pattern 'This is line xxxx' where xxxx goes from 1 to 10,000.  (This file will compress well and will take more than a single block.)
* A new symlink `hello2.soft` that links to `testdir/hello2.txt`
* A hardlink `testdir/hello2.hard` that is linked to `testdir/hello2.txt`
* An annoying link `testdir/up` that links to `..`
* A unix device `device1` created with mknod 10,20 or something.
* A unix file system named pipe called `fifo1` made with [mkfifo](https://man7.org/linux/man-pages/man1/mkfifo.1.html)

This will give us more things that we can test with the disk image.

Thanks!

Taha-Ebrahim:
Yes, making the changes now! Might take a little while to update now that school has begun!

simsong:
> Yes, making the changes now! Might take a little while to update now that school has begun!

Better you than me!

PR #1481, opened 2019-04-02
4902 add support for paths that are relative to user folder

jkho:


tools/autotools/LogicalImagerPathRule.cpp
```diff
@@ -38,10 +76,14 @@ bool LogicalImagerPathRule::matches(TSK_FS_FILE * /*fs_file*/, const char *path)
     if (path == NULL)
         return false;
 
-    std::string lowercasePath = LogicalImagerRuleBase::toLower(path);
+    const std::string lowercasePath = TskHelper::toLower(path);
 
     for (auto it = std::begin(m_paths); it != std::end(m_paths); ++it) {
-        if (lowercasePath.find(*it) != std::string::npos) {
+        if (it->find(lowerCaseUserFolder) != std::string::npos) {
```
bcarrier: Seems like this should be a bit more restrictive and ensure that 'it' startsWith lowerCaseUserFolder.  not just that it contains the string.

PR #1493, opened 2019-04-16
Tl filter fixes

millmanorama:
this is part of the fixes for an unreported bug where adding filters resulted in duplicated filters in the GUI

please take https://github.com/sleuthkit/sleuthkit/pull/1492 first


rcordovano:
@millmanorama, please merge the main feature branch into the source branch for this PR to reduce the diffs, thanks.

rcordovano:
@kellykelly3, this PR is ready for your review now, thanks!

bindings/java/src/org/sleuthkit/datamodel/timeline/TimelineFilter.java
```diff
@@ -410,10 +419,66 @@ public RootFilter copyOf() {
...
+			hash = 17 * hash + Objects.hashCode(this.hashFilter);
+			hash = 17 * hash + Objects.hashCode(this.textFilter);
+			hash = 17 * hash + Objects.hashCode(this.typeFilter);
+			hash = 17 * hash + Objects.hashCode(this.dataSourcesFilter);
+			hash = 17 * hash + Objects.hashCode(this.fileTypesFilter);
+			hash = 17 * hash + Objects.hashCode(this.namedSubFilters);
+			return hash;
+		}
+
+		@Override
```
kellykelly3: Is there away to make hashCode and equals more dynamic so that if other xxxFilter are added that these two functions may not have to be revisited?  Can you use nameSubFilters to loop through the filters so that you don't have to call out each one individually?

PR #1513, opened 2019-06-05
5044: occasional exception during IG db initialization

raman-bt:
- SQLite sometimes lies in columnExists() - ignore exception in alterTable, when trying to add a column that already exists.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
@@ -316,6 +316,15 @@ public final class CaseDbAccessManager {
 			statement = connection.createStatement();
 			statement.execute(sql);
 		} catch (SQLException ex) {
+			// SQLite occasionally returns false for columnExists() if a table was just created with that column
+			// leading to "duplicate column name" exception.
+			// We ignore this exception
+			if (DbType.SQLITE == tskDB.getDatabaseType() &&
+					alterSQL.toLowerCase().contains("add column") &&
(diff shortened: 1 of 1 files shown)


bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -316,6 +316,15 @@ public void alterTable(final String tableName, final String alterSQL, final Case
 			statement = connection.createStatement();
 			statement.execute(sql);
 		} catch (SQLException ex) {
+			// SQLite occasionally returns false for columnExists() if a table was just created with that column
+			// leading to "duplicate column name" exception.
+			// We ignore this exception
+			if (DbType.SQLITE == tskDB.getDatabaseType() &&
+					alterSQL.toLowerCase().contains("add column") &&
+					ex.getMessage().toLowerCase().contains("duplicate column name")) {
+				logger.log(Level.WARNING, String.format("Column being added by SQL = %s already exists in table %s", alterSQL, tableName));
```
rcordovano: I don't think that it is critical that we log this exception in this special case. It is not really actionable by either us or the user. Do you agree or disagree?

raman-bt: I don't have a strong opinion, I'm ok with getting rid of the log message here.

rcordovano: Okay, let's leave it after all. It might be interesting if we ever see it.

PR #1518, opened 2019-06-07
2414 remove public data source deletion api from sleuthkitcase

markmckinnon:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -7512,6 +7551,35 @@ public void setImagePaths(long obj_id, List<String> paths) throws TskCoreExcepti
 		}
 	}
 
+	private void deleteDataSource(long obj_id) throws TskCoreException {
```
dannysmyda: This method should have a comment indicating that it is intentionally private in order to deter direct invocation and that the decision was to reflect this method in order to circumnavigate public API. Otherwise, the lack of background info and warnings from Netbeans makes it look like it was missed in a refactor.

markmckinnon: Added comment to state private keeping out of public API

PR #1521, opened 2019-06-11
Retry if ReadFile fails with ERROR_NOT_READY

zweger:
This is a fix for some strange behavior I observed running on Windows 7. If you have an open file handle to a volume device (`\\?\Volume{guid}`), and that volume is formatted while the handle is still open, then all subsequent attempts to read from that handle will fail with ERROR_NOT_READY (21). If the file handle is reopened, then the volume can be read from again.

This patch adjusts the behavior of `raw_read_segment` so that, upon ReadFile failing with ERROR_NOT_READY, will attempt to reopen the file handle and retry the read.

### DIFF
--- a/tsk/img/raw.c
+++ b/tsk/img/raw.c
@@ -39,8 +39,89 @@
 #define S_IFDIR __S_IFDIR
 #endif
 
+/**
+ * \internal
+ * Close the file handle held by a particular split disk image
+ *
+ * @param raw_info Disk image containing the split image
+ * @param cimg split image to close
+ */
+static void
+raw_close_segment(IMG_RAW_INFO * raw_info, IMG_SPLIT_CACHE * cimg) {
+    if (cimg->fd != 0) {
+        if (tsk_verbose) {
+            tsk_fprintf(stderr,
+                "raw_close_segment: closing file %" PRIttocTSK "\n",
+                raw_info->img_info.images[cimg->image]);
+        }
+#ifdef TSK_WIN32
+        CloseHandle(cimg->fd);
+#else
+        close(cimg->fd);
+#endif
+        raw_info->cptr[cimg->image] = -1;
+    }
+}
 
-/** 
+/**
+ * \internal
+ * Open a split disk image
+ *
+ * @param raw_info Disk image to be opened
+ * @param slot Cache slot for the segment to be opened into
+ * @param idx Index of the disk image in the set to read from
+ *
+ * @return NULL on error
+ */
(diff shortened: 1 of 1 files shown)


bcarrier:
If the volume has been reformatted, that's basically a moral equivalent to a new volume/file system. It seems like knowing that and opening it again would be logical. 

I'd rather know that something happened on the device that I had opened and reset accordingly.  I'm concerned about data in other layers (such as caches or file system sizes) that would no longer be valid. 

What was your use case when you had this issue?

zweger:
Funny you mention the file system size - it's also possible to resize a volume while there's already a handle open against it. No error is triggered on subsequent reads (unless you're reading past the end of the shrunk image, I suppose). If the volume size grew, then most applications probably wouldn't realize it.

In my application I'm using a long-lived `TSK_IMG_INFO` to monitor a live device on a (Windows) system over a period of some time. I'd like to be able to process the device possibly multiple times, or possibly with some delay between opening the image and processing it.

Formatting the volume does invalidate the filesystem, but not necessarily the disk/volume (formatting a volume doesn't change the device name). 

Caching is an issue in this use case - not just because the device changes when it's formatted, but also because it's a live system and always changing. So, I've separately disabled caching in my application to avoid getting stale data.

Your idea that library users should reopen the image when the underlying handle has become stale is reasonable, as that would also pick up any changes to the image size and clear the cache. Do you think it would be preferable to add something like a `TSK_ERR_IMG_READ_REOPEN` error code to indicate the image should be reopened in situations like this?

simsong:
What do we want to do with this? @bcarrier ? @jayaramcs ?

zweger:
I have half a mind to close this PR - in the 5 years since I first opened it, my use case for it has gone away.

I think there may be _something_ here worth tweaking, to do with error handling/reporting/etc., but I don't think the retry loop in this PR is it. If anyone else is inclined to take a look, all of the oddball Windows behaviors I've mentioned above are still to the best of my knowledge correct.

simsong:
I think that we will close this and allow the user to retry at the command line.

simsong:
> I have half a mind to close this PR - in the 5 years since I first opened it, my use case for it has gone away.
> 
> I think there may be _something_ here worth tweaking, to do with error handling/reporting/etc., but I don't think the retry loop in this PR is it. If anyone else is inclined to take a look, all of the oddball Windows behaviors I've mentioned above are still to the best of my knowledge correct.

Thanks for the response!

PR #1537, opened 2019-06-20
5230 add version number to logical imager config file

jkho:


tools/logicalimager/LogicalImagerConfiguration.h
```diff
@@ -38,11 +38,15 @@ class LogicalImagerConfiguration
...
 
 private:
     LogicalImagerConfiguration(const LogicalImagerConfiguration &) = delete;
 
     std::vector<LogicalImagerRuleSet *> m_ruleSets;
     bool m_finalizeImageWriter = false;
+    std::string m_version;
     LogicalImagerRuleSet::matchCallback m_callbackFunc;
+
+    const std::string m_DefaultVersion = std::string("1.0");
```
APriestman: "m_CurrentVersion" seems more accurate here

tools/logicalimager/Version.h
```diff
@@ -0,0 +1,68 @@
...
+        str = std::to_string(m_major) + ".x";
+        return str;
+    }
+
+    bool operator < (const Version& other) const {
+        if (m_major < other.m_major)
+            return true;
+        if (m_minor < other.m_minor)
+            return true;
+        return false;
```
APriestman: This logic doesn't look correct. If we're comparing 3.1 with 2.2 (other), it looks like it will return true.

PR #1538, opened 2019-06-21
Use TSK_OFF_T for file size

jkho:


tools/logicalimager/LogicalImagerRuleSet.cpp
```diff
@@ -53,8 +53,8 @@ time_t stringToTimet(const std::string &datetimeStr) {
     return time;
 }
 
-int getPositiveInt(const std::string &key, nlohmann::json ruleJson) {
-    int size;
+TSK_OFF_T getPositiveInt(const std::string &key, nlohmann::json ruleJson) {
```
APriestman: The date rule also calls this and is now generating a warning about the conversion from TSK_OFF_T to int

jkho:
I pushed a change. minDays should be int, not TSK_OFF_T

PR #1542, opened 2019-06-26
Fix NPE

jkho:


tsk/img/img_writer.cpp
```diff
@@ -206,8 +206,22 @@ static TSK_RETVAL_ENUM addNewBlock(TSK_IMG_WRITER* writer, TSK_OFF_T addr, char
     writer->blockToSectorNumber[blockNum] = uint32_t(writer->nextDataOffset / VHD_SECTOR_SIZE);
 
     char * fullBuffer = (char *)tsk_malloc(writer->blockSize * sizeof(char));
+    if (fullBuffer == NULL) {
+        return TSK_ERR;
+    }
+
     char * sectorBitmap = (char *)tsk_malloc(writer->sectorBitmapLength * sizeof(char));
+    if (fullBuffer == NULL) {
```
APriestman: Is this a cut and paste error? You should be testing sectorBitmap I think

tsk/img/img_writer.cpp
```diff
@@ -206,8 +206,22 @@ static TSK_RETVAL_ENUM addNewBlock(TSK_IMG_WRITER* writer, TSK_OFF_T addr, char
...
+        free(fullBuffer);
+        return TSK_ERR;
+    }
+
     unsigned char * completedSectors = (unsigned char *)tsk_malloc(((writer->sectorBitmapArrayLength) * sizeof(char)));
+    if (fullBuffer == NULL) {
```
APriestman: Another cut and paste error?

tsk/img/img_writer.cpp
```diff
@@ -234,6 +248,8 @@ static TSK_RETVAL_ENUM addNewBlock(TSK_IMG_WRITER* writer, TSK_OFF_T addr, char
 
     /* Write the new offset to the BAT */
     if (TSK_OK != seekToOffset(writer, writer->batOffset + 4 * blockNum)) {
+        free(fullBuffer);
+        free(sectorBitmap);
```
APriestman: Do you need to also free completedSectors here?

jkho: The original code never frees completedSectors, because it is assigned outside
`writer->blockToSectorBitmap[blockNum] = completedSectors;`

I am keeping the original logic. Is this ok?

APriestman: Yeah that's fine - was just making sure it wasn't an oversight.

tsk/img/img_writer.cpp
jkho: yes

tsk/img/img_writer.cpp
jkho: yes, I pushed a change.

PR #1544, opened 2019-06-26
Update tsk_logical_imager.cpp

jkho:


tools/logicalimager/tsk_logical_imager.cpp
```diff
@@ -42,6 +42,12 @@ static TSK_TCHAR *progname;
 
 static std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
 
+static void pressAnyKeyToExit(int code) {
+    std::cout << std::endl << "Press any key to exit";
+    (void)getchar();
```
APriestman: It seems like I have to press enter, not any character

jkho: Fixed

PR #1551, opened 2019-06-27
5247 logical imager needs to show basic progress

jkho:


jkho:
I don't understand the codacy error

APriestman:
Maybe try changing the second %zi to %zu ?

PR #1556, opened 2019-06-27
5285 logical imager vhd seems to get corrupted when we run out of space

jkho:


tsk/img/img_writer.h
```diff
@@ -36,6 +36,7 @@ extern "C" {
         int finishProgress;
         int cancelFinish;
         int inFinalizeImageWriter;
+        int writeError;
```
bcarrier: Can you add a 1-line comment about each of these flags and when they are set?

tsk/img/img_writer.cpp
```diff
@@ -190,11 +190,42 @@ static TSK_RETVAL_ENUM addToExistingBlock(TSK_IMG_WRITER* writer, TSK_OFF_T addr
...
+    if (TSK_OK == seekToOffset(writer, lastPosition)) {
+        writeFooter(writer); // write footer at lastPosition prior to WriteFile error
+    }
+    else {
+        int lastError = GetLastError();
+        tsk_error_reset();
+        tsk_error_set_errno(TSK_ERR_IMG_WRITE);
+        tsk_error_set_errstr("addNewBlock: error seekToOffset",
+            lastError);
+    }
```
bcarrier: We should return an error here so that someone knows the error codes are set.

PR #1560, opened 2019-07-02
Update img_open.cpp

jkho:


tsk/img/img_open.cpp
```diff
@@ -256,6 +268,14 @@ tsk_img_open(int num_img,
         return NULL;
     }
 
+    /* check if img_info is good */
+    if (img_info == NULL) {
+        tsk_error_reset();
```
APriestman: I think just return null here - the error isn't necessarily "unsupported type", and the image open methods may have already set a more specific error.

PR #1565, opened 2019-07-11
add promptBeforeExit default is true in case it is missing from confi…

jkho:
…g file


tools/logicalimager/tsk_logical_imager.cpp
```diff
@@ -49,6 +49,14 @@ static void pressAnyKeyToExit(int code) {
     exit(code);
 }
 
+static void mayBePressAnyKeyToExit(int code, bool promptBeforeExit) {
```
APriestman: Rename this to something like "handleExit"

PR #1568, opened 2019-07-17
1286 - Moving TimelineEvent creations from UI to TimelineManager

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TimelineEvent.java
```diff
@@ -86,7 +86,7 @@
 	 */
 	// @@@ This should be package scope, but it is currently being used in DetailsViewModel.
```
raman-bt: I guess this  scope related comment is now obsolete?

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -1023,6 +1023,75 @@ private static int booleanToInt(boolean value) {
...
+						resultSet.getString("med_description"), // NON-NLS
+						resultSet.getString("short_description"), // NON-NLS
+						resultSet.getInt("hash_hit") != 0, //NON-NLS
+						resultSet.getInt("tagged") != 0);
+				
+				events.add(event);
+            }
+			
+		} catch (SQLException ex) {
+			throw new TskCoreException("Error getting count of events from db: " + querySql, ex); // NON-NLS
```
raman-bt: we are getting the events themselves, not just the count here, right? Should fix the log message in that case.

PR #1569, opened 2019-07-18
CT-2984 (handle max key/val name lengths properly)

crayy8:
max name lengths represent # of characters. We were not handling ascii vs utf16 names properly.

### DIFF
--- a/rejistry++/src/NKRecord.cpp
+++ b/rejistry++/src/NKRecord.cpp
@@ -58,8 +58,8 @@ namespace Rejistry {
         int32_t offset = (int32_t)getDWord(CLASSNAME_OFFSET_OFFSET);
         uint16_t length = getWord(CLASSNAME_LENGTH_OFFSET);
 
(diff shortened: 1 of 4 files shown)


rejistry++/src/NKRecord.cpp
```diff
@@ -58,6 +58,8 @@ namespace Rejistry {
         int32_t offset = (int32_t)getDWord(CLASSNAME_OFFSET_OFFSET);
         uint16_t length = getWord(CLASSNAME_LENGTH_OFFSET);
 
+		// Not sure why we are performing this check. I haven't found any documentation on
```
APriestman: Please replace tabs with spaces here and in the rest of the PR

crayy8: Updated in https://github.com/sleuthkit/sleuthkit/pull/1569/commits/3cbc5c2ba0a658fff60a63f90929c3fed162766a

rcordovano:
@APriestman, do you approve this PR after the latest commit by cray?

PR #1574, opened 2019-07-22
1139 & 1308 - Fixed descriptions for emails, messages, and call logs.

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TimelineEventType.java
```diff
@@ -44,6 +44,8 @@
  number of levels and types. 
  */
 public interface TimelineEventType extends Comparable<TimelineEventType> {
+	
+	static final int EMAIL_FULL_DESCRIPTION_LENGTH_MAX = 150;
```
rcordovano: This should be private.

kellykelly3: It cannot be.  Java doesn't appear to allow static private members of an interface.  I could just hard code the value (used at line 309).

rcordovano: It's fine as it is, thanks. Learn something new everyday!

PR #1578, opened 2019-07-23
1310 Added getDBSchemaCreationVersion

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -906,8 +908,11 @@ private void updateDatabaseSchema(String dbPath) throws Exception {
 				dbSchemaVersion = updateFromSchema7dot2toSchema8dot0(dbSchemaVersion, connection);
 				dbSchemaVersion = updateFromSchema8dot0toSchema8dot1(dbSchemaVersion, connection);
 				dbSchemaVersion = updateFromSchema8dot1toSchema8dot2(dbSchemaVersion, connection);
+				dbSchemaVersion = updateFromSchema8dot2toSchema8dot3(dbSchemaVersion, connection);
 				statement = connection.createStatement();
 				connection.executeUpdate(statement, "UPDATE tsk_db_info SET schema_ver = " + dbSchemaVersion.getMajor() + ", schema_minor_ver = " + dbSchemaVersion.getMinor()); //NON-NLS
+				connection.executeUpdate(statement, "UPDATE tsk_db_info_extended SET value = " + dbSchemaVersion.getMajor() + " WHERE name = 'SCHEMA_MAJOR_VERSION'"); //NON-NLS
```
raman-bt: There are static final variables defined for these strings ('SCHEMA_MAJOR_VERSION' and 'SCHEMA_MINOR_VERSION').  We should use those, instead of hard coding them here.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -923,6 +928,41 @@ private void updateDatabaseSchema(String dbPath) throws Exception {
...
+		ResultSet resultSet = null;
+		String createdSchemaMajorVersion = "0";
+		String createdSchemaMinorVersion = "0";
+		acquireSingleUserCaseReadLock();
+		try {
+			statement = connection.createStatement();
+			resultSet = connection.executeQuery(statement, "SELECT name, value FROM tsk_db_info_extended");
+			while(resultSet.next()) {
+				String name = resultSet.getString("name");
+				if (name.equals(CREATION_SCHEMA_MAJOR_VERSION_KEY) || name.equals("CREATED_SCHEMA_MAJOR_VERSION")){
```
raman-bt: we should define static final variables for 'CREATED_SCHEMA_MAJOR_VERSION' & 'CREATED_SCHEMA_MINOR_VERSION' and use those, instead of hard-coding the literals here.

Update: Having looked at it further down, if CREATED_* names were just accidental and we are fixing them back to CREATION_*, then I'm fine with hard coding them.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1773,6 +1813,131 @@ private CaseDbSchemaVersionNumber updateFromSchema8dot1toSchema8dot2(CaseDbSchem
...
+					break;
+				case SQLITE:
+					primaryKeyType = "INTEGER";
+					break;
+				default:
+					throw new TskCoreException("Unsupported data base type: " + getDatabaseType().toString());
+			}
+			
+			
+			try{
```
raman-bt: Schema 8.2 already creates a table tsk_event_types, and adds these 7 rows.  Does that need to be dropped first? Or does this need an "IF NOT EXISTS" clause?  

I can't tell how this tsk_event_types table differs from one in 8.2 (i.e. why re-create it all ?)

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
kellykelly3: That's why I hard coded them.  :-)

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
kellykelly3: I didn't know about the IF NOT EXISTS, I will added that and it might fix the bug you found.

Schema 8.2 did not include the tsk_event_types table.

The tsk_event_types table is only created if you upgraded from 8.1 to 8.2.  The code to add the tsk_event_types table was not added to the cpp code for the 8.2.

Therefore when you upgrade from 8.2 to 8.3 the table may or may not exist.

PR #1584, opened 2019-07-30
Update max key name size

crayy8:
Max key name size can be larger than what Microsoft docs state.

This is a minor update related to changes made in https://github.com/sleuthkit/sleuthkit/pull/1569.


crayy8:
@APriestman I just realized the data type was too small for the new value we wanted so I've updated that.

PR #1590, opened 2019-07-31
Add check for FAT device

jkho:


tools/logicalimager/tsk_logical_imager.cpp
```diff
@@ -672,7 +672,7 @@ static BOOL getPhysicalDrives(std::vector<std::wstring> &phyiscalDrives) {
             }
         }
     } else {
-        consoleOutput(stderr, "QueryDosDevice() return error: %d\n", GetLastError());
+        fprintf(stderr, "QueryDosDevice() return error: %d\n", GetLastError());
```
APriestman: Change all these back to consoleOutput?

jkho: When these functions are called, the output directory and the console.txt file are not created yet.

APriestman: If you add a null check to logOutputToFile, you could always call consoleOutput instead of having to keep track of whether the code executes before or after the log file is created.

tools/logicalimager/tsk_logical_imager.cpp
```diff
@@ -864,6 +864,15 @@ static bool hasTskLogicalImager(const TSK_TCHAR *image) {
...
                 // found it
                 result = true;
+                TSK_FS_INFO *fsInfo = *fsListIter;
+                TSK_FS_TYPE_ENUM fileSystemType = fsInfo->ftype;
+                if (fileSystemType == TSK_FS_TYPE_FAT12 ||
+                    fileSystemType == TSK_FS_TYPE_FAT16 ||
+                    fileSystemType == TSK_FS_TYPE_FAT32 ||
+                    fileSystemType == TSK_FS_TYPE_FAT_DETECT) {
+                    consoleOutput(stderr, "Error: Writing to FAT device is not supported.");
+                    pressAnyKeyToExit(1);
```
APriestman: This I think should call handleExit. To avoid confusion, I think you should delete pressAnyKeyToExit entirely and fold it into handleExit.

tools/logicalimager/tsk_logical_imager.cpp
APriestman: Change this and the other fprints back to consoleOutput now that it does the null check?

PR #1601, opened 2019-08-13
5375 allow logical imager to write individual files and not to a vhd

jkho:


tools/logicalimager/tsk_logical_imager.cpp
```diff
@@ -1269,15 +1308,25 @@ main(int argc, char **argv1)
...
+        if (createVHD) {
+            if (img->itype == TSK_IMG_TYPE_RAW) {
+                if (tsk_img_writer_create(img, (TSK_TCHAR *)outputFileNameW.c_str()) == TSK_ERR) {
+                    tsk_error_print(stderr);
+                    consoleOutput(stderr, "Failed to initialize VHD writer\n");
+                    handleExit(1);
+                }
+            }
+            else {
+                consoleOutput(stderr, "Image is not a RAW image, VHD will not be created\n");
```
APriestman: Would it make more sense to default to extract files if a VHD can not be made?

PR #1610, opened 2019-08-20
Added flag to enable c++11 features so build won't fail on OSX

ethteck:
I started suddenly seeing build issues on my Mac, presumably after some update, that hinted that the compiler was not using C++11 features. 

```
db_sqlite.cpp:1009:16: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
    for (const auto entry : timeMap)
               ^
db_sqlite.cpp:1009:27: warning: range-based for loop is a C++11 extension [-Wc++11-extensions]
    for (const auto entry : timeMap)
                          ^
db_sqlite.cpp:1228:41: error: non-aggregate type 'const std::map<int64_t, time_t>' (aka 'const map<long long, long>') cannot be initialized with an initializer list
        const std::map<int64_t, time_t> timeMap = {
                                        ^         ~
```

I found that by adding this flag, I was able to solve the problem. I assume a default setting in an XCode tool changed at some point recently, but aside from that possibility, I'm not sure why this started happening.


esaunders:
I've also been looking into this recently and I believe the solution is to use the autoconf AX_CXX_COMPILE_STDCXX (or the AX_CXX_COMPILE_STDCXX_11 version) macro which will do the right thing based on the particular compiler being used.
See https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html (and/or https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx_11.html).

Aside: unless you are using a really really old compiler the preferred compiler flag would be "-std=c++11". (See https://stackoverflow.com/questions/24589105/what-is-the-difference-between-std-c0x-and-std-c11)

ethteck:
I tried using those actually, but I was seeing circular dependency errors, and I'm not familiar enough with this framework to really figure out what was going on. But I agree that's the best course of action.

rcordovano:
This PR is superseded by the work in https://github.com/sleuthkit/sleuthkit/pull/1611.

PR #1612, opened 2019-08-22
5394: defined new account types for mobile messaging & social media a…

raman-bt:
…pps.


raman-bt:
@bcarrier: could you please create a app-db-parser branch for Sleuthkit and redirect this PR to that branch.

PR #1613, opened 2019-08-22
1324 - addFileSystemFile function

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -5743,6 +5744,115 @@ public FileSystem addFileSystem(long parentObjId, long imgOffset, TskData.TSK_FS
...
+			statement.setLong(7, metaAddr);											// meta_addr
+			statement.setInt(8, metaSeq);											// meta_addr
+			statement.setShort(9, TskData.TSK_DB_FILES_TYPE_ENUM.FS.getFileType());	//type
+			statement.setShort(10, (short) 1);										// has_path
+			TSK_FS_NAME_TYPE_ENUM dirType = isFile ? TSK_FS_NAME_TYPE_ENUM.REG : TSK_FS_NAME_TYPE_ENUM.DIR;
+			statement.setShort(11, dirType.getValue());								// dir_type
+			TSK_FS_META_TYPE_ENUM metaType = isFile ? TSK_FS_META_TYPE_ENUM.TSK_FS_META_TYPE_REG : TSK_FS_META_TYPE_ENUM.TSK_FS_META_TYPE_DIR;
+			statement.setShort(12, metaType.getValue());							// meta_type
+			statement.setShort(13, dirFlag.getValue());								// dir_flags
+			short metaFlags = (short) (dirFlag.getValue() | TSK_FS_META_FLAG_ENUM.USED.getValue());
```
bcarrier: We should have them specify metaflag explicitly.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -5743,6 +5744,115 @@ public FileSystem addFileSystem(long parentObjId, long imgOffset, TskData.TSK_FS
...
+			statement.setShort(14, metaFlags);										// meta_flags
+			statement.setLong(15, size);
```
bcarrier: We should check size and set it to 0 if it is <0 (like the C++ code does).

PR #1625, opened 2019-09-12
refactor logical imager to make paths more clear.  Other clean up.  N…

bcarrier:
…o logic changes. Fixed duplicate iterator name

### DIFF
--- a/tools/logicalimager/tsk_logical_imager.cpp
+++ b/tools/logicalimager/tsk_logical_imager.cpp
@@ -3,7 +3,7 @@
  ** The Sleuth Kit
  **
  ** Brian Carrier [carrier <at> sleuthkit [dot] org]
- ** Copyright (c) 2010-2011 Brian Carrier.  All Rights reserved
+ ** Copyright (c) 2019 Basis Technology.  All Rights reserved
  **
  ** This software is distributed under the Common Public License 1.0
  **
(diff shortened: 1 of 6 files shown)


tools/logicalimager/tsk_logical_imager.cpp
```diff
@@ -392,28 +483,48 @@ main(int argc, char **argv1)
...
+        ReportUtil::SetPromptBeforeExit(config->getPromptBeforeExit());
+        createVHD = config->getCreateVHD();
+    }
+    catch (std::exception &e) {
+        std::cerr << e.what() << std::endl;
+        ReportUtil::handleExit(1);
+    }
+
+    // If CWD is FAT, exit with error because it cannot create files greater 4 GB
+    if (createVHD && DriveUtil::cwdIsFAT(cwd)) {
```
jkho: Need to call cwdIsFAT because it outputs cwd. If non-VHD, a file can be greater than 4 GB.
if (DriveUtil::cwdIsFAT(cwd)) {

jkho: No action needed. I have merged Brian's commit into my branch for 5503.

rcordovano: @jkho, so is it correct that this PR can be closed?

jkho: correct.

PR #1628, opened 2019-09-16
5492: Refactor AppDBParserHelper

raman-bt:
- moved to sleuthkit/datamodel/blackboardutils
 - broke up into multiple classes
 - consolidated several addAtribute() calls into a single addAttributes() call
 - updated javadocs to indicate required and optional arguments

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/AbstractArtifactHelper.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/AbstractArtifactHelper.java
@@ -1,0 +1,61 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.sleuthkit.datamodel.blackboardutils;
+
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.SleuthkitCase;
+
+/**
+ * An abstract base class for classes that helps ingest modules create artifacts.
+ * 
+ */
+public abstract class AbstractArtifactHelper {
+	
+	private final SleuthkitCase caseDb;
+	private final AbstractFile srcAbstractFile;	// artifact source
+    private final String moduleName;			// module creating the artifacts
+	
+	/**
+	 * Creates an artifact helper.
+	 * 
+	 * @param caseDb Sleuthkit case db
+     * @param moduleName name module using the helper
+     * @param srcFile source file
+	 */
+	public AbstractArtifactHelper(SleuthkitCase caseDb, String moduleName, AbstractFile srcFile) {
+        this.moduleName = moduleName;
+        this.srcAbstractFile = srcFile;
+        this.caseDb = caseDb;
+	}
+	
+	/**
+	 * Returns the source abstract file.
+	 * 
+	 * @return source abstract file
+	 */
+	AbstractFile getAbstractFile() {
+		return this.srcAbstractFile;
+	}
+	
+	/**
+	 * Returns the sleuthkit case.
+	 * 
+	 * @return sleuthkit case
+	 */
+	SleuthkitCase getSleuthkitCase() {
+		return this.caseDb;
+	}
+	
+	/**
+	 * Returns module name.
+	 * 
+	 * @return module name 
+	 */
+	String getModuleName() {
+		return this.moduleName;
+	}
+	
+}
--- a/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/AbstractArtifactHelper.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/AbstractArtifactHelper.java
@@ -1,7 +1,20 @@
 /*
- * To change this license header, choose License Headers in Project Properties.
- * To change this template file, choose Tools | Templates
- * and open the template in the editor.
+ * Autopsy Forensic Browser
+ *
+ * Copyright 2019 Basis Technology Corp.
+ * Contact: carrier <at> sleuthkit <dot> org
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
  */
 package org.sleuthkit.datamodel.blackboardutils;
 
@@ -9,53 +22,54 @@ import org.sleuthkit.datamodel.AbstractFile;
 import org.sleuthkit.datamodel.SleuthkitCase;
 
 /**
- * An abstract base class for classes that helps ingest modules create artifacts.
- * 
+ * An abstract base class for classes that helps ingest modules create
+ * artifacts.
+ *
  */
 public abstract class AbstractArtifactHelper {
-	
+
 	private final SleuthkitCase caseDb;
 	private final AbstractFile srcAbstractFile;	// artifact source
-    private final String moduleName;			// module creating the artifacts
-	
+	private final String moduleName;			// module creating the artifacts
+
 	/**
 	 * Creates an artifact helper.
-	 * 
-	 * @param caseDb Sleuthkit case db
-     * @param moduleName name module using the helper
-     * @param srcFile source file
+	 *
+	 * @param caseDb     Sleuthkit case db
+	 * @param moduleName name module using the helper
+	 * @param srcFile    source file
 	 */
 	public AbstractArtifactHelper(SleuthkitCase caseDb, String moduleName, AbstractFile srcFile) {
-        this.moduleName = moduleName;
-        this.srcAbstractFile = srcFile;
-        this.caseDb = caseDb;
+		this.moduleName = moduleName;
+		this.srcAbstractFile = srcFile;
+		this.caseDb = caseDb;
 	}
-	
+
 	/**
 	 * Returns the source abstract file.
-	 * 
+	 *
 	 * @return source abstract file
 	 */
 	AbstractFile getAbstractFile() {
 		return this.srcAbstractFile;
 	}
-	
+
 	/**
 	 * Returns the sleuthkit case.
-	 * 
+	 *
 	 * @return sleuthkit case
 	 */
 	SleuthkitCase getSleuthkitCase() {
 		return this.caseDb;
 	}
-	
+
 	/**
 	 * Returns module name.
-	 * 
-	 * @return module name 
+	 *
+	 * @return module name
 	 */
 	String getModuleName() {
 		return this.moduleName;
 	}
-	
+
 }
--- a/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactHelper.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactHelper.java
@@ -18,7 +18,10 @@
  */
 package org.sleuthkit.datamodel.blackboardutils;
 
+import java.util.Collection;
+import org.apache.commons.lang3.StringUtils;
 import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.BlackboardAttribute;
 import org.sleuthkit.datamodel.SleuthkitCase;
 
 /**
@@ -26,7 +29,7 @@ import org.sleuthkit.datamodel.SleuthkitCase;
  * artifacts.
  *
  */
-public abstract class AbstractArtifactHelper {
+class ArtifactHelper {
 
 	private final SleuthkitCase caseDb;
 	private final AbstractFile srcAbstractFile;	// artifact source
@@ -39,7 +42,7 @@ public abstract class AbstractArtifactHelper {
 	 * @param moduleName name module using the helper
 	 * @param srcFile    source file
 	 */
-	public AbstractArtifactHelper(SleuthkitCase caseDb, String moduleName, AbstractFile srcFile) {
+	protected ArtifactHelper(SleuthkitCase caseDb, String moduleName, AbstractFile srcFile) {
 		this.moduleName = moduleName;
 		this.srcAbstractFile = srcFile;
 		this.caseDb = caseDb;
@@ -72,4 +75,30 @@ public abstract class AbstractArtifactHelper {
 		return this.moduleName;
 	}
 
+	/**
+	 * Creates and adds an attribute to the list, if the attribute value is not empty or null.
+	 *
+	 * @param attrValue attribute value 
+	 * @param attributeType attribute type
+	 * @param attributes list of attributes to add to.
+	 * 
+	 */
+	protected void addAttributeIfNotNull(String attrValue, BlackboardAttribute.ATTRIBUTE_TYPE attributeType, Collection<BlackboardAttribute> attributes) {
+		if (!StringUtils.isEmpty(attrValue)) {
+			attributes.add(new BlackboardAttribute(attributeType, getModuleName(), attrValue));
+		}
+	}
+
+	/**
+	 * Creates and adds an attribute to the list, if the attribute value is not 0.
+	 *
+	 * @param attrValue attribute value 
+	 * @param attributeType attribute type
+	 * @param attributes list of attributes to add to.
+	 */
+	protected void addAttributeIfNotZero(long attrValue, BlackboardAttribute.ATTRIBUTE_TYPE attributeType, Collection<BlackboardAttribute> attributes) {
+		if (attrValue > 0) {
+			attributes.add(new BlackboardAttribute(attributeType, getModuleName(), attrValue));
+		}
+	}
 }
--- a/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
@@ -23,7 +23,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import org.apache.commons.lang3.StringUtils;
 import org.sleuthkit.datamodel.AbstractFile;
 import org.sleuthkit.datamodel.Blackboard;
 import org.sleuthkit.datamodel.BlackboardArtifact;
--- a/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
@@ -1,0 +1,177 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
+package org.sleuthkit.datamodel.blackboardutils;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.apache.commons.lang3.StringUtils;
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.Blackboard;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+
+
+/**
+ * Helper class to help module modules create artifacts.
+ *
+ * This class helps ingest modules create miscellaneous artifacts.
+ * 
+ */
+public final class ArtifactsHelper extends AbstractArtifactHelper {
+	private static final Logger logger = Logger.getLogger(ArtifactsHelper.class.getName());
+
+	
+	/**
+	 * Creates an artifact helper for artifacts.
+	 * 
+	 * @param caseDb Sleuthkit case db
+     * @param moduleName name module using the helper
+     * @param srcFile source file
+	 * 
+	 */
+	public ArtifactsHelper(SleuthkitCase caseDb, String moduleName, AbstractFile srcFile) {
+		super(caseDb, moduleName, srcFile);
+	}
+	
+	
+	
+	
+	/**
+     * Adds a TSK_GPS_TRACKPOINT artifact.
+     * 
+      * @param latitude location latitude, required
+     * @param longitude location longitude, required
+     * @param timeStamp date/time trackpoint recorded, can be 0 if not available
+     * @param name trackpoint name, can be empty/null if not available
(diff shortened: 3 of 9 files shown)


bindings/java/src/org/sleuthkit/datamodel/blackboardutils/AbstractArtifactHelper.java
```diff
@@ -0,0 +1,61 @@
+/*
```
dannysmyda: Needs header

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/AbstractArtifactHelper.java
```diff
@@ -0,0 +1,61 @@
...
+package org.sleuthkit.datamodel.blackboardutils;
+
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.SleuthkitCase;
+
+/**
+ * An abstract base class for classes that helps ingest modules create artifacts.
+ * 
+ */
+public abstract class AbstractArtifactHelper {
```
dannysmyda: indentation is a little off, could you run src formatting on this file?

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,177 @@
+/*
```
dannysmyda: needs header

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,746 @@
...
+ * apps and create artifacts.
+ */
+public final class CommunicationArtifactsHelper extends AbstractArtifactHelper {
+    
+    private static final Logger logger = Logger.getLogger(CommunicationArtifactsHelper.class.getName());		
+			
+    /**
+     * Enum for message read status
+     */
+    public enum MessageReadStatusEnum {
```
dannysmyda: Enum is not needed in the name

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,746 @@
...
+     * @param readStatus message read or not, UNKNOWN if not available
+     * @param subject message subject, may be empty or null
+     * @param messageText message body, may be empty or null
+     * @param threadId, message thread id, may be empty or null
+	 * @param otherAttributesList other attributes, may be an empty list
+     * 
+     * 
+     * @return message artifact
+     */
+    public BlackboardArtifact addMessage( String messageType,
```
dannysmyda: it was my understanding that the npath problems were to be addressed in this refactor. @rcordovano since this would not affect the public API, are we deferring this (or not doing it)?

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/WebBrowserArtifactsHelper.java
```diff
@@ -0,0 +1,583 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
```
dannysmyda: needs header

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/WebBrowserArtifactsHelper.java
```diff
@@ -0,0 +1,583 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.Blackboard;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * Class to help ingest modules create Web Browser artifacts.
```
dannysmyda: Could this be combined into one sentence?

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,177 @@
...
+import org.sleuthkit.datamodel.AbstractFile;
+import org.sleuthkit.datamodel.Blackboard;
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+
+
+/**
+ * Helper class to help module modules create artifacts.
```
dannysmyda: could this be combined into one sentence?

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,746 @@
...
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskDataException;
+
+
+/**
+ * A helper class to support modules that parse SQLite databases from mobile 
+ * apps and create artifacts.
+ */
+public final class CommunicationArtifactsHelper extends AbstractArtifactHelper {
```
dannysmyda: Can src formatting in Netbeans be run on this class? Some indentation is off.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * This class helps ingest modules create miscellaneous artifacts.
+ * 
+ */
+public final class ArtifactsHelper extends AbstractArtifactHelper {
```
dannysmyda: Can you run the netbeans source formatting tool on this file? Some indentation is off.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/WebBrowserArtifactsHelper.java
```diff
@@ -0,0 +1,583 @@
...
+ *
+ * This class helps ingest modules create sundry/misc artifacts.
+ * 
+ */
+public final class WebBrowserArtifactsHelper extends AbstractArtifactHelper {
```
dannysmyda: can src formatting be run on this file also?

dannysmyda:
Sorry, I didn't approve, i meant to request feedback

raman-bt:
Addressed all review comments except the Codacy comments regarding complexity.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,180 @@
...
+	 * @param programName name of program that recorded trackpoint , can be
+	 *                    empty or null if not available
+	 *
+	 * @return artifact added
+	 */
+	public BlackboardArtifact addGPSLocation(double latitude, double longitude,
+			long timeStamp, String name, String programName) {
+
+		return addGPSLocation(latitude, longitude, timeStamp, name, programName,
+				Collections.<BlackboardAttribute>emptyList());
```
rcordovano: It is not necessary to supply the type parameter. You can just call Collections.emptyList() for greater simplicity. This comment applies to various call sites in this code.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,180 @@
...
+			gpsTrackpointArtifact.addAttributes(otherAttributesList);
+
+			// post artifact 
+			getSleuthkitCase().getBlackboard().postArtifact(gpsTrackpointArtifact, getModuleName());
+		} catch (TskCoreException ex) {
+			logger.log(Level.SEVERE, "Unable to add GPS trackpoint artifact", ex); //NON-NLS
+			return null;
+		} catch (Blackboard.BlackboardException ex) {
+			logger.log(Level.SEVERE, String.format("Unable to post artifact %s", ((gpsTrackpointArtifact != null) ? gpsTrackpointArtifact.getArtifactID() : "")), ex);  //NON-NLS
+		}
```
rcordovano: Low level utility classes should not take responsibility for exception handling policy. This is a project coding guideline. The exceptions should be thrown so that the client code can decide what to do with them. For example, the client module may wish to return an error code from its "process" method.

This comment applies to multiple locations in this PR.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,180 @@
...
+			Collection<BlackboardAttribute> otherAttributesList) {
+
+		BlackboardArtifact gpsTrackpointArtifact = null;
+		try {
+			Collection<BlackboardAttribute> attributes = new ArrayList<>();
+
+			// Create artifact
+			gpsTrackpointArtifact = getAbstractFile().newArtifact(BlackboardArtifact.ARTIFACT_TYPE.TSK_GPS_TRACKPOINT);
+
+			// Add basic attributes
```
rcordovano: I think it would be more informative to say "Construct attributes."

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,180 @@
...
+			attributes.add(new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_GEO_LONGITUDE, getModuleName(), longitude));
+
+			addAttributeIfNotZero(timeStamp, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_DATETIME, attributes);
+
+			addAttributeIfNotNull(name, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_NAME, attributes);
+			addAttributeIfNotNull(programName, BlackboardAttribute.ATTRIBUTE_TYPE.TSK_PROG_NAME, attributes);
+
+			// add the attributes 
+			gpsTrackpointArtifact.addAttributes(attributes);
+			gpsTrackpointArtifact.addAttributes(otherAttributesList);
```
rcordovano: Why make two database transactions and two round trips here? The two collections of attributes can be combined and added in a single addAttributes call. 

This comment applies to multiple locations in this PR.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+import org.sleuthkit.datamodel.DataSource;
+import org.sleuthkit.datamodel.Relationship;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+import org.sleuthkit.datamodel.TskDataException;
+
+/**
+ * A helper class to support modules that parse SQLite databases from mobile
+ * apps and create artifacts.
+ */
```
rcordovano: This comment is no longer true. This is a more general utility now.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+
+		CommunicationDirection(String dir) {
+			this.dirStr = dir;
+		}
+
+		public String getString() {
+			return dirStr;
+		}
+	}
+
```
rcordovano: Why do two of these enums get display names and the other does not?

If we need display names, it is better to have a getDisplayName method. The getString method name used here is too close to the name of the toString method provided by the Enum base class.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+	 * It creates a DeviceAccount instance to use as a self account.
+	 *
+	 * @param caseDb       Sleuthkit case db
+	 * @param moduleName   name module using the helper
+	 * @param srcFile      source file being processed by the module
+	 * @param accountsType account types created by this module
+	 *
+	 * @throws TskCoreException
+	 */
+	public CommunicationArtifactsHelper(SleuthkitCase caseDb, String moduleName, AbstractFile srcFile, Account.Type accountsType) throws TskCoreException {
```
rcordovano: The documentation incorrectly refers to AppDbParser and to "apps." It is now more general.

Since this is public API documentation, it would be useful to beef it up to help module writers. What is a self account? What does it mean for the helper to create account types? We have not said a thing about this class creating both accounts and artifacts, at least until we proceed to the next constructor.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+	}
+
+	/**
+	 * Constructs a AppDB parser helper for the given DB file.
+	 *
+	 * This constructor is for Apps that do have app specific account
+	 * information for the device owner to create a 'self' account.
+	 *
+	 * It creates a an account instance with specified type & id and uses it as
+	 * a self account.
```
rcordovano: See previous comment.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+		super(caseDb, moduleName, srcFile);
+
+		this.accountsType = accountsType;
+		this.selfAccountInstance = getSleuthkitCase().getCommunicationsManager().createAccountFileInstance(selfAccountType, selfAccountAddress.getUniqueID(), moduleName, getAbstractFile());
+	}
+
+	/**
+	 * Creates and adds a TSK_CONTACT artifact to the case, with specified
+	 * attributes. Also creates an account instance of specified type for the
+	 * contact with the specified ID.
```
rcordovano: The documentation here makes me look for the specified account type. We should either repeat the documentation of this from the constructor, or refer back to it.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+	 *
+	 * @param contactAccountUniqueID unique id for contact account
```
rcordovano: What is a contact account unique id, asks the module writer? Is this required or optional?

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+	 * @param contactName            Name of contact, required
+	 * @param phoneNumber            primary phone number for contact, may be
+	 *                               empty or null
+	 * @param homePhoneNumber        home phone number, may be empty or null
```
rcordovano: This is public documentation and should be polished. For example, we should either use lower case or upper case for parameter documentation. Please be consistent. When I'm reading documentation, lack of consistency suggests lack of attention to detail. I prefer initial upper case - perhaps this should become a project coding guideline.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+	 * @param messageType message type, may be empty or null
+	 * @param direction   message direction, UNKNOWN if not available
+	 * @param fromAddress sender address, may be null
+	 * @param toAddress   recipient address, may be null
+	 * @param dateTime    date/time of message, 0 if not available
+	 * @param readStatus  message read or not, UNKNOWN if not available
+	 * @param subject     message subject, may be empty or null
+	 * @param messageText message body, may be empty or null
+	 * @param threadId,   message thread id, may be empty or null
+	 *
```
rcordovano: The parameter documentation here strikes me as odd. Why can everything be UNKNOWN, empty, or null? What are the minimum attributes for a message artifact?

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -0,0 +1,719 @@
...
+	 * Also creates an account instance for the caller/callee, and creates a
+	 * relationship between the self account and the caller/callee account.
+	 *
+	 * @param direction     call direction
+	 * @param fromAddress   caller address, may be empty
+	 * @param toAddress     callee address, may be empty
+	 * @param startDateTime start date/time
+	 * @param endDateTime   end date/time
+	 *
```
rcordovano: How can the from and to addresses be empty (or null?)?

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/WebBrowserArtifactsHelper.java
```diff
@@ -0,0 +1,538 @@
...
+			return null;
+		} catch (Blackboard.BlackboardException ex) {
+			logger.log(Level.SEVERE, String.format("Unable to post artifact %s", ((webHistoryArtifact != null) ? webHistoryArtifact.getArtifactID() : "")), ex);  //NON-NLS
+		}
+
+		// return the artifact
+		return webHistoryArtifact;
+	}
+
+	// TBD: this is duplicated in Autopsy/NetworkUtils
```
rcordovano: We can create a util in TSK and have the Autopsy util delegate to it.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactHelper.java
```diff
@@ -0,0 +1,104 @@
...
+	private final String moduleName;			// module creating the artifacts
+
+	/**
+	 * Creates an artifact helper.
+	 *
+	 * @param caseDb     Sleuthkit case db
+	 * @param moduleName name module using the helper
+	 * @param srcFile    source file
+	 */
+	protected ArtifactHelper(SleuthkitCase caseDb, String moduleName, AbstractFile srcFile) {
```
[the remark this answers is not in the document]
raman-bt: Is it a requirement that this class should be subclassable from outside this package?  I assumed not. The reason I made the constructor protected was to prevent anyone outside of the package actually creating objects of this class( effectively making it an abstract base class).  I wanted this class to be abstract but Codacy (and you) complained about it not having any abstract method. I think a class should be allowed to be abstract without having any abstract method, depending on the use case. This class is good example of a class that should only be a base class but has no common method that all it's subclasses need to implement.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/WebBrowserArtifactsHelper.java
raman-bt: Should this util class go in the datamodel package? or the newly created datamodel.blackboardutils package? I'm leaning towards latter. 

Or we create a new package altogether.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactHelper.java
```diff
@@ -0,0 +1,104 @@
...
+	 *
+	 * @param attrValue attribute value 
+	 * @param attributeType attribute type
+	 * @param attributes list of attributes to add to.
+	 */
+	protected void addAttributeIfNotZero(long attrValue, BlackboardAttribute.ATTRIBUTE_TYPE attributeType, Collection<BlackboardAttribute> attributes) {
+		if (attrValue > 0) {
+			attributes.add(new BlackboardAttribute(attributeType, getModuleName(), attrValue));
+		}
+	}
```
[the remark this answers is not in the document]
raman-bt: long and string type attributes is all I needed right now. I figured we can add other types as and when needed - in keeping with "add only what's necessary"

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactsHelper.java
```diff
@@ -0,0 +1,180 @@
...
+import org.sleuthkit.datamodel.BlackboardArtifact;
+import org.sleuthkit.datamodel.BlackboardAttribute;
+import org.sleuthkit.datamodel.SleuthkitCase;
+import org.sleuthkit.datamodel.TskCoreException;
+
+/**
+ * This class helps ingest modules create miscellaneous artifacts.
+ *
+ */
+public final class ArtifactsHelper extends ArtifactHelper {
```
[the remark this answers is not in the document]
raman-bt: Just to clarify, ArtifactHelper (singular) is the base class.  ArtifactsHelper is the class that has methods for miscellaneous artifacts. 

But i agree with names being too close, and will rename ArtifactHelper  to ArtifactHelperBase.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactHelper.java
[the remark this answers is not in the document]
raman-bt: I realize, belatedly, that 'protected' does not prevent a non-subclass in same package from accessing it anyway. So default/package it is.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
raman-bt: The MessageReadStatus eventually goes into an integer type attribute, which is why  strings weren't defined for it.  I can define the string but it would only be used for logging purposes or such.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
rcordovano: We are either deferring or not doing anything more; the latter is most likely at this time.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/ArtifactHelper.java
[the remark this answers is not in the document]
rcordovano: Okay.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/WebBrowserArtifactsHelper.java
rcordovano: I'm not sure what vision @bcarrier has for the evolution of the TSK Java bindings. I could see creating a new package or putting it in the orignal datamodel package. I would not put it in datamodel.blackboardutils because I think it has greater scope than that.

For the sake of this PR, I think we should not hold things up waiting for Brian to communicate his wishes. For now, we can live with the code duplication in the non-public methods where this code currently resides.

PR #1655, opened 2019-10-01
Update README with links

f97gujo:
Update references to docs to links to the wiki page


rcordovano:
@APriestman, are these links to the wiki that is being phased out?

APriestman:
The wiki is no longer being updated but it's still the best source for this information. I think we should accept this - if we migrate the wiki data in the future we'll update all links then.

PR #1656, opened 2019-10-03
5602 an error with any single drive prevents logical imager from running on drives after that one

jkho:


rcordovano:
@jkho, please merge the develop branch into this branch to cut down on the diffs.

PR #1658, opened 2019-10-07
5640 Check that ID list is not empty

APriestman:


bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -831,12 +835,16 @@ private void updateEventSourceTaggedFlag(CaseDbConnection conn, Collection<Long>
...
+			if (! eventIDs.isEmpty()) {
+				String sql = "UPDATE tsk_event_descriptions SET hash_hit = 1" + " WHERE event_description_id IN (" + buildCSVString(eventIDs.values()) + ")"; //NON-NLS
+				try {
+					updateStatement.executeUpdate(sql); //NON-NLS
+					return eventIDs.keySet();
+				} catch (SQLException ex) {
+					throw new TskCoreException("Error setting hash_hit of events.", ex);//NON-NLS
+				}
+			} else {
+				return new HashSet<>();
```
rcordovano: Why not just return eventIDs.keySet() as we do elsewhere?

PR #1664, opened 2019-10-09
5589 Fix memory leaks in TSK

jkho:


tsk/base/mymalloc.c
```diff
@@ -44,10 +44,15 @@ tsk_malloc(size_t len)
 void *
 tsk_realloc(void *ptr, size_t len)
 {
-    if ((ptr = realloc(ptr, len)) == 0) {
```
bcarrier: I don't see what these changes fix?  Aren't they just the same thing, but more verbose?

jkho: If/when realloc failed, ptr is assigned to NULL. The original ptr is lost so a memory leak.
This was reported by VLD.

jkho: https://stackoverflow.com/questions/4548385/realloc-leaking-memory

bcarrier: Oh, we should make a comment then to that effect.  Otherwise, someone will return it back to what it was.

jkho: done

PR #1665, opened 2019-10-09
5589 Fix memory leaks in Rejistry++

jkho:


crayy8 (approved):
These changes look good.

PR #1677, opened 2019-10-30
First cut at removing FX dependency from TSK datamodel.

esaunders:


bindings/java/src/org/sleuthkit/datamodel/TimelineFilter.java
```diff
@@ -414,7 +411,7 @@ public String getDisplayName() {
 		 * @return The required substring.
 		 */
 		public synchronized String getSubstring() {
-			return descriptionSubstring.getValue();
+			return descriptionSubstring;
```
rcordovano: This method has a poor name. We should fix it while we have the opportunity to do so.

PR #1691, opened 2019-11-15
5698: associate message attachment files with the message.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -770,10 +770,30 @@ public void addAttachments(BlackboardArtifact message, MessageAttachments attach
...
+				AbstractFile attachedFile = message.getSleuthkitCase().getAbstractFileById(attachedFileObjId);
+				associateAttachmentWithMesssge(message, attachedFile);
+			}
+		}
 	}
 
+	/**
+	 * Creates a TSK_ASSOCIATED_OBJECT artifact between the attachment file and 
+	 * the message.
+	 */
```
rcordovano: Missing param docs.

raman-bt: I can certainly add it, but this is a private method, not an API.

PR #1700, opened 2019-11-27
5744 Change event table names to fix bug with incorrect column name in pre…

APriestman:
…vious update code.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
@@ -97,7 +97,7 @@ public class SleuthkitCase {
 	 * tsk/auto/tsk_db.h.
 	 */
 	private static final CaseDbSchemaVersionNumber CURRENT_DB_SCHEMA_VERSION
-			= new CaseDbSchemaVersionNumber(8, 3);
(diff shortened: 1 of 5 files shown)


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1913,6 +1912,136 @@ private CaseDbSchemaVersionNumber updateFromSchema8dot2toSchema8dot3(CaseDbSchem
...
+					+ " event_id " + primaryKeyType + " PRIMARY KEY, "
+					+ " event_type_id BIGINT NOT NULL REFERENCES tsk_event_types(event_type_id) ,"
+					+ " event_description_id BIGINT NOT NULL REFERENCES tsk_event_description(event_description_id),"
+					+ " time INTEGER NOT NULL, "
+					+ " UNIQUE (event_type_id, event_description_id, time))"
+			);						
+
+			// Copy to the new tsk_event_description table
+			existingEvents = getExistingEventDescriptionsStatement.executeQuery("SELECT * FROM tsk_event_descriptions");
+			while (existingEvents.next()) {
```
rcordovano: It seems that we should do an ORDER BY for at least cosmetic reasons.

PR #1705, opened 2019-12-02
5748: Update attachment file matching algorithm.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/blackboardutils/FileAttachment.java
```diff
@@ -155,6 +147,62 @@ private String normalizePath(String path) {
...
+
+		long bestMatchAllocatedFileObjId = -1;
+		int bestMatchAllocatedFileCount = 0;
+		long bestMatchUnallocatedFileObjId = -1;
+		int bestMatchUnallocatedFileCount = 0;
+
+		for (AbstractFile file : matchedFiles) {
+			if (file.isMetaFlagSet(TskData.TSK_FS_META_FLAG_ENUM.ALLOC)) {
+				bestMatchAllocatedFileCount++;
+				if (dataSource == file.getDataSource()) {
```
dannysmyda: dataSource is a parameter, there is no guarantee that they reference the same object.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/FileAttachment.java
```diff
@@ -155,6 +147,62 @@ private String normalizePath(String path) {
...
+					bestMatchAllocatedFileCount = 1;
+					bestMatchAllocatedFileObjId = file.getId();
```
dannysmyda: If there are multiple matches with only 1 being from the target data source, should you return that one? It seems like this is true from your comment above. However, if the target data source file is the first match of 5, lets say, then it will not be appropriately returned (count will be 5, not one).

rcordovano:
Is there anything we can do to drive down the cyclometric complexity of findAttachmentFile? Just glancing at it, it looks like there are a couple of nested ifs that can be unrolled with an && operator...

raman-bt:
> Is there anything we can do to drive down the cyclometric complexity of findAttachmentFile? Just glancing at it, it looks like there are a couple of nested ifs that can be unrolled with an && operator...

PR #1706, opened 2019-12-02
1706 Added list to keep track of MFT entries already seen when processing …

esaunders:
…NTFS attribute lists. This fixes a potential infinite loop which ends up blowing the stack.


tsk/fs/ntfs.c
```diff
@@ -2600,11 +2608,21 @@ ntfs_proc_attrlist(NTFS_INFO * ntfs,
         /* Process the attribute seq for this MFT entry and add them
          * to the TSK_FS_META structure
          */
+        if (processed_inum_list != NULL && tsk_stack_find(processed_inum_list, mftToDo[a])) {
+            if (tsk_verbose)
+                tsk_fprintf(stderr, "ntfs_proc_attrlist: skipping previously processed MFT entry\n");
```
bcarrier: I think we should set the error codes in this situation and return error.

esaunders: Do you think one of the existing error codes fits the bill or should I add a new one?

bcarrier: TSK_ERR_FS_CORRUPT should work.

esaunders: Ok, I've made the change.

tsk/fs/ntfs.c
```diff
@@ -2600,11 +2608,31 @@ ntfs_proc_attrlist(NTFS_INFO * ntfs,
...
+                fs_file->meta->addr,
+                tsk_getu48(fs->endian, mft->base_ref));
+            free(mft);
+            free(map);
+            free(buf);
+            return TSK_COR;
+        }
+
+        if (processed_inum_list == NULL) {
+            processed_inum_list = mftSeenList = tsk_stack_create();
```
APriestman: This could use a comment. I believe mftSeenList is being used to check if the stack was created in this method so we know whether to free it?

esaunders: Yes, that's right. I've added a comment to the code.

PR #1716, opened 2019-12-04
5712 Remove dependency on JavaFX from TSK

rcordovano:


rcordovano:
There is a corresponding Autopsy PR: https://github.com/sleuthkit/autopsy/pull/5468.

PR #1717, opened 2019-12-05
5486: String index out of bound.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/blackboardutils/FileAttachment.java
```diff
@@ -74,10 +74,15 @@ public FileAttachment(SleuthkitCase caseDb, Content dataSource, String pathName)
 
 		//normalize the slashes.
 		this.filePathName = normalizePath(pathName);
-
+		
+		
 		String fileName = filePathName.substring(filePathName.lastIndexOf('/') + 1);
-		String parentPathSubString = filePathName.substring(0, filePathName.lastIndexOf('/'));
+		if (fileName.isEmpty()) {
```
rcordovano: I'm guessing this came up. How do we get an attachment path with no file name component?

rcordovano: Ah, I see the answer in Jira.

bindings/java/src/org/sleuthkit/datamodel/blackboardutils/FileAttachment.java
```diff
@@ -74,10 +74,15 @@ public FileAttachment(SleuthkitCase caseDb, Content dataSource, String pathName)
...
+			throw new TskCoreException(String.format("No name specified for attachment: %s", pathName ));
+		}
```
rcordovano: Please add the data source object ID to the exception message as well as the file path.

PR #1746, opened 2019-12-18
5900 Fix compiler warnings.

APriestman:
Remove password param from usage statements.


esaunders:
The only comment I would have on this is the use of TSK_OFF_T for the attributeLength and rawSize parameters. The documentation on TSK_OFF_T is a little vague as to when it should be used but I'm wondering if there is a particular reason to use it here instead of uint64_t?

APriestman:
I'm using it because that's the type for "size" in TSK_FS_ATTR and that's what's being passed around. I agree that uint64_t seems like a better choice, but I didn't want to change existing code or cast between the two.

PR #1750, opened 2019-12-19
Lazy creation of Self Account Instance - do not create till actually …

raman-bt:
…needed.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
@@ -120,7 +120,10 @@ public final class CommunicationArtifactsHelper extends ArtifactHelperBase {
 	}
 
 	// 'self' account for the application being processed by the module. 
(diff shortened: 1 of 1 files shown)


bindings/java/src/org/sleuthkit/datamodel/blackboardutils/CommunicationArtifactsHelper.java
```diff
@@ -864,4 +869,16 @@ private void addMessageReadStatusIfKnown(MessageReadStatus readStatus, Collectio
 		}
 	}
 
+	/**
+	 * Returns self account instance.  Lazily creates it if one doesn't exist yet.
+	 * 
+	 * @return Self account instance.
+	 * @throws TskCoreException 
+	 */
+	private AccountFileInstance getSelfAccountInstance() throws TskCoreException {
```
rcordovano: We should probably synchronize this method. As we drive down the call stack we eventually get to an INSERT/IGNORE, but that does not ensure visibility of the value of selfAccountInstance across all threads and making that field volatile would still leave us with a check-then-act race condition. Objects of this class were immutable and therefore thread-safe; I think we should keep it thread-safe as utility class for module developers.

PR #1758, opened 2019-12-31
5915: Add new account types on a upgrade.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -1907,6 +1907,19 @@ private CaseDbSchemaVersionNumber updateFromSchema8dot2toSchema8dot3(CaseDbSchem
...
+			statement.execute("insert into account_types(type_name, display_name) values('LINE', 'LINE')");
+			statement.execute("insert into account_types(type_name, display_name) values('SKYPE', 'Skype')");
+			statement.execute("insert into account_types(type_name, display_name) values('TANGO', 'Tango')");
+			statement.execute("insert into account_types(type_name, display_name) values('TEXTNOW', 'TextNow')");
+			statement.execute("insert into account_types(type_name, display_name) values('THREEMA', 'ThreeMa')");
+			statement.execute("insert into account_types(type_name, display_name) values('VIBER', 'Viber')");
+			statement.execute("insert into account_types(type_name, display_name) values('XENDER', 'Xender')");
+			statement.execute("insert into account_types(type_name, display_name) values('ZAPYA', 'Zapya')");
+			statement.execute("insert into account_types(type_name, display_name) values('SHAREIT', 'ShareIt')");
+
```
wschaeferB: We probably want to put these in the 8dot3toSchema8dot4 upgrade instead of the 8.2 to 8.3 upgrade. 
We would also want to surround it with a check of the db created version when we move it to check that the data base was created with a version prior to 8.3.  In this way users that had already upgrades a case from 8.2 to 8.3 prior to the newest release would not be left with a broken case.

PR #1778, opened 2020-01-15
Add pkg-config support

uckelman-sf:
This PR

* produces a `.pc` file for libtsk so that it can be found via `pkg-config`,
* lets `configure` find libtsk's dependencies via `pkg-config`,
* eliminates code duplication among the checks for libtsk's dependencies, 
* fixes a few infelicities in how cppunit is used, and
* fixes a typo in configure.ac which keeps the `-pthread` flag out of `LIBS`.


jonstewart:
*bump*

bcarrier:
Can you separate this into two PRs? One to make the .pc and another to change how TSK finds dependencies?

The creation of the .pc file seems very low risk and easy to merge in. 

The changing of TSK dependency finding is much more risky (to me at least) since I think every autoconf PR in the past has broken something and this is a sizable change that I don't fully understand yet.

jonstewart:
That makes sense. There should be a separate PR for the CppUnit fixes, too.

uckelman-sf:
General build fixes are now in PR #1875.

uckelman-sf:
cppunit fixes are now in PR #1876.

uckelman-sf:
pkg-config support is now in PR #1877.

uckelman-sf:
> Can you separate this into two PRs? One to make the .pc and another to change how TSK finds dependencies?

Not in any way which makes sense, because producing a correct .pc file depends on dependency discovery. See below for more details.
 
> The creation of the .pc file seems very low risk and easy to merge in.

`Requires.private` in the .pc file is supposed to be filled with the names of dependencies which can be located via pkg-config, whilst `Libraries.private` is supposed to be filled with the flags required by libraries which cannot be located via pkg-config. If we don't do the work to find dependencies, we can't produce a correct .pc file, as we can't tell which dependencies should be listed in `Requires.private` and which should be listed in `Libraries.private`, and that defeats the purpose of using pkg-config at all.

In order to make the diff of `configure.ac` easier to follow, I've moved the macros I defined out to an auxiliary file. This way, the other purpose of this PR can be better observed as well, namely to eliminate the code duplicated each time an optional dependency was checked. The checks for afflib, zlib, libew, libvhdi, and libvmdk are all identical and one line each in `configure.ac` now.

I'm closing this PR in favor of #1877.

PR #1779, opened 2020-01-16
Change Ivy download URL to HTTPS

uckelman-sf:
Maven stopped supporting HTTP on 15 Jan 2020. Changed Ivy download URL to HTTPS.

All of your Travis builds will fail without this fix.

### DIFF
--- a/bindings/java/build.xml
+++ b/bindings/java/build.xml
@@ -90,7 +90,7 @@
 		</path>
 	</target>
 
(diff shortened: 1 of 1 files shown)


esaunders:
Because upgrading Ivy requires more work than we want to take on at this time we've gone with the following solution for the Maven issue: https://github.com/sleuthkit/sleuthkit/pull/1788

PR #1780, opened 2020-01-16
5962 Force reads to be sector-aligned if reading from a local disk

APriestman:


uckelman-sf:
The problem you're having on Travis is because Maven turned off HTTP in favor of HTTPS yesterday. See PR #1779 where I'm trying to resolve this problem.

APriestman:
We're also working on a fix here https://github.com/sleuthkit/autopsy/pull/5597/files

uckelman-sf:
Have you considered upgrading Ivy? You don't need to make URL changes other than the Maven one in `bindings/java/build.xml` that way. There's another issue which arises with Ivy 2.5.0 that I don't yet understand (an error from javac), but that may be worth resolving regardless---Ivy 2.3.0-rc1is 8 years old at this point.

rcordovano:
@APriestman, it looks like Codacy is unhappy about some unused variable assignments. It looks like at least one is a local. Is there a fix here?

APriestman:
The unused one is used. We put a 64-bit value into quad part and then pass the low part and the high part into SetFilePointer(). The other is only showing up because I changed a variable name.

PR #1781, opened 2020-01-16
1387 gps artifact timeline support

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TimelineEventArtifactTypeImpl.java
```diff
@@ -126,6 +131,10 @@ TimelineEventDescriptionWithTime makeEventDescription(BlackboardArtifact artifac
 		}
 		BlackboardAttribute timeAttribute = artifact.getAttribute(getDateTimeAttributeType());
 		if (timeAttribute == null) {
+			if (artifact.getArtifactID() != TSK_GPS_TRACK.getTypeID()) {
+				return makeRouteDescription(artifact);
+			}
+
```
rcordovano: My reading of this code is that the design calls for TimelineEventArtifactTypeImpl to be customized by supplying a "timeline attribute" and a set of TSKCoreCheckedFunction implementations to convert that attribute into event descriptions. Making an end run around the design does not seem like a good idea to me. We should talk about your motivation for doing so.

PR #1782, opened 2020-01-16
5967: Path for derived file attachments should show the path in the d…

raman-bt:
…ata source, not local path in the case.


raman-bt:
@rcordovano: can this be merged?

PR #1784, opened 2020-01-17
5868 Keep duplicate image handle in cache under a random path

APriestman:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -800,8 +800,10 @@ private static long openImage(String[] imageFiles, int sSize, boolean useCache,
 					nonNullCaseDbPointer = HandleCache.getDefaultCaseDbPointer();
 				}
 				
-				// If we're getting a fresh copy, move the existing cache reference so
-				// it won't be used by any subsequent calls to openImage but will still be valid.			
+				// If we're getting a fresh copy and an image with this path is already
+				// in the cache, move the existing cache reference so it won't be used by 
+				// any subsequent calls to openImage but will still be valid if any objects
+				// have it cached.
```
rcordovano: Let's explicitly mention adding the same data source twice, please. We could even reference the story as JIRA-XXXX.

PR #1795, opened 2020-01-26
upgrade ivy to fix fetch url errors

xloem:
This fixes initial setup errors in the Java bindings that occur now if the ivy dependencies are not already present, because server policies have changed since 2012 to reject http, and ivy 2.3 doesn't use https.

### DIFF
--- a/bindings/java/build.xml
+++ b/bindings/java/build.xml
@@ -90,7 +90,7 @@
 		</path>
 	</target>
 
-	<property name="ivy.install.version" value="2.3.0-rc2"/>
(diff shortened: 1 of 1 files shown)


esaunders:
This should have been fixed by https://github.com/sleuthkit/sleuthkit/pull/1779.
Are you still having issues after that change?

FYI: We had a similar submission here https://github.com/sleuthkit/sleuthkit/pull/1788

xloem:
Thanks and sorry, I must have been working off an old checkout.

PR #1799, opened 2020-01-27
5955-Dont-add-events-if-they-are-too-far-in-the-future

markmckinnon:
Do not add timeline events if dates for files and artifacts are 12 years or greater from the current date/time.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
@@ -25,6 +25,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -36,6 +37,8 @@ import java.util.Objects;
 import static java.util.Objects.isNull;
 import java.util.Optional;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import org.joda.time.DateTimeZone;
 import org.joda.time.Interval;
@@ -51,6 +54,8 @@ import static org.sleuthkit.datamodel.StringUtils.buildCSVString;
  */
 public final class TimelineManager {
 
+	private static final Logger logger = Logger.getLogger(TimelineManager.class.getName());
+	
 	/**
 	 * Timeline event types added to the case database when it is created.
 	 */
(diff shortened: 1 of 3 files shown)


bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -605,8 +619,11 @@ private long addEventDescription(long dataSourceObjId, long fileObjId, Long arti
 			return Optional.empty();
 		}
 		long time = eventPayload.getTime();
-		// if the time is legitimate ( greater than zero ) insert it into the db
-		if (time <= 0) {
+		// if the time is legitimate ( greater than zero and less then 12 years from present time) insert it into the db
+		if (time <= 0 || time >= MAX_TIMESTAMP_TO_ADD) {
+			if (time >= MAX_TIMESTAMP_TO_ADD) {
+				logger.log(Level.WARNING, String.format("Date/Time discarded from Timeline for %s for artifact %s", artifact.getDisplayName(), eventPayload.getDescription(TimelineLevelOfDetail.HIGH)));
```
raman-bt: Include the source file object id in the log.

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -25,6 +25,7 @@
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.time.Instant;
```
raman-bt: Update the year in Copyright statement in file header to 2020 (and in all affected files)

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -501,7 +511,7 @@ private long addEventDescription(long dataSourceObjId, long fileObjId, Long arti
 
 			for (Map.Entry<TimelineEventType, Long> timeEntry : timeMap.entrySet()) {
 				Long time = timeEntry.getValue();
-				if (time > 0) {// if the time is legitimate ( greater than zero ) insert it
+				if (time > 0 || time <= MAX_TIMESTAMP_TO_ADD) {// if the time is legitimate ( greater than zero or less then 12 years from current date) insert it
```
rcordovano: @markmckinnon and @raman-bt, shouldn't this be an AND test? With short-circuit evaluation of the conditional, all times greater than zero will pass the test.

raman-bt: You're right Richard. Ugh, I missed that.

rcordovano:
@raman-bt, please re-review, thanks.

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
markmckinnon: Fixed

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
markmckinnon: FIxed

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
markmckinnon: Fixed

PR #1800, opened 2020-01-27
1401 Add SleuthkitCase function to get count of artifacts by type and data source

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -2934,6 +2934,42 @@ public long getBlackboardArtifactsTypeCount(int artifactTypeID) throws TskCoreEx
 			releaseSingleUserCaseReadLock();
 		}
 	}
+	
+		/**
```
APriestman: Please fix this indent.

PR #1801, opened 2020-01-28
Added HFS types newly missing from tsk_fs_type_toname()

uckelman-sf:
* TSK_FS_TYPE_HFS_DETECT enum changed its value when APFS support was added. It's  no longer equal to the TSK_FS_TYPE_HFS enum, so no name was being found for filesystems identified as that; similarly, there was no entry for the newly added TSK_FS_TYPE_HFS_LEGACY. Both are added. (This worked correctly in 4.7.0.)

* Don't use TSK_FS_TYPE_X_DETECT enum when it has the same value as the TSL_FS_TYPE_X enum. That's just asking for trouble.

### DIFF
--- a/tsk/fs/fs_types.c
+++ b/tsk/fs/fs_types.c
@@ -36,16 +36,18 @@ typedef struct {
  * legacy strings - in order of expected usage
  */
 static FS_TYPES fs_type_table[] = {
-    {"ntfs", TSK_FS_TYPE_NTFS_DETECT, "NTFS"},
+    {"ntfs", TSK_FS_TYPE_NTFS, "NTFS"},
     {"fat", TSK_FS_TYPE_FAT_DETECT, "FAT (Auto Detection)"},
     {"ext", TSK_FS_TYPE_EXT_DETECT, "ExtX (Auto Detection)"},
-    {"iso9660", TSK_FS_TYPE_ISO9660_DETECT, "ISO9660 CD"},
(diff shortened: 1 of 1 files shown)


uckelman-sf:
The last push fixes a typo in the commit message.

bcarrier:
I just made commits 53bcfc1 and 8de7596e to clarify more about the types via comments. 

There are a couple of things going on here:
1) HFS_LEGACY was added as a type, but is not actually used anywhere.   I think we can just take it away and resolve a bunch of issues....

2) If we do need to keep HFS_LEGACY, tsk_hfs_open should set a specific type in TSK_FS_INFO  once it detects the type.  The _DETECT versions should not be stored in TSK_FS_INFO.  I added a comment to reflect this. 

3) It is correct to add HFS-based entries to fs_type_table so that we have each unique value listed in there, but we also need unique names so that we can map from string to ID.  In the PR, they all have a name of "hfs", so if someone types '-f hfs', then it could map to any of them, but will probably go to HFS+ since it is listed first.   I added comments to reflect this too. 

But, the first question is why HFS_LEGACY was added as a type and not used.  Any memory of this @jtsylve? Is it dead code that can be removed?

bcarrier:
I heard from Joe that he had a use for HFS_LEGACY, so I kept it in.
- hfs_open was already setting the type to FS_HFS, so #2 above was not relevant. 
- I added some comments and unique names to fs_types.

Thanks!

PR #1805, opened 2020-01-30
6008 Update SleuthkitCase.java

markmckinnon:
Add file to timeline when it is added into the filesystem


raman-bt:
@markmckinnon: could you merge develop into this branch now that @rcordovano has merged the fix for 5955 into develop?

I need both to test this.

PR #1814, opened 2020-02-05
1406 - Deprecated TSK_GPS_TRACKPOINT

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -962,6 +962,7 @@ public int hashCode() {
 		/**
 		 * A GPS track point (geolocation data).
 		 */
+		@Deprecated
```
rcordovano: Please also add a deprecated tag to the doc with a "use this instead" clause. E.g., you will find this in the docs for an old Case method: 

@deprecated Use closeCurrentCase instead.

kellykelly3:
@rcordovano Don't merge this yet, I might pull the functions from GeoArtifactHelper back in depending on the results of the conversation in EUR-1407

PR #1817, opened 2020-02-06
1408 & 1409: Added new TSK_GEO_WAYPOINTS attribute to TSK_GPS_ROUTE artifact

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/Bundle.properties
```diff
@@ -193,7 +193,8 @@ BlackboardAttribute.tskgroups.text=Groups
 BlackboardAttribute.tskattachments.text=Message Attachments
 BlackboardAttribute.tskdronehpdistance.text=Distance From Home Point
 BlackboardAttribute.tskdronedistancetraveled.text=Total Distance Traveled
-BlackboardAttribute.tskgeopath.text=Waypoint List
+BlackboardAttribute.tskgeopath.text=List of Track Points
+BlackboardAttribute.tskgeowaypoints.text=List of Waypoints Points
```
raman-bt: The last "... Points" seems extraneous here.

PR #1819, opened 2020-02-10
Format and copyright update for Examiner.java

rcordovano:


bindings/java/src/org/sleuthkit/datamodel/Examiner.java
```diff
@@ -1,7 +1,7 @@
 /*
  * Sleuth Kit Data Model
  *
- * Copyright 2018 Basis Technology Corp.
+ * Copyright 2018-2020 Basis Technology Corp.
```
gdicristofaro: Looks good.

PR #1834, opened 2020-02-18
6010 Move case creation to Java

APriestman:


bindings/java/src/org/sleuthkit/datamodel/SQLHelper.java
```diff
@@ -0,0 +1,91 @@
...
+	
+	// Get the description column name for the tsk_vs_parts table.
+	// This varies between SQLite and PostgreSQL.
+	String getVSDescColName();
+
+
+	/**
+	 * PostgreSQL-specific implementation
+	 */
+	class PostgreSQLHelper implements SQLHelper {
```
rcordovano: I think it would be better to put the concrete implementations of the interface in their own files.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
```diff
@@ -0,0 +1,595 @@
...
+	}
+	
+	/**
+	 * Add and initialize the database tables 
+	 * 
+	 * @param conn the database connection
+	 * 
+	 * @throws TskCoreException 
+	 */
+	private void addTables(Connection conn) throws TskCoreException {
```
rcordovano: I see Codacy's point regarding the length of this method. How about breaking it up into some smaller methods?

bindings/java/src/org/sleuthkit/datamodel/SQLHelper.java
rcordovano: Ann and I discussed this and decided to leave things as is for now.

PR #1854, opened 2020-03-19
6130 Calculate sector size for raw images

APriestman:


tsk/img/raw.c
```diff
@@ -630,6 +630,112 @@ get_size(const TSK_TCHAR * a_file, uint8_t a_is_winobj)
...
+    // Success
+    return 1;
+}
+
+/**
+ * Attempts to calculate the actual sector size needed for reading the image.
+ * If successful, the calculated sector size will be stored in raw_info. If it
+ * fails the sector_size field will not be updated.
+*/
+void
```
bcarrier: can be static.

tsk/img/raw.c
```diff
@@ -630,6 +630,112 @@ get_size(const TSK_TCHAR * a_file, uint8_t a_is_winobj)
...
 }
 
+#ifdef TSK_WIN32
+/**
+* \internal
+* Test seeking to the given offset and then reading a sector.
+* @return 1 if the seek/read is successful, 0 if not
+*/
+static int
+test_sector_read(HANDLE file_handle, TSK_OFF_T offset, DWORD len, char * buf) {
```
bcarrier: Should document argument units. Namely if offset is in bytes or sectors, etc.

tsk/img/raw.c
```diff
@@ -683,6 +786,23 @@ raw_open(int a_num_img, const TSK_TCHAR * const a_images[],
...
+    /* Set the sector size using the first image file or the given parameter*/
+    img_info->sector_size = 512;
+#ifdef TSK_WIN32
+    /* On Windows, figure out the actual sector size if one was not given.
+     * This is to prevent problems reading from devices later. */
+    if (a_ssize) {
+        img_info->sector_size = a_ssize;
+    }
+    else {
+        find_sector_size(raw_info, a_images[0], first_seg_size);
```
bcarrier: We should do this only for devices.

tsk/img/raw.c
```diff
@@ -630,6 +630,112 @@ get_size(const TSK_TCHAR * a_file, uint8_t a_is_winobj)
...
+        else {
+            needs_sector_alignment = 1;
+        }
+    }
+
+    // If reading a sector starting at offset 1 failed, the assumption is that we have a device
+    // that requires reads to be sector-aligned. 
+    if (needs_sector_alignment) {
+        // Start at 512 and double up to max_sector_size (4096)
+        unsigned int sector_size = 512;
```
bcarrier: I'd simplify this logic and either ignore whatever is in img_info or just always start with that (which should be 512).

tsk/img/raw.c
```diff
@@ -630,6 +630,112 @@ get_size(const TSK_TCHAR * a_file, uint8_t a_is_winobj)
...
+    return 1;
+}
+
+/**
+ * Attempts to calculate the actual sector size needed for reading the image.
+ * If successful, the calculated sector size will be stored in raw_info. If it
+ * fails the sector_size field will not be updated.
+*/
+void
+find_sector_size(IMG_RAW_INFO * raw_info, const TSK_TCHAR * image_name, TSK_OFF_T first_seg_size) {
```
bcarrier: I'd rename first_seg_size to be something like image_size.

bcarrier: I'd also rename the method to be something lke set_device_sector_size

PR #1861, opened 2020-03-25
6172: Validation of correlation attribute value not happening correctly

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -1315,52 +1315,19 @@ int getAccountTypeId(Account.Type accountType) {
 	 *
 	 * @return The normalized account id.
 	 */
-	private String normalizeAccountID(Account.Type accountType, String accountUniqueID) {
+	private String normalizeAccountID(Account.Type accountType, String accountUniqueID) throws TskCoreException {
```
dannysmyda: I don't like the idea of throwing a TskCoreException. Reason being, as a developer, I can't tell if I need to immediately shut down my module due to a fatal error, or if its really not a big deal because its dirty data. In fact, I think that Skype, TextNow, and others will immediately shut down their current table parsing upon encountering TskCoreException because they've assumed this is reserved for unrecoverable errors. By current table parsing, I mean they might go on to try the Calls table if they hit a TskCoreException while reading the Contacts table.

dannysmyda: Do we really consider this to be an exceptional case anyways? What's the problem with logging an acknowledgement that the account will be dropped?

bindings/java/src/org/sleuthkit/datamodel/CommunicationsUtils.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+package org.sleuthkit.datamodel;
+
+import org.apache.commons.validator.routines.EmailValidator;
+
+/**
+ * Provides general utility methods related to communications artifacts.
+ * 
+ */
+public final class CommunicationsUtils {
```
dannysmyda: We should probably write some tests for this file, mainly for normalizePhoneNum. Do you agree?

dannysmyda:
Could you add a private constructor in CommunicationUtils to make Codacy happy?

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
raman-bt: Are you referring to the exception being thrown from the normalizeAccountID() method?  That method is private.

Or are you referring to the exception being thrown from the methods in the CommunicationsUtils() class?  If so, what's the alternative - define a new exception? I agree that it should be something that can be ignored, without shutting down processing.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsUtils.java
raman-bt: As it happens, it does get tested, albeit indirectly, by CorrelationAttributeNormalizerTest.

raman-bt: Tagging @bcarrier here since it's  change to public API and he needs to approve it.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
dannysmyda: The exception gets propagated whenever the user calls a public function in CommunicationManager that depends on normalizeAccountID. Lets take createAccountFileInstance() as an example. Its composed of 3 function calls, all of them produce TskCoreExceptions. The underlying issue can range from being unable to get a connection to the Case DB (due to the case being closed), to data not being valid. 

Forget the outside module developer, how are we (you and I and Richard) suppose to differentiate between TskCoreException instances to ensure we do the right thing for the user? Read the exception message text and if else on its contents?

I was hopeful that we could just avoid adding to this problem, but logging an acknowledgement and returning probably isn't okay either. I don't think anyone is expecting a null AccountFileInstance reference from this function.

dannysmyda: It doesn't matter if its private, it will get propagated out to the caller in some function, for example createAcountFileInstance(). 

Here's what I'm really getting at, TskCoreException is not the correct exception here. TskDataException is better, but the APIs for CommunicationManager don't throw TskException, which is the common base class between TskCore and TskData and you can't change the signature without breaking code.

Also, logging an acknowledgement won't work either. Take createAccountFileInstance() for example, its expected to return an AccountFileInstance and I don't think anyone is every expecting a null from this function.

Its fine as is I guess.

raman-bt: I don't disagree with you on the need of having multiple exception types. 
But I think we should hash it out and do it as a separate story, it affects the public API signature.

I do think our handling of the TskCoreException in the module is at a very coarse level. We need a finer grain exception and a handler which says - "something is wrong with this particular contact/calllog/message - skip it and keep going", instead of stop everything.

PR #1870, opened 2020-04-03
6219: TskCoreException thrown from normalizePhoneNum.

raman-bt:
Do not attempt to create an account when we do not have a valid phone number or email address.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/CommunicationsUtils.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/CommunicationsUtils.java
@@ -45,7 +45,7 @@ public final class CommunicationsUtils {
 	 * 
 	 */
 	public static String normalizePhoneNum(String phoneNum) throws TskCoreException {
-		if (phoneNum.matches("\\+?[0-9()\\-\\s]+")) {
+		if (isValidPhoneNumber(phoneNum)) {
            return phoneNum.replaceAll("[^0-9\\+]", "");
         } else {
             throw new TskCoreException(String.format("Input string is not a valid phone number: %s", phoneNum));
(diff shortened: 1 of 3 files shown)


bindings/java/src/org/sleuthkit/datamodel/CommunicationsUtils.java
```diff
@@ -53,22 +53,69 @@ public static String normalizePhoneNum(String phoneNum) throws TskCoreException
...
+	/**
+	 * Checks if the given string is a valid phone number.
+	 *
+	 * @param phoneNum Phone number string to check.
+	 *
+	 * @return True if the given string is a valid phone number, false otherwise.
+	 */
+	public static boolean isValidPhoneNumber(String phoneNum) {
+		if (!StringUtils.isEmpty(phoneNum)) {
+			return phoneNum.matches("\\+?[0-9()\\-\\s]+");
```
rcordovano, line 101: Where did this validity check come from? Do we need to replace the code where this came from with a call to this method?

dannysmyda: Not the author, but that is from the normalizePhoneNumber method. I agree that this should call that method and catch the exception to ensure they don't diverge in their validation logic.

dannysmyda:
@raman-bt Don't forget about addWebFormAddress() on the WebBrowserArtifactsHelper.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsUtils.java
dannysmyda: @raman-bt Could you change this via discussion above?

PR #1877, opened 2020-04-09
Add pkg-config support

uckelman-sf:
* Produce a .pc file for libtsk.
* Move the libdl check to occur before the non-pkgconfig sqlite3 check,
and do it only once.
* Add dependency CFLAGS to CXXFLAGS, since there are now cpp files which
need them.
* Defined TSK_OPT_DEP_CHECK to eliminate duplicate code when checking for
optional dependencies.
* Added file for AX_PKG_CHECK_MODULES. Enabled use of pkg-config for
depenencies which are known to support it.
* Set library variable to "no" when AC_CHECK_HEADERS or AC_CHECK_LIB fails.


bcarrier:
Is there anyone else who wants this feature who can test it and help this get merged in?

bcarrier:
Disclosure: I did no testing on this except to confirm that it doesn't break any of our builds.

uckelman-sf:
> Disclosure: I did no testing on this except to confirm that it doesn't break any of our builds.

I'm happy to field questions anyone has about pkg-config support.

PR #1891, opened 2020-04-21
1440 & 1441 add tag set support

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TagSet.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+package org.sleuthkit.datamodel;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * A TagSet is a name group of TagNames.
```
rcordovano, line 27: name => named

bindings/java/src/org/sleuthkit/datamodel/TagSet.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+public class TagSet {
+
+	private final String setName;
+	private final int id;
```
rcordovano, line 32: Does this field map to a long in the case database? In any case, we should be consistent - we are using longs for generated IDs pretty much everywhere else.

bindings/java/src/org/sleuthkit/datamodel/TagSet.java
```diff
@@ -0,0 +1,128 @@
...
+	 * Returns the name of the tag set.
+	 * 
+	 * @return Tag set name.
+	 */
+	public String getName() {
+		return setName;
+	}
+
+	/**
+	 * Returns a list of the TagName object that belong to the tag set.
```
rcordovano, line 60: object -> objects

bindings/java/src/org/sleuthkit/datamodel/TagSet.java
```diff
@@ -0,0 +1,128 @@
...
+	 * 
+	 * @return A unmodifiable list of TagName objects.
```
rcordovano, line 62: A -> An

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -477,6 +479,10 @@ public TimelineManager getTimelineManager() throws TskCoreException {
 	public synchronized CaseDbAccessManager getCaseDbAccessManager() throws TskCoreException {
 		return dbAccessManager;
 	}
+	
+	public synchronized TaggingManager getTagsManager() {
```
rcordovano, line 483: This should probably be getTaggingManager.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -9712,33 +9768,9 @@ public TagName addOrUpdateTagName(String displayName, String description, TagNam
 	 *
 	 * @throws TskCoreException
 	 */
+	@Deprecated
```
rcordovano, line 9771: Please add an @deprecated entry to the method header comment as well, wherever we are deprecating existing methods.

tsk/auto/db_postgresql.cpp
```diff
@@ -562,8 +562,10 @@ int TskDbPostgreSQL::initialize() {
         attempt_exec("CREATE TABLE tsk_files_derived (obj_id BIGSERIAL PRIMARY KEY, derived_id BIGINT NOT NULL, rederive TEXT, FOREIGN KEY(obj_id) REFERENCES tsk_objects(obj_id) ON DELETE CASCADE)", "Error creating tsk_files_derived table: %s\n")
         ||
         attempt_exec("CREATE TABLE tsk_files_derived_method (derived_id BIGSERIAL PRIMARY KEY, tool_name TEXT NOT NULL, tool_version TEXT NOT NULL, other TEXT)", "Error creating tsk_files_derived_method table: %s\n")
-        ||
-        attempt_exec("CREATE TABLE tag_names (tag_name_id BIGSERIAL PRIMARY KEY, display_name TEXT UNIQUE, description TEXT NOT NULL, color TEXT NOT NULL, knownStatus INTEGER NOT NULL)", "Error creating tag_names table: %s\n")
+		||
+		attempt_exec("CREATE TABLE tag_sets (tag_set_id BIGSERIAL PRIMARY KEY, name TEXT UNIQUE);", "Error creating tag_sets table: %s\n")
+		||
+        attempt_exec("CREATE TABLE tag_names (tag_name_id BIGSERIAL PRIMARY KEY, display_name TEXT UNIQUE, description TEXT NOT NULL, color TEXT NOT NULL, knownStatus INTEGER NOT NULL, tag_set_id INTEGER, FOREIGN KEY(tag_set_id) REFERENCES tag_sets(tag_set_id) ON DELETE CASCADE)", "Error creating tag_names table: %s\n")
```
rcordovano, line 568: I don't believe we can do the cascading delete without adding the same constraint to the FKs into this table from content_tags and blackboard_artifact_tags.

kellykelly3: In standup we decided to replace the cascade with a SET NULL.

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
```diff
@@ -0,0 +1,458 @@
...
+	 * Construct a TaggingManager for the given SleuthkitCase.
+	 *
+	 * @param skCase The SleuthkitCase.
+	 */
+	TaggingManager(SleuthkitCase skCase) {
+		this.skCase = skCase;
+	}
+
+	/**
+	 * Returns a list of all the TagSet that exist in the case.
```
rcordovano, line 46: TagSet -> TagSets

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
```diff
@@ -0,0 +1,458 @@
...
+			throw new TskCoreException("Error occurred getting TagSet list.", ex);
+		} finally {
+			connection.close();
+			skCase.releaseSingleUserCaseWriteLock();
+		}
+		return tagSetList;
+	}
+
+	/**
+	 * Inserts a row into the reports table in the case database.
```
rcordovano, line 75: reports -> tag sets

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
```diff
@@ -0,0 +1,458 @@
...
+
+				resultSet.next();
+				int setID = resultSet.getInt(1);
+
+				tagSet = new TagSet(setID, name);
+			}
+			connection.commitTransaction();
+		} catch (SQLException ex) {
+			connection.rollbackTransaction();
+			throw new TskCoreException(String.format("Error adding report %s", name), ex);
```
rcordovano, line 107: report -> tag set

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
```diff
@@ -0,0 +1,458 @@
...
+			connection.close();
+			skCase.releaseSingleUserCaseWriteLock();
+		}
+
+		return tagSet;
+	}
+
+	/**
+	 * Remove a row from the tag set table. All tags belonging to the given tag
+	 * set will also be removed.
```
rcordovano, line 118: We should actually call out that the tag names in the set will not be deleted, nor will any tags with the tag names from the deleted tag set be deleted.

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
```diff
@@ -0,0 +1,459 @@
...
+										resultSet.getString("login_name"));
+
+						removedTags.add(bat);
+					}
+				}
+
+				if (!removedTags.isEmpty()) {
+					// Remove the tags.
+					String removeQuery = String.format("DELETE FROM blackboard_artifact_tags WHERE tag_id IN (SELECT tag_id FROM blackboard_artifact_tags JOIN tag_names ON tag_names.tag_name_id = blackboard_artifact_tags.tag_name_id WHERE artifact_id = %d AND tag_names.tag_set_id = %d)", artifact.getId(), tagSetId);
+					try (Statement stmt = connection.createStatement()) {
```
rcordovano, line 226: I think we already have all of the tag IDs in hand from the previous query and can just make a CSV-list of the IDs instead of doing another SELECT for the IN clause. Or is this just being careful about changes that might have occurred since the previous query?

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
```diff
@@ -0,0 +1,459 @@
...
+	/**
+	 * Returns a list of all of the TagNames that are apart of the given TagSet.
+	 *
+	 * @param tagSetId ID of a TagSet.
+	 *
+	 * @return List of TagNames for the TagSet or empty list if none were found.
+	 *
+	 * @throws TskCoreException
+	 */
+	List<TagName> getTagNamesByTagSetID(int tagSetId) throws TskCoreException {
```
rcordovano, line 352: I believe that this method can be private.

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
kellykelly3: The original query was a little simpler, but didn't work, this was the solution on found on the internet.  I hadn't though about the fact I already had the ids.  I will change it to a comma separated list of ideas.  Good idea!

tsk/auto/db_sqlite.cpp
```diff
@@ -285,7 +285,7 @@ TskDbSqlite::initialize()
 	}
 
 	if (attempt_exec
-        ("CREATE TABLE tsk_objects (obj_id INTEGER PRIMARY KEY, par_obj_id INTEGER, type INTEGER NOT NULL, UNIQUE (obj_id), FOREIGN KEY (par_obj_id) references tsk_objects(obj_id) ON DELETE CASCADE);",
+        ("CREATE TABLE tsk_objects (obj_id INTEGER PRIMARY KEY, par_obj_id INTEGER, type INTEGER NOT NULL, UNIQUE (obj_id), FOREIGN KEY (par_obj_id) references tsk_objects(obj_id) ON DELETE SET NULL);",
```
rcordovano, line 288: This is a mistake. It would break data source deletion.

bindings/java/src/org/sleuthkit/datamodel/TaggingManager.java
```diff
@@ -0,0 +1,483 @@
...
+
+	/**
+	 * Add the given TagName to the TagSet.
+	 *
+	 * @param tagSet	 The tag set being added to.
+	 * @param tagName	The tag name to add to the set.
+	 *
+	 * @throws TskCoreException
+	 */
+	public void addTagNameToTagSet(TagSet tagSet, TagName tagName) throws TskCoreException {
```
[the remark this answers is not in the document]
kellykelly3: That does seem more consistent with existing code.

If we are going to return a TagSet a check should be made sure the tageName is not already in the TagSet.  If tagName is already in tagSet what the desired behavior?
1. Return the existing TagSet object
2. Make a copy of the passed in TagSet object
3. Throw an exception.

My inclination is to do 1 or 2.  I don't see a need to throw an exception unless you believe it would inform the coder that they maybe making a mistake.  Frankly I would lean towards 1 because nothing has changed.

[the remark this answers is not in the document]
rcordovano: We can't change the TagSet passed in because we have made TagSet objects immutable, and the tag set has changed - it now has a new tag name. #3 does not apply because it is perfectly fine to add a tag name to a tag set. We have made TagSet _objects_ immutable, but we have not made tag set read only after creation. #2 is the right approach.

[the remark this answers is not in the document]
kellykelly3: Two is what I wrote on Saturday.  :-)

PR #1897, opened 2020-04-22
6095: Persona API.

raman-bt:


raman-bt:
Closing this PR.  Central Repo will define its own Examiner class, to keep it unambiguous.

PR #1908, opened 2020-04-30
1442 image gallery tag changes

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TagName.java
```diff
@@ -33,39 +33,45 @@
...
+		YELLOW	("Yellow",	"#FFFF00"), //NON-NLS
+		OLIVE	("Olive",	"#808000"), //NON-NLS
+		LIME	("Lime",	"#00FF00"), //NON-NLS
+		GREEN	("Green",	"#008000"), //NON-NLS
+		AQUA	("Aqua",	"#00FFFF"), //NON-NLS
+		TEAL	("Teal",	"#008080"), //NON-NLS
+		BLUE	("Blue",	"#0000FF"), //NON-NLS
+		NAVY	("Navy",	"#000080"), //NON-NLS
+		FUCHSIA	("Fuchsia", "#FF00FF"), //NON-NLS
+		PURPLE	("Purple",	"#800080"); //NON-NLS
```
rcordovano, line 52: What do the '#' chars signify in these strings?

bindings/java/src/org/sleuthkit/datamodel/TagName.java
```diff
@@ -33,39 +33,45 @@
...
 			this.name = name;
 		}
 
 		String getName() {
 			return name;
 		}
+		
+		public String getHexColorCode() {
+			return hexString;
+		}
```
rcordovano, line 74: Would it make more sense to a client if this were named getRgbValue with a comment that the return value was in hex format?

kellykelly3: That is a better name.

bindings/java/src/org/sleuthkit/datamodel/TagName.java
kellykelly3: The # is part of the hex value.  I think it has something to do with the HTML specification?  I quick google, didn't tell me why the # is there other than its expected by the <style> tag.

rcordovano: It sounds like the '#' character is HTML-markup specific, then. I imagine you are trimming it off in IG code I haven't seen yet. I think we could probably drop it from the enum, but I don't think it much matters.

kellykelly3: Nope.  I don't trim it.  The color methods in both java se and javafx take it with the #.  In fact if we didn't have it on there I think I would have to add it, maybe.

PR #1915, opened 2020-05-08
6333 Store the data source ID in the tsk_fs_info table.

APriestman:
Simply getFileSystems method.


eugene7646:
@APriestman Codacy claims that "private long dataSourceId" in FileSystem.java is unused. Otherwise code looks good.

APriestman:
> @APriestman Codacy claims that "private long dataSourceId" in FileSystem.java is unused. Otherwise code looks good.

It is. I may just remove it. My feeling was that since we have it now we might as well store it in the DataSource object, but I don't want to change the public API by adding a getter for it.

PR #1918, opened 2020-05-08
6311 Batch file inserts

APriestman:


esaunders:
Is Codacy still not happy with this? Do we care?

APriestman:
Can't do anything about the too many fields one. It would probably be happier if the try-catch blocks were combined but then we'd lose displaying exactly what file an error occurred on. And I don't think there's much else in that logic that could change.

rcordovano:
@APriestman and @esaunders, see note in JIra.

PR #1926, opened 2020-05-12
6312 Folder prioritization

APriestman:


tsk/fs/fs_dir.c
```diff
@@ -18,6 +18,10 @@
 #include "tsk_fatfs.h"
 
 
+#ifdef TSK_WIN32
+#define strncasecmp _strnicmp
```
bcarrier, line 22: This can go into tsk_base.h

PR #1936, opened 2020-05-21
1457 Add getTagSet method to TagName

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -11507,7 +11507,7 @@ long getId() {
 				+ "FROM tsk_objects INNER JOIN blackboard_artifacts " //NON-NLS
 				+ "ON tsk_objects.obj_id=blackboard_artifacts.obj_id " //NON-NLS
 				+ "WHERE (tsk_objects.par_obj_id = ?)"),
-		INSERT_OR_UPDATE_TAG_NAME("INSERT INTO tag_names (display_name, description, color, knownStatus) VALUES (?, ?, ?, ?) ON CONFLICT (display_name) DO UPDATE SET description = ?, color = ?, knownStatus = ?"),
+		INSERT_OR_UPDATE_TAG_NAME("INSERT INTO tag_names (display_name, description, color, knownStatus) VALUES (?, ?, ?, ?) ON CONFLICT (display_name, tag_set_id) DO UPDATE SET description = ?, color = ?, knownStatus = ?"),
```
gdicristofaro, line 11510: This currently allows for overwriting the known_status values of standard tag sets.  For instance, I was able to make Bookmark's notable and similar.  Just double checking that is expected behavior.

kellykelly3: @rcordovano 
Greg has an interesting point, if the user when into the Tags Option panel and defined a new tag "Bookmark" they could change the property of the standard tags.  This wasn't an issue before because that was allow, but now nothing is preventing the user from redefining one of the "standard tags" in the options panel.

To fix this issue we would need to add code where ever the user can define tags to check the standard tags for name collision.

rcordovano: It was my understanding that we were not going to show the standard tags in the options panel any more, because they are no longer editable.

rcordovano: Actually, that's not the point, I just realized. Please write a bug report and move on.

kellykelly3: It's not the point and we are not showing them in the options panel, unless the user hit the add button and called the new tag "Bookmark" (or any of the standard tags) it would end up in the list of user defined tags.

PR #1941, opened 2020-05-28
6392 Send files to ingest stream

APriestman:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -537,6 +538,51 @@ public void run(String deviceId, String[] imageFilePaths, int sectorSize) throws
...
+							throw new TskCoreException("initAddImgNat returned a NULL TskAutoDb pointer");
+						}
+					}
+					if (imageHandle != 0) {
+						runAddImgNat(tskAutoDbPointer, deviceId, imageHandle, timeZone, imageWriterPath);
+					}
+				} finally {
+					releaseTSKReadLock();
+				}
+			}
```
rcordovano, line 585: With the "do nothing" implementation of AddDataSourceCallbacks, we can eliminate this code duplication. The older version of this method can call this one, supplying an instance of the "do nothing" AddDataSourceCallbacks class.

bindings/java/src/org/sleuthkit/datamodel/JniDbHelper.java
```diff
@@ -40,6 +40,7 @@
     
     private final SleuthkitCase caseDb;
     private CaseDbTransaction trans = null;
+    AddDataSourceCallbacks addDataSourceCallbacks;
```
rcordovano, line 43: This instance variable should be private and final.

bindings/java/src/org/sleuthkit/datamodel/JniDbHelper.java
```diff
@@ -380,45 +391,51 @@ private long addBatchedFilesToDb() {
                     logger.log(Level.SEVERE, "Error adding file to the database - parent object ID: " + computedParentObjId
                             + ", file system object ID: " + fileInfo.fsObjId + ", name: " + fileInfo.name, ex);
                     revertTransaction();
+                    batchedFiles.clear();
                     return -1;
                 }
             }
             commitTransaction();
+            
+            if (addDataSourceCallbacks != null) {
```
rcordovano, line 400: Null check not needed.

PR #1948, opened 2020-06-15
6425 Relax CommunicationManager's account validation

dannysmyda:


bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -1316,18 +1316,48 @@ int getAccountTypeId(Account.Type accountType) {
...
+	 * Normalize the given phone number to be suitable for storage.
+	 *
+	 * @param phoneNum The phone number to normalize
+	 *
+	 * @return The normalized phone number.
+	 */
+	private String normalizePhoneNumber(String phoneNumber) {
+		String strippedOfNonDigits = phoneNumber.replaceAll("\\D", "");
+
+		if (strippedOfNonDigits.isEmpty()) {
```
dannysmyda, line 1340: This whole if statement doesn't really make sense to me, this is just how it was previously.
Seems like we should instead throw a TskCoreException.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsUtils.java
```diff
@@ -58,64 +68,65 @@ public static String normalizePhoneNum(String phoneNum) throws TskCoreException
...
-	 * @return True if the given string is a valid phone number, false otherwise.
+	 * @return True if the given string is a valid phone number, false
+	 *         otherwise.
 	 */
 	public static boolean isValidPhoneNumber(String phoneNum) {
-		if (!StringUtils.isEmpty(phoneNum)) {
-			return phoneNum.matches("\\+?[0-9()\\-\\s]+");
+		if (!StringUtils.isBlank(phoneNum)) {
+			// Any number of digits is good enough.
+			return !phoneNum.replaceAll(NON_DIGITS, "").isEmpty();
```
raman-bt, line 115: I'm afraid this definition of a valid phone number is wrong.  

According to this, each of these is a valid phone number:
"1+1"
"! 1      3"
"2 #!!!!!%$#@ 3"
"2 #!sdaswfdsafasdfasda!!!!%$#@ "

I can understand our desire to relax rules for length (the old method didn't apply any length restrictions) , and allow some telephony signs (+ * # are the only valid chars allowed in a sequence of phone digits), but this applies no rules at all. 

We need to bring this back to functional parity with what was (leading +, digits, parenthesis, spaces), and allow '*' & '#'.

dannysmyda: Well these were all valid with the previous regex:
+-----
()()()()
-         -
)))))

At least now we know there are digits. Having worked with validating phone numbers before I can tell you its a rabbit hole.

raman-bt: Fair enough.  The input to this isn't totally junk , so its not likely any of above examples would be found that need to be validated as phone numbers.

The new method, however also says "dsmyda01@gmail.com" is a valid phone number.

The kind of input strings the CVT and communication artifact viewers  need to handle look like these:

5978620523@s.whatsapp.net
8:live:.cid.8f278e2e9be3ae57
128
bba1b3f8a10533b9	
kysm-g920v06157df651d9371e509fff3e650ff8d0	

Each of them is being reported as a valid phone number when they aren't.  

This change breaks the Artifact content viewers.

PR #1956, opened 2020-06-23
6459 package case json exporter

dannysmyda:


dannysmyda:
@rcordovano There is an equal split between Codacy warnings about unused private fields and classes missing a header comment. For the former, Gson will use reflections to access these, so they aren't unused. I can create getters for each field to make these go away, do you think its worth it? I'll go ahead and add a quick class header comment for each file.

rcordovano:
I don't think we should add getters and setters to seventy or so POJOs that exist solely to support JSON serialization. There are better uses of your time at this point.

PR #1981, opened 2020-07-20
6635 Throw an exception if there is no entry in the HandleCache for this c…

APriestman:
…ase.


bindings/java/src/org/sleuthkit/datamodel/SleuthkitJNI.java
```diff
@@ -49,6 +51,8 @@
  */
 public class SleuthkitJNI {
 
+	private static final Logger logger = Logger.getLogger(SleuthkitCase.class.getName());
```
bcarrier, line 54: Need to update class name.

APriestman: Oops. Fixed now.

PR #1998, opened 2020-08-12
Feature/qcow2

cmueller-tp:
I've rebased #886 to the current develop branch and fixed some memory leaks I've discovered.

I've had some trouble merging the vcxproj files as I'm not using VS/Windows. Could someone verify that the changes are working properly, please?

I'd really like to get this merged and am willing to write fuzzers and make a thorough code review if there is interest in merging this.

### DIFF
--- a/tsk/img/Makefile.am
+++ b/tsk/img/Makefile.am
@@ -4,7 +4,8 @@ EXTRA_DIST = .indent.pro
 noinst_LTLIBRARIES = libtskimg.la
 libtskimg_la_SOURCES = img_open.cpp img_types.c raw.c raw.h \
     aff.c aff.h ewf.cpp ewf.h tsk_img_i.h img_io.c mult_files.c \
-    vhd.c vhd.h vmdk.c vmdk.h img_writer.cpp img_writer.h
+    vhd.c vhd.h vmdk.c vmdk.h img_writer.cpp img_writer.h \
+    qcow2.c qcow2.h
 
 indent:
 	indent *.c *.h
--- a/tsk/img/img_open.cpp
+++ b/tsk/img/img_open.cpp
@@ -18,6 +18,8 @@
 
 #include "raw.h"
 
+#include "qcow2.h"
+
 #if HAVE_LIBAFFLIB
 #include "aff.h"
 #endif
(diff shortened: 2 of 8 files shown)


cmueller-tp:
how are things like this handled in TSK?
https://app.codacy.com/manual/brianc/sleuthkit/file/47418953796/issues/source?fileBranchId=19680258#l216

How would you like it to be resolved?
I guess the easiest and cleanest way is to use `if (set != NULL) {` in line 233 

The other issue is a bug in Codacy. The other union member are used after `li.QuadPart` is set.

I don't know what the cause is for the appveyor issue but I think it's not an issue with this PR.

What else do you need to get this merged?

rcordovano:
**how are things like this handled in TSK?**

In general, don't assign values that are always overwritten. The local variable can be declared but not initialized in the same statement if it is always initialized one way or another in the code that follows. 

**I guess the easiest and cleanest way is to use if (set != NULL) { in line 233**

That approach appears to create the potential for a null pointer de-reference:
image

**What else do you need to get this merged?**

I am a bit concerned by the copyright assertion:
"Copyright (c) 2017 Communication Security Establishment.  All rights reserved"
Has permission to open source this code been granted?

Ultimately, what gets merged is up to @bcarrier. If he tells us to go forward, the PR will be reviewed and tested by a developer at Basis Technology. Otherwise, the PR can be left open so that interested parties can use it, even if it does not become part of mainline TSK.

Thanks!

cmueller-tp:
> **how are things like this handled in TSK?**
> 
> In general, don't assign values that are always overwritten. The local variable can be declared but not initialized in the same statement if it is always initialized one way or another in the code that follows.
> 
> **I guess the easiest and cleanest way is to use if (set != NULL) { in line 233**
> 
> That approach appears to create the potential for a null pointer de-reference:
> image

Yeah, just checking both gives less room for errors.

> **What else do you need to get this merged?**
> 
> I am a bit concerned by the copyright assertion:
> "Copyright (c) 2017 Communication Security Establishment. All rights reserved"
> Has permission to open source this code been granted?

As the original author of this code has an open PR for this (#886), I guess so.

> Ultimately, what gets merged is up to @bcarrier. If he tells us to go forward, the PR will be reviewed and tested by a developer at Basis Technology. Otherwise, the PR can be left open so that interested parties can use it, even if it does not become part of mainline TSK.
> 
> Thanks!


Well the reason I'm working on it is that I'd really like to have qcow2 support in upstream TSK ;)
As I said, I'm willing to put work in this when I get the confirmation from @bcarrier that it will get merged. 

Have a nice day everyone!

cmueller-tp:
Oh, can someone tell my why appveyor is failing and Travis is not even triggering anymore?

PR #2008, opened 2020-08-17
6659 - Upgraded Ivy to 2.5.0 for Tika 1.24.1

dannysmyda:


dannysmyda:
> Are you also planning on taking care of the Case UCO build.xml file?

Yes, I upgraded the case-uco project as well.

PR #2021, opened 2020-09-14
Sleuthkit change add Blackboard artifact api with transaction

jayaramcs:


raman-bt:
@rcordovano: can this be merged now? particularly since it is not targeted to develop anymore and I need some of this work.

PR #2023, opened 2020-09-15
6806 Fix osx warnings

APriestman:


bcarrier:
I didn't review any of the logic changes, but I can confirm the warnings are gone.

PR #2025, opened 2020-09-17
6847 - Hook Domain Discovery UI up to Thumbnail API

dannysmyda:
As a result of 6847, thread interruption at the connection pool level was logging hundreds of warnings with messages confirming what is expected behavior (More details in JIRA 6847). This PR changes the cut off level of the pool to include only SEVEREs.


wschaeferB (approved):
It works, and those messages definitely do cause anything of value which is also logged to be quite hard to find, so it seems good to me. I personally don't see a problem with hiding non-severe messages from the com.mchange.v2.log package so I have approved it.

@rcordovano I would like your input on whether this is an approach we want to use, since when it goes into changing sleuthkit I start to question things.

rcordovano:
@wschaeferB, please refer to Jira for my answer to your excellent question.

PR #2038, opened 2020-10-05
6889 Add relationships on single transaction

APriestman:


rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -381,17 +383,50 @@ public void addRelationships(AccountFileInstance sender, List<AccountFileInstanc
...
+				break;
+			default:
+				throw new TskCoreException("Unknown DB Type: " + db.getDatabaseType().name());
+		}		
+		
+		CaseDbTransaction trans = db.beginTransaction();	
+		try {
+			SleuthkitCase.CaseDbConnection connection = trans.getConnection();
+			PreparedStatement preparedStatement = connection.getPreparedStatement(query, Statement.NO_GENERATED_KEYS);
+
```
rcordovano, line 405: I believe that both CaseDatabase and PreparedStatement are AutoCloseable so the Codacy issue regarding closing resources could be resolved with a try-with-resources.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -381,17 +383,50 @@ public void addRelationships(AccountFileInstance sender, List<AccountFileInstanc
...
+			case POSTGRESQL:
+				query = "INSERT " + query + " ON CONFLICT DO NOTHING";
+				break;
+			case SQLITE:
+				query = "INSERT OR IGNORE " + query;
+				break;
+			default:
+				throw new TskCoreException("Unknown DB Type: " + db.getDatabaseType().name());
+		}		
+
```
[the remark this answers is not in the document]
APriestman: No - the transaction gets the single user case lock and releases it in commit/rollaback
private CaseDbTransaction(SleuthkitCase sleuthkitCase, CaseDbConnection connection) throws TskCoreException {
	this.connection = connection;
	this.sleuthkitCase = sleuthkitCase;
	try {
		this.connection.beginTransaction();
	} catch (SQLException ex) {
		throw new TskCoreException("Failed to create transaction on case database", ex);
	}
	sleuthkitCase.acquireSingleUserCaseWriteLock();
}

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
APriestman: I can put the CaseDbConnection part in a try with resources, but the prepared statement really shouldn't be. The transaction holds onto those for future use and closes them itself when it is closed (yes that's at the same time in our case, but we still shouldn't be manually closing it).

APriestman: Actually the CaseDbTransaction closes the connection itself, so it seems pointless to put that in a try-with-resources. It's part of the CaseDbTransaction object, along with its prepared statements.

rcordovano (approved):

PR #2045, opened 2020-10-20
6914 Add support for implicit sparse blocks at end of files

APriestman:
Builds on https://github.com/sleuthkit/sleuthkit/pull/2040


APriestman:
Remaking as standalone

PR #2046, opened 2020-10-20
6914 Ext4 - Add implicit sparse blocks to end of file

APriestman:


esaunders:
Looks like Codacy is pointing out that while there is a NULL check on fs_attr on line 1598, there is an attempt to de-reference it on line 1596.

APriestman:
Weird - I thought I went in and fixed that. Maybe I never pushed it up. I'll change it.

tsk/fs/ext2fs.c
```diff
@@ -1585,6 +1585,49 @@ ext2fs_extent_tree_index_count(TSK_FS_INFO * fs_info,
     return count;
 }
 
+/** \internal
+* If the file length is longer than what is in the attr runs, add a sparse
+* data run to cover the rest of the file.
+*
+* @return 0 if successful or -1 on error.
```
esaunders, line 1592: Looks like only 0 or 1 get returned.

tsk/fs/ext2fs.c
```diff
@@ -1585,6 +1585,49 @@ ext2fs_extent_tree_index_count(TSK_FS_INFO * fs_info,
...
+            return 1;
+        }
+        data_run->offset = end_of_runs;
+        data_run->addr = 0;
+        data_run->len = total_file_blocks - end_of_runs;
+        data_run->flags = TSK_FS_ATTR_RUN_FLAG_SPARSE;
+
+        // Save the run.
+        if (tsk_fs_attr_add_run(fs_info, fs_attr, data_run)) {
+            return 1;
```
esaunders, line 1626: Do you need to free the attr run here?

esaunders (approved):

PR #2054, opened 2020-10-30
fix of possible memory overflow.

ihsinme:
according to the standard, the third parameter of the strncat function is not the size of the buffer, but the number of bytes that can be placed in the buffer without a terminating character.
https://en.cppreference.com/w/c/string/byte/strncat
```
 Appends at most count characters from the character array pointed to by src, stopping if the null character is found, to the end of the null-terminated byte string pointed to by dest. The character src [0] replaces the null terminator at the end of dest. The terminating null character is always appended in the end (so the maximum number of bytes the function may write is count + 1).
```
in your project file tsk_comparedir.cpp, there are places of unsafe use of the strncat function. if the source buffer is completely filled, zero writing may occur outside the buffer boundaries on the stack.

### DIFF
--- a/tools/autotools/tsk_comparedir.cpp
+++ b/tools/autotools/tsk_comparedir.cpp
@@ -161,18 +161,18 @@ uint8_t
     struct stat status;
 
     strncpy(fullPath, m_lclDir, TSK_CD_BUFSIZE);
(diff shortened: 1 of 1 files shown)


ihsinme:
what can I do with this PR?

APriestman (approved):

PR #2061, opened 2020-11-03
3979: add basic AnalysisResult infrastructure.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/AnalysisResult.java
```diff
@@ -0,0 +1,76 @@
...
+public class AnalysisResult extends BlackboardArtifact {
+
+	private final String conclusion;	// conclusion of analysis - may be empty
+	private final Score score;			// score from the analysis
+	private final String configuration; // name of a configuration file/element that guides this analysis, may be empty.
+	private final String justificaion;  // justification/explanation from the analysis, may be empty.
+
+	private boolean ignore_result = false;
+
+	public AnalysisResult(Score score, String conclusion, String configuration, String justificaion, SleuthkitCase sleuthkitCase, long artifactID, long sourceObjId, long artifactObjId, long dataSourceObjId, int artifactTypeID, String artifactTypeName, String displayName, ReviewStatus reviewStatus) {
```
bcarrier, line 36: Constructors can hopefully be package-scoped.

bindings/java/src/org/sleuthkit/datamodel/AnalysisResult.java
```diff
@@ -0,0 +1,76 @@
...
+
+	public AnalysisResult(Score score, String conclusion, String configuration, String justificaion, SleuthkitCase sleuthkitCase, long artifactID, long sourceObjId, long artifactObjID, long dataSourceObjID, int artifactTypeID, String artifactTypeName, String displayName, ReviewStatus reviewStatus, boolean isNew) {
+		super(sleuthkitCase, artifactID, sourceObjId, artifactObjID, dataSourceObjID, artifactTypeID, artifactTypeName, displayName, reviewStatus, isNew);
+		this.score = score;
+		this.conclusion = conclusion;
+		this.configuration = configuration;
+		this.justificaion = justificaion;
+	}
+
+	public String getConclusion() {
```
bcarrier, line 52: We should document if they return NULL or empty string if not set.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -123,7 +143,56 @@ public void postArtifacts(Collection<BlackboardArtifact> artifacts, String modul
...
+	/**
+	 * Adds new analysis result artifact.
+	 *
+	 * @param artifactType      Type of analysis result artifact to create.
+	 * @param obj_id            Object id of parent.
+	 * @param datasource_obj_id Data source object id.
+	 * @param score	            Score associated with this analysis result.
+	 * @param conclusion        Conclusion of the analysis, may be empty.
+	 * @param configuration     Configuration association with this analysis,
+	 *                          may be empty.
```
bcarrier, line 156: "empty" meaning empty string or NULL?

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -621,6 +630,10 @@ public BlackboardArtifact newArtifact(int artifactTypeID) throws TskCoreExceptio
 		throw new TskCoreException("Cannot create artifact of an artifact. Not supported.");
 	}
 
+	@Override
+	public AnalysisResult newAnalysisResult(BlackboardArtifact.Type artifactType, Score score, String conclusion, String configuration, String justification, Collection<BlackboardAttribute> attributesList) throws TskCoreException {
```
bcarrier, line 634: This should be allowed for extracted data.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
```diff
@@ -267,6 +269,23 @@ private void createArtifactTables(Statement stmt) throws SQLException {
...
 	}
 	
+	private void createAnalysisResultsTables(Statement stmt) throws SQLException  {
+		stmt.execute("CREATE TABLE tsk_analysis_result (obj_id " + dbQueryHelper.getBigIntType() + " NOT NULL, "
+				+ "conclusion TEXT, "
+				+ "significance INTEGER NOT NULL, "
+				+ "confidence INTEGER NOT NULL, "
+				+ "configuration TEXT, justification TEXT, "
+				+ "ignore_score INTEGER DEFAULT 0, " // boolean	
+				+ "FOREIGN KEY(obj_id) REFERENCES tsk_objects(obj_id) ON DELETE CASCADE"
```
bcarrier, line 279: Can this foreign key go to the artifacts table?

bindings/java/src/org/sleuthkit/datamodel/Report.java
```diff
@@ -236,6 +237,11 @@ public BlackboardArtifact newArtifact(int artifactTypeID) throws TskCoreExceptio
 		return db.newBlackboardArtifact(artifactTypeID, objectId);
 	}
 
+	@Override
+	public AnalysisResult newAnalysisResult(BlackboardArtifact.Type artifactType, Score score, String conclusion, String configuration, String justification, Collection<BlackboardAttribute> attributesList) throws TskCoreException {
```
bcarrier, line 241: If we allow them to make artifacts on reports, we can let them make analysis results I think.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -4590,6 +4672,70 @@ BlackboardArtifact newBlackboardArtifact(int artifact_type_id, long obj_id, Stri
...
+	 * @param conclusion Conclusion, may be empty.
+	 * @param configuration Configuration used by analysis, may be empty.
+	 * @param justification Justification, may be empty.
+	 * @param connection Database connection to use.
+	 * 
+	 * @return Analysis result.
+	 * 
+	 * @throws TskCoreException 
+	 */
+	public AnalysisResult newAnalysisResult(BlackboardArtifact.Type artifactType, long obj_id, long data_source_obj_id, Score score, String conclusion, String configuration, String justification, CaseDbConnection connection) throws TskCoreException {
```
bcarrier, line 4692: I think this should be package-scope or moved to Blackboard.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -3136,6 +3138,60 @@ public long getBlackboardArtifactsTypeCount(int artifactTypeID, long dataSourceI
...
+	 * @param artifactTypeID
+	 * @param dataSourceObjId Object id of the data source.
+	 * 
+	 * @return a list of analysis results.
+	 *
+	 * @throws TskCoreException If a critical error occurred
+	 *                          within tsk core and artifacts could not be
+	 *                          queried.
+	 */
+	public List<AnalysisResult> getAnalysisResults(int artifactTypeID, long dataSourceObjId) throws TskCoreException {
```
bcarrier, line 3156: I think this logic should go to Blackboard.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
raman-bt: I originally did have this FK refer to the blackboard_artifacts(artifact_obj_id) column.  

With SQLite, I was getting an error getting my INSERT_ANALYSIS_RESULT prepared statement, which kept failing with an error message "A problem occurred while trying to acquire a cached PreparedStatement in a background thread"

When I tried with PostGres, it gave me a more helpful message pointing out that column artifact_obj_id in the blackboard_artifacts is not UNIQUE.  When I modified the create table SQL for blackboard_artifacts to add a UNIQUE constraint to that column (perhaps we intentionally or unintentionally dropped that when we moved DB creation from C++ to Java?), it worked on PostGres.  

I will try again with SQLite and if adding the UNIQUE constraints works, I will set the FK to the blackboard_artifacts table.

raman-bt: Added a UNIQUE(artifact_object_id) constraint to blackboard_artifacts table and changes the FK constraint in tsk_analysis_results to blackboard_artifacts (artifact_object_id)

raman-bt:
@rcordovano, @bcarrier, @jayaramcs 

Could this PR me merged soon, keeping in mind this is not going into develop yet?

I need these changes along with what's on cyber_objects branch, and possibly additional changes to be made on top,  to test ingest of the (CT) 3.0 JSON

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -1183,7 +1186,21 @@ public BlackboardArtifact newArtifact(int artifactTypeID) throws TskCoreExceptio
 		return getSleuthkitCase().newBlackboardArtifact(artifactTypeID, getId(), dataSourceObjectId);
 	}
 
-
+	@Override
```
bcarrier, line 1189: Why do you need this versus using the one in AbstractContent?

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -123,7 +146,93 @@ public void postArtifacts(Collection<BlackboardArtifact> artifacts, String modul
...
+	 * @throws BlackboardException exception thrown if a critical error occurs
+	 *                             within TSK core
+	 */
+	public AnalysisResult newAnalysisResult(BlackboardArtifact.Type artifactType, long obj_id, long datasource_obj_id, Score score, String conclusion, String configuration, String justification, Collection<BlackboardAttribute> attributesList, CaseDbTransaction transaction) throws BlackboardException {
+		
+		try {
+			AnalysisResult analysisResult =  caseDb.newAnalysisResult(artifactType, obj_id, datasource_obj_id, score, conclusion, configuration, justification, transaction.getConnection());
+			
+			// add the given attributes
+			addBlackboardAttributes(attributesList, artifactType.getTypeID(), analysisResult.getArtifactID(), transaction);
```
jayaramcs, line 177: @raman-bt  I think this call is not a very good idea. I added this function but it is not appropriate - I think I will need to change it (thinking about the how to ) 
The issue here is that the attributes list will in fact get persisted, but the AnalysisResutl (the underlying BlackboardArtifact) will not cache the attributes that we just added. therefore the returned AnalysisResult will need to go back to the db to get the attributes back out. During ingest, this can be a performance constraint.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -123,7 +146,93 @@ public void postArtifacts(Collection<BlackboardArtifact> artifacts, String modul
...
+	 *
+	 * @throws TskCoreException exception thrown if a critical error occurs
+	 *                          within TSK core
+	 */
+	public List<AnalysisResult> getAnalysisResults(int artifactTypeID, long dataSourceObjId) throws TskCoreException {
+		
+		CaseDbConnection connection = caseDb.getConnection();
+		caseDb.acquireSingleUserCaseReadLock();
+		Statement s = null;
+		ResultSet rs = null;
```
jayaramcs, line 201: try with resources?

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
rcordovano: @jayaramcs and @raman-bt, is the plan to just pass the attributes through to the new SleuthkitCase.newAnalysisResult() method?

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -123,7 +146,93 @@ public void postArtifacts(Collection<BlackboardArtifact> artifacts, String modul
...
+	 *
+	 * @param artifactTypeID Type of results to get.
+	 * @param dataSourceObjId Data source to look under
+	 *
+	 * @return list of analysis results.
+	 *
+	 * @throws TskCoreException exception thrown if a critical error occurs
+	 *                          within TSK core
+	 */
+	public List<AnalysisResult> getAnalysisResults(int artifactTypeID, long dataSourceObjId) throws TskCoreException {
```
rcordovano, line 196: We should check the category of the artifact type passed in by the caller and immediately throw an exception if the type is not in the analysis result category.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -193,6 +195,15 @@ public String getArtifactTypeName() {
...
 	}
 
+	/**
+	 * Gets the artifact type category for this artifact.
+	 *
+	 * @return The artifact type category.
+	 */
+	public CategoryType getCategoryType() {
+		return ARTIFACT_TYPE.fromID(artifactTypeId).getCategory();
+	}
```
rcordovano, line 205: I don't think this method is needed at this time. Please remove it until we have a use case for it, at which point it would probably be better to just return the entire Type object and let the caller query the Type for its Category.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -4326,6 +4331,25 @@ String addSourceToArtifactAttribute(BlackboardAttribute attr, String source) thr
...
+	 * @param category		   Artifact type category.
+	 * 
+	 *
+	 * @return Type of the artifact added.
+	 *
+	 * @throws TskCoreException exception thrown if a critical error occurs
+	 * @throws TskDataException exception thrown if given data is already in db
+	 *                          within tsk core
+	 */
+	public BlackboardArtifact.Type addBlackboardArtifactType(String artifactTypeName, String displayName, BlackboardArtifact.CategoryType category) throws TskCoreException, TskDataException {
```
rcordovano, line 4352: We should not add new artifact-related methods to the public API of SleuthkitCase.  See email for discussion. This method should only be exposed via Blackboard.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
jayaramcs: @rcordovano @raman-bt I have created a new PR https://github.com/sleuthkit/sleuthkit/pull/2067 that allows for adding a blackboard artifact and its attributes in a single api call. I think it is more appropriate to use that API here.

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
raman-bt: We don't.  Removing this.

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -1183,7 +1186,21 @@ public BlackboardArtifact newArtifact(int artifactTypeID) throws TskCoreExceptio
...
+			AnalysisResult result = getSleuthkitCase().getBlackboard().newAnalysisResult(artifactType, getId(), dataSourceObjectId, score, conclusion, configuration, justification, attributesList, trans);
+			
+			trans.commit();
+			return result;
+		}
+		catch (BlackboardException ex) {
+			trans.rollback();
+			throw new TskCoreException("Error adding analysis result.", ex);
+		}
+	}
```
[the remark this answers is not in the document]
raman-bt: It is indeed. I overlooked the inheritance hierarchy.  Removed.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
raman-bt: @rcordovano : I don't quite follow this comment.   Could you please elaborate?

 addBlackboardArtifactType() is already a method in SleuthKitCase.  I did add an overloaded variant that takes in the Category as well (and modified existing overload to call the other with category EXTRACTED_DATA by default.

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -25,13 +25,16 @@
...
+import java.util.Collection;
 import java.util.List;
 import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TimeZone;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import org.sleuthkit.datamodel.Blackboard.BlackboardException;
+import org.sleuthkit.datamodel.SleuthkitCase.CaseDbTransaction;
```
rcordovano: I think all three of these imports are now unused.

bindings/java/src/org/sleuthkit/datamodel/AbstractContent.java
```diff
@@ -316,6 +319,22 @@ public BlackboardArtifact newArtifact(int artifactTypeID) throws TskCoreExceptio
...
+		CaseDbTransaction trans = db.beginTransaction();
+		try {
+			AnalysisResult result = db.getBlackboard().newAnalysisResult(artifactType, objId, this.getDataSource().getId(), score, conclusion, configuration, justification, attributesList, trans);
+			
+			trans.commit();
+			return result;
+		}
+		catch (BlackboardException ex) {
+			trans.rollback();
+			throw new TskCoreException("Error adding analysis result.", ex);
```
rcordovano, line 334: A little more debug info, e.g., the object ID of the artifact, might be helpful.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -4326,6 +4331,25 @@ String addSourceToArtifactAttribute(BlackboardAttribute attr, String source) thr
...
+	 * @param category		   Artifact type category.
+	 * 
+	 *
+	 * @return Type of the artifact added.
+	 *
+	 * @throws TskCoreException exception thrown if a critical error occurs
+	 * @throws TskDataException exception thrown if given data is already in db
+	 *                          within tsk core
+	 */
+	public BlackboardArtifact.Type addBlackboardArtifactType(String artifactTypeName, String displayName, BlackboardArtifact.Category category) throws TskCoreException, TskDataException {
```
rcordovano, line 4352: We should not add this as public API to SleuthkitCase. I have a task from Brian to do after this effort that entails "migrating" all public blackboard related methods from SleuthkitCase to Blackboard. The "migration" will not be a refactoring per se, because we will be deprecating the SleuthkitCase blackboard methods. Either putting this entirely in Blackboard or at least making this a package method instead of a public method will mean there will one less thing to deprecate.

rcordovano (changes requested):

rcordovano (approved):

PR #2067, opened 2020-11-09
API updates newBlackboardArtifact

jayaramcs:
- newBlackboardArtifact takes a collection of attributes and a boolean allowDuplicates flag transaction may be null. if it is null a new transaction will be created and used
- Removed Blackboard#addBlackboardAttributes method. This function does not update the Artifact object cache and can potentially cause issues if the caller is not aware of this.
- Added BlackboardArtifact#addAttributes method that does update its cache. The downside of this change is that the caller needs to load an Artifact into memory if it does not have it.  
This API is however safer 
 a) One cannot add arbitrarily add attributes to a nonexistent Artifact  
 b) The caller will have the opportunity to validate or duplicate check the attribute list etc.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
@@ -456,58 +456,59 @@ public final class Blackboard {
 
 
 	/**
-	 * Add a set of blackboard artifacts within a transaction context
-	 *
-	 * @param attributes	 A set of blackboard attribute.
-	 * @param artifactTypeId type of artifact associated with the attributes
-	 * @param artifactId     The artifact id of the blackboard artifact
-	 * @param transaction	 the transaction in the scope of which the operation
-	 *		                 is to be performed, managed by the caller
-	 *
-	 * @throws TskCoreException  thrown if a critical error occurs.
-	 */
-	public void addBlackboardAttributes(Collection<BlackboardAttribute> attributes, int artifactTypeId, long artifactId, SleuthkitCase.CaseDbTransaction transaction) throws TskCoreException {
-		if (attributes.isEmpty()) {
-			return;
-		}
-		if (transaction == null) {
-			throw new TskCoreException("Passed null CaseDbTransaction");
-		}
-		try {
-			SleuthkitCase.CaseDbConnection connection = transaction.getConnection();
-			for (final BlackboardAttribute attr : attributes) {
-				attr.setArtifactId(artifactId);
-				caseDb.addBlackBoardAttribute(attr, artifactTypeId, connection);
-			}
-		} catch (SQLException ex) {
-			throw new TskCoreException("Error adding blackboard attributes", ex);
-		}
-	}
-
-
-	/**
 	 * Add a new blackboard artifact with the given type.
 	 *
-	 * This api executes in the context of a transaction.
+	 * This api executes in the context of a transaction if one is provided.
 	 *
-	 * @param artifactType the type the given artifact should have
-	 * @param objId	   the content object id associated with this artifact
-	 * @param dataSourceObjId the data source object.
-	 * @param transaction  the transaction in the scope of which the operation
-	 *                     is to be performed, managed by the caller
+	 * @param artifactType    The type of the artifact.
+	 * @param sourceObjId     The content that is the source of this artifact.
+	 * @param dataSourceObjId The data source the artifact source content
+	 *                        belongs to, may be the same as the sourceObjId.
+	 * @param attributes      The attributes.
+	 * @param allowDuplicates Whether or not a duplicate artifact should be
+	 *                        created if the artifact already exists.
+	 * @param transaction     the transaction in the scope of which the
+	 *                        operation is to be performed. Null may be
+	 *                        provided, if one is not available. 
 	 *
 	 * @return a new blackboard artifact
 	 *
 	 * @throws TskCoreException exception thrown if a critical error occurs
 	 *                          within tsk core
 	 */
-	public BlackboardArtifact newBlackboardArtifact(BlackboardArtifact.Type artifactType, long objId, long dataSourceObjId, CaseDbTransaction transaction) throws TskCoreException {
+	public BlackboardArtifact newBlackboardArtifact(BlackboardArtifact.Type artifactType, long sourceObjId, long dataSourceObjId,
+			Collection<BlackboardAttribute> attributes, boolean allowDuplicates, final CaseDbTransaction transaction) throws TskCoreException {
+
+	    CaseDbTransaction localTrans = null;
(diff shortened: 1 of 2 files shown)


raman-bt:
@rcordovano: could you please review and merge this?  I need it to finish up addressing your review comments on 
https://github.com/sleuthkit/sleuthkit/pull/2061

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -384,6 +385,41 @@ public void addAttributes(Collection<BlackboardAttribute> attributes) throws Tsk
...
+	 *
+	 * @throws TskCoreException If an error occurs and the attributes were not
+	 *                          added to the artifact.
+	 */
+	public void addAttributes(Collection<BlackboardAttribute> attributes, final SleuthkitCase.CaseDbTransaction caseDbTransaction) throws TskCoreException {
+
+		if (attributes.isEmpty()) {
+			return;
+		}
+		if (caseDbTransaction == null) {
```
rcordovano, line 406: The transaction should be a required argument, as it is currently documented, and we should throw an IllegalArgumentException if it is not provided. If a caller wants to have an implicit transaction, they can just call addAttributes(Collection<BlackboardAttribute> attributes) directly.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -384,6 +385,41 @@ public void addAttributes(Collection<BlackboardAttribute> attributes) throws Tsk
...
+	 *                          caller
+	 *
+	 * @throws TskCoreException If an error occurs and the attributes were not
+	 *                          added to the artifact.
+	 */
+	public void addAttributes(Collection<BlackboardAttribute> attributes, final SleuthkitCase.CaseDbTransaction caseDbTransaction) throws TskCoreException {
+
+		if (attributes.isEmpty()) {
+			return;
+		}
```
rcordovano, line 405: We should probably throw an IllegalArgumentException if the collection is empty, since that is a programming error and we are doing the caller no favors by allowing it. This requires changing the conditional on line 497 in Blackboard.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -384,6 +385,41 @@ public void addAttributes(Collection<BlackboardAttribute> attributes) throws Tsk
...
+		try {
+			for (final BlackboardAttribute attribute : attributes) {
+				attribute.setArtifactId(artifactId);
+				attribute.setCaseDatabase(getSleuthkitCase());
+				getSleuthkitCase().addBlackBoardAttribute(attribute, artifactTypeId, caseDbTransaction.getConnection());
+			}
+			attrsCache.addAll(attributes);
+		} catch (SQLException ex) {
+			throw new TskCoreException("Error adding blackboard attributes", ex);
```
rcordovano, line 418: To help us if we ever have to do debugging that involves this code, please add the object ID and/or artifact ID of the artifact to the error message.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -455,59 +455,60 @@ private boolean attributesMatch(Collection<BlackboardAttribute> fileAttributesLi
...
+
+			if (isNewLocalTx) {
+				localTrans.commit();
+			}
+			return blackboardArtifact;
+
+		} catch (TskCoreException ex) {
+			if (isNewLocalTx) {
+				localTrans.rollback(); // An exception here is not caught. This can cause confusion 
+			}
```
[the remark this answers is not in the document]
jayaramcs: @rcordovano  The reason I did not log this is that this class does not have an initialized logger. I wasn't sure what standard is used in TSK for loggers. I shall create a logger and log this

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -455,59 +455,60 @@ private boolean attributesMatch(Collection<BlackboardAttribute> fileAttributesLi
...
-	 * @param dataSourceObjId the data source object.
-	 * @param transaction  the transaction in the scope of which the operation
-	 *                     is to be performed, managed by the caller
+	 * @param artifactType    The type of the artifact.
+	 * @param sourceObjId     The content that is the source of this artifact.
+	 * @param dataSourceObjId The data source the artifact source content
+	 *                        belongs to, may be the same as the sourceObjId.
+	 * @param attributes      The attributes.
+	 * @param allowDuplicates Whether or not a duplicate artifact should be
+	 *                        created if the artifact already exists.
```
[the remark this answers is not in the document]
jayaramcs: @rcordovano The use case here is to allow callers to make the inserts without incurring the additional cost of duplicate checks. CT keeps track of its Artifacts and is guaranteed to not insert duplicates. 
Based on the email discussions the API you proposed has the allowDuplicates flag in it - I assumed that it was the public API you were comfortable with.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -455,59 +455,60 @@ private boolean attributesMatch(Collection<BlackboardAttribute> fileAttributesLi
...
-			throw new TskCoreException("Passed null CaseDbTransaction");
+			localTrans = caseDb.beginTransaction();
+			isNewLocalTx = true;
+		}else {
+			localTrans = transaction;
 		}
-		return caseDb.newBlackboardArtifact(artifactType.getTypeID(), objId,
+
+		try {
+			// TODO: Perform duplicate checks if requested.
```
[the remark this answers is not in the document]
jayaramcs: @rcordovano please advise on the JRIA #. I can make the API change or the TODO change, I am neutral on the specific API form.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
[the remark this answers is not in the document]
rcordovano: Moving the conversation to email.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
jayaramcs: @rcordovano Did you mean a TSK exception or really mean an uncaught IllegalArgumentException?

rcordovano: I meant IllegalArgumentException. I sent you an email with the rationale.

PR #2072, opened 2020-11-10
3979: added getAnalysisResults() api to Content.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/AbstractContent.java
```diff
@@ -420,6 +420,23 @@ public BlackboardArtifact getGenInfoArtifact(boolean create) throws TskCoreExcep
...
+		return db.getBlackboard().getAnalysisResultsWhere(" obj_id = " + objId); //NON-NLS
+	}
+	
+
+	@Override
+	public List<AnalysisResult> getAnalysisResults(BlackboardArtifact.Type artifactType) throws TskCoreException {
+		
+		if (artifactType.getCategory() != BlackboardArtifact.Category.ANALYSIS_RESULT) {
+			throw new TskCoreException(String.format("Artifact type %s is not in analysis result catgeory.", artifactType.getTypeName()));
+		}
```
rcordovano, line 434: I believe this sort of checking for valid arguments is important. However, invalid arguments should result in an IllegalArgumentException, not a TSKCoreException. The error is a programming error, not a problem in the TSK Core; by throwing an IllegalArgumentException, we force the developer to fix the programming error rather than log and otherwise ignore it in the TskCoreException handling.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -237,6 +232,65 @@ public AnalysisResult newAnalysisResult(BlackboardArtifact.Type artifactType, lo
...
+	 *           the form "SELECT * FROM blackboard_artifacts WHERE XYZ".
+	 *
+	 * @return A list of BlackboardArtifact objects.
+	 *
+	 * @throws SQLException     Thrown if there is a problem iterating through
+	 *                          the result set.
+	 * @throws TskCoreException Thrown if there is an error looking up the
+	 *                          artifact type id
+	 */
+	private List<AnalysisResult> resultSetToAnalysisResults(ResultSet resultSet) throws SQLException, TskCoreException {
```
rcordovano, line 279: Please give the method docs a consistent style. These become part of online API documentation and a consistent style is a good thing in that setting. Personally, I prefer using sentence capitalization and punctuation. This is what was done for the method docs added to Content.

rcordovano (changes requested):

PR #2074, opened 2020-11-11
6910 Initial file repository work

APriestman:


dannysmyda (approved):

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -791,6 +802,13 @@ public String getMetaFlagsAsString() {
 		return str;
 	}
 
+	/**
+	 * @return The meta flags as stored in the database.
+	 */
+	public short getMetaFlagsAsInt() {
+		return TSK_FS_META_FLAG_ENUM.toInt(metaFlags);
+	}
```
rcordovano, line 810: This seems a little oddly named. Why are we not returning the enum as for getDirFlag()? Also, we are truncating the return value to a short.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
APriestman: we're getting the value that is stored in the database, and it's stored as a short. And the value isn't the enum, it's the set of values combined with an OR statement.

PR #2083, opened 2020-11-21
Data source updates 

jayaramcs:
Following new APIs have been introduced in **`DataSource`** 
- `void setAcquisitionToolDetails(DataSource datasource, String name, String version, String settings) throws TskCoreException `
- `String getAcquisitionToolSettings() throws TskCoreException;`
- `String getAcquisitionToolName() throws TskCoreException;`
- `String getAcquisitionToolVersion() throws TskCoreException;` 
- `Long getDateAdded() throws TskCoreException;`

**`Image`** and **`LocalFilesDataSource`** both have the same implementation that calls the API in **`SleuthkitCase`**

**`SleuthkitCase`**
- `Long getDataSourceInfoLong(DataSource datasource, String columnName) throws TskCoreException`
- `String getDataSourceInfoString(DataSource datasource, String columnName) throws TskCoreException `
I am not thrilled about these two new functions. I created them to reduce the code duplication and several new getters in SleuthkitCase class. Ideally, we'd want to update the Image constructor. However, `g/setAcquisitionDetails` apis follow a similar load/set on-demand model. 

@rcordovano @bcarrier Can you guys please review

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/DataSource.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/DataSource.java
@@ -73,15 +73,11 @@ public interface DataSource extends Content {
 	 * 
 	 * @throws TskCoreException Thrown if the data can not be written
 	 */
-	default void setAcquisitionDetails(String details) throws TskCoreException {
-		setAcquisitionDetails(details, null, null, null);
-	}
+	void setAcquisitionDetails(String details) throws TskCoreException;
 
 	/**
-	 * Sets the acquisition details along with any settings used as well as
-	 * module name and module version.
-	 *
-	 * @param details             The acquisition details
+	 * Sets the acquisition settings used as well as module name and module version.
+	 * 
 	 * @param acquisitionSettings Any settings specific to the acquisition. May
 	 *                            be Null.
 	 * @param moduleName          The module name. May be Null
@@ -89,7 +85,7 @@ public interface DataSource extends Content {
 	 *
 	 * @throws TskCoreException Thrown if the data can not be written
 	 */
-	public void setAcquisitionDetails(String details, String acquisitionSettings, String moduleName, String moduleVersion) throws TskCoreException;
+	public void setAcquisitionSettings(String acquisitionSettings, String moduleName, String moduleVersion) throws TskCoreException;
 	
 	/**
 	 * Gets the acquisition details field from the case database.
--- a/bindings/java/src/org/sleuthkit/datamodel/DataSource.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/DataSource.java
@@ -85,7 +85,7 @@ public interface DataSource extends Content {
 	 *
 	 * @throws TskCoreException Thrown if the data can not be written
 	 */
-	public void setAcquisitionSettings(String acquisitionSettings, String moduleName, String moduleVersion) throws TskCoreException;
+	void setAcquisitionSettings(String acquisitionSettings, String moduleName, String moduleVersion) throws TskCoreException;
 	
 	/**
 	 * Gets the acquisition details field from the case database.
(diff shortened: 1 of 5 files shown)


bindings/java/src/org/sleuthkit/datamodel/DataSource.java
```diff
@@ -83,4 +95,40 @@
...
+	String getAcquisitionSettings() throws TskCoreException;
+
+	/**
+	 * Gets the module name field from the case database.
+	 *
+	 * @return The module name
+	 *
+	 * @throws TskCoreException Thrown if the data can not be read
+	 */
+	String getModuleName() throws TskCoreException;
```
rcordovano, line 115: If module name and version are part of the acquisition settings, why have separate getter methods for the acquisition settings, module name and module version? We seem to be overloading the term "acquisition settings."

We could define an AcquistionSettings class and return an instance from a single getAcquisitionSettings method. If we pass an AcquistionSettings object to setAcquistionSettings as well, it would future-proof us if there is any possibility that AcquisitionSettings will expand in the future. 

Or maybe the acquisition settings string is actually module configuration?

Or maybe module name and module version are not acquisition settings at all?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -9285,6 +9314,42 @@ void setImageName(String name, long objId) throws TskCoreException {
...
+	 * @param sectorSize The sector size
+	 *
+	 * @throws TskCoreException Thrown if the values were not 0 in DB
+	 *
+	 */
+	void setImageSizes(Image image, long totalSize, long sectorSize) throws TskCoreException {
+
+		if (image.getSize() > 0 || image.getSsize() > 0) {
+			throw new TskCoreException(String.format("Error updating image sizes for object ID %d. Cannot update non-zero total size (%d) or sector size (%d)",image.getId(), image.getSize(), image.getSsize()));
+		}
```
rcordovano, line 9337: This method has a check-then-act race condition. Why is it necessary to enforce this restriction? Is seems like we are putting application/business logic in the data layer.

bindings/java/src/org/sleuthkit/datamodel/DataSource.java
```diff
@@ -74,6 +74,18 @@
...
 	 */
 	void setAcquisitionDetails(String details) throws TskCoreException;
+
+	/**
+	 * Sets the acquisition settings used as well as module name and module version.
+	 * 
+	 * @param acquisitionSettings Any settings specific to the acquisition. May
+	 *                            be Null.
+	 * @param moduleName          The module name. May be Null
+	 * @param moduleVersion       The module's version number. May be Null.
```
rcordovano, line 84: I think we should be more explicit about what module name and module version mean in this documentation. I actually don't know.

bindings/java/src/org/sleuthkit/datamodel/Image.java
```diff
@@ -539,6 +598,23 @@ public String getAcquisitionDetails() throws TskCoreException {
...
+	 * Updates the image's total size and sector size.This function may be used
+	 * to update the sizes after the image was created.
+	 *
+	 * Can only update the sizes if they were not set before. Will throw
+	 * TskCoreException if the values in the db are not 0 prior to this call.
+	 *
+	 * @param totalSize  The total size
+	 * @param sectorSize The sector size
+	 *
+	 * @throws TskCoreException Thrown if the values were not 0 in DB
```
rcordovano, line 611: Why is it necessary to enforce this restriction? Is seems like we are putting application/business logic in the data layer and there is a classical check-then-act race condition in the underlying SleuthkitCase method (see comment there).

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
jayaramcs: Fair question. The only reason I have is that I assumed that size is immutable. Once the value is set, we should not allow the change. 
But seeing this is business logic,  I will remove the condition.

bindings/java/src/org/sleuthkit/datamodel/DataSource.java
jayaramcs: @rcordovano At this stage, I am ok with having the module version and module name be part of the AcquistionSettings string (since I am putting a JSON string into settings) and remove the dedicated apis. 

This will eliminate the need for a new class and the settings can be whatever the modules need. 

Let me know if you prefer a new class instead.

rcordovano: @jayaramcs, I am good with either solution. I'd like to hear what @bcarrier prefers, since his design called for separate columns in the case database.

bindings/java/src/org/sleuthkit/datamodel/Image.java
bcarrier: The theory was that we only want someone to change the size if it was not set when the image was initially created.  So, the 0 check was to prevent any random module from updating the settings. I think it is a good safety check to minimize changes that any module can make. 

We have similar logic elsewhere.

bindings/java/src/org/sleuthkit/datamodel/DataSource.java
```diff
@@ -76,16 +76,16 @@
...
 
 	/**
-	 * Sets the acquisition settings used as well as module name and module version.
-	 * 
-	 * @param acquisitionSettings Any settings specific to the acquisition. May
-	 *                            be Null.
-	 * @param moduleName          The module name. May be Null
-	 * @param moduleVersion       The module's version number. May be Null.
+	 * Sets the acquisition tool details such as its name, version number and 
+	 * any settings used during the acquisition tool to acquire data.
```
rcordovano, line 80: Quick comment, this sentence should probably be: "Sets the acquisition tool details such as its name, version number and any settings used during the acquisition to acquire data."

rcordovano (approved):

PR #2086, opened 2020-11-24
6984 add GPS area artifact

APriestman:


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -1295,8 +1295,13 @@ public int hashCode() {
 		 * Indicates that the file had a yara pattern match hit.
 		 */
 		TSK_YARA_HIT(66, "TSK_YARA_HIT",	
-				bundle.getString("BlackboardArtifact.tskWebAccountType.text"));		
-
+				bundle.getString("BlackboardArtifact.tskWebAccountType.text")),
```
kellykelly3, line 1298: Make sure this merge works right. line 1298 is a bug that I fixed.

kellykelly3 (approved):

PR #2097, opened 2020-12-07
3993 score manager

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	/**
+	 * Recalculate and update the final score of the specified object.
+	 *
+	 * @param objId Object id.
+	 *
+	 * @return Final score of the object.
+	 */
+	public Score recalculateAggregateScore(long objId) throws TskCoreException {
```
rcordovano, line 177: I'm confused by this method. If we are updating/recalculating the aggregate score of an object every time a new analysis result is created, what is the use case for this method?

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/AggregateScoreChangedEvent.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+final public class AggregateScoreChangedEvent implements TskEvent {
+
+	public AggregateScoreChangedEvent(long dataSourceId, ImmutableSet<ScoreChange> scoreChanges) {
+		this.dataSourceId = dataSourceId;
+		this.scoreChanges = scoreChanges;
+	}
+
+	private final long dataSourceId;
+	private final ImmutableSet<ScoreChange> scoreChanges;
```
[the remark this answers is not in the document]
raman-bt: Added, I missed this new file in the commit.

bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
raman-bt: The updateAggregateScore() is to update the score as each new analysis result is being added - usually during ingest.

The intent for recalculateAggregateScore() is to be used, say if a user marks one of the analysis results  to be invalid/ignored.

rcordovano: Wouldn't it be better if invalidating an analysis result caused the AnalysisResult object to call recalculateAggregateScore(), in which case it does not need to be public API? Then the client doesn't have to have special knowledge of how scoring works, can't forget to call recalculateAggregateScore(), and the solution is more OO and less procedural as well?

Also, are we actually implementing this use case at this time, or are we committing to new public API speculatively?

bindings/java/src/org/sleuthkit/datamodel/AggregateScoresChangedEvent.java
```diff
@@ -0,0 +1,46 @@
...
+package org.sleuthkit.datamodel;
+
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * Event to indicate that aggregate score of objects has changed.
+ */
+final public class AggregateScoresChangedEvent implements TskEvent {
+
+	public AggregateScoresChangedEvent(long dataSourceId, ImmutableSet<ScoreChange> scoreChanges) {
```
bcarrier, line 29: This can probably be package scope.

bindings/java/src/org/sleuthkit/datamodel/ScoreChange.java
```diff
@@ -0,0 +1,53 @@
...
+ * This class encapsulates a score change.
+ */
+final public class ScoreChange {
+
+	private final long objId;
+	private final long dataSourceObjectId;
+	private final Score oldScore;
+	private final Score newScore;
+
+	public ScoreChange(long objId, long dataSourceObjectId, Score oldScore, Score newScore) {
```
bcarrier, line 31: Can be package scoped.

bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
```diff
@@ -0,0 +1,259 @@
...
+				break;
+			case SQLITE:
+				query = "INSERT OR IGNORE " + query;
+				break;
+			default:
+				throw new TskCoreException("Unknown DB Type: " + db.getDatabaseType().name());
+		}
+
+		try {
+			db.acquireSingleUserCaseReadLock();
```
bcarrier, line 156: Should be the write lock.

bindings/java/src/org/sleuthkit/datamodel/TskEvent.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.sleuthkit.datamodel;
+
+/**
+ *
+ * Defines an interface implemented by events published by Sleuthkit.
+ */
+public interface TskEvent {
```
bcarrier, line 25: I'd propose we name this something like TskEventWithDataSource since we may have events for reports, which are not always tied to a single DS.

rcordovano (changes requested):

rcordovano (approved):

PR #2100, opened 2020-12-10
7113 - Improve Domain Cache Performance

dannysmyda:


bindings/java/src/org/sleuthkit/datamodel/BlackboardAttribute.java
```diff
@@ -531,7 +542,7 @@ public String getDisplayString() {
 
 			case DATETIME: {
 				try {
-					final Content dataSource = getParentArtifact().getDataSource();
+					final Content dataSource = sleuthkitCase.getContentById(parentDataSourceID);
```
bcarrier, line 545: It's possible that ID isn't set,  This should have logic to get the ID from getParentArtifact() if it is not set.

PR #2101, opened 2020-12-11
Read unallocated ranges in APFSSpaceman only once

DrKelwin:
blkls for only unallocated blocks works very slow with APFS. The common problem is that `APFSFSCompat::block_walk` does `fs->block_getflags(fs, addr)` for each block to check is this block from unallocated range or not. `block_getflags()` re-reads full info about unallocated ranges each time it called. So, all the time blkls spends inside `APFSSpaceman::unallocated_ranges()`:

apfs_bug

On my test image I had only ~67mb after 5 minutes of blkls working.

In this PR I propose to read unallocated ranges only once. 
Also, I've added binary search instead of linear to check if block is in unallocated ranges. It's possible, because unallocated ranges are stored in a sorted order.

CPU time after:
apfs_fix

After this fix I had ~43gb after 5 minutes.


simsong:
Do you have a disk image that I can use to test this?

simsong:
I am closing and reopening this to get the tests to run.

simsong:
This is really nice, but I want an APFS disk to test it with.

simsong:
@uckelman-sf  - You're the C++ expert; can you take a look at this and let me know what you think?

uckelman-sf:
I'll have a look, but it may be a few days before I have time.

simsong:
Hi @DrKelwin . I really want to thank you for this contribution. However, I can't accept it  until we have better coverage of the patch in the unit tests.
image
As you can see, the majority of the lines in the patch are not tested, so I don't really know that they work.  Is there a way we can test this with our existing test images? If not, can you recommend a test image and some additional unit tests that we can do? Even if it is a new test image and we are just running fiwalk to auto_loaddb on it, that would be sufficient.

PR #2102, opened 2020-12-14
fix assignment instead of comparison

ihsinme:
there is a mistake in the code. which is difficult to notice. I suggest the following fix.

### DIFF
--- a/tsk/fs/yaffs.cpp
+++ b/tsk/fs/yaffs.cpp
@@ -3185,7 +3185,7 @@ TSK_FS_INFO *
     fs->block_post_size = yaffsfs->spare_size;
     fs->block_count = img_info->size / (fs->block_pre_size + fs->block_size + fs->block_post_size);
     fs->first_block = 0;
-    fs->last_block_act = fs->last_block = fs->block_count ? fs->block_count - 1 : 0;
+    fs->last_block_act = fs->last_block == fs->block_count ? fs->block_count - 1 : 0;
 
     /* Set the generic function pointers */
     fs->inode_walk = yaffsfs_inode_walk;


APriestman:
I believe the code is correct. In most of the file systems, the corresponding line looks like this:
fs->last_block = fs->last_block_act = fs->block_count - 1;
Which is setting both last_block and last_block_act to one less than the block count. The code in question is doing the same thing but with a check to make sure we're not setting the two last block fields to -1 if fs->block_count is zero.

ihsinme:
> I believe the code is correct. In most of the file systems, the corresponding line looks like this:
> fs->last_block = fs->last_block_act = fs->block_count - 1;
> Which is setting both last_block and last_block_act to one less than the block count. The code in question is doing the same thing but with a check to make sure we're not setting the two last block fields to -1 if fs->block_count is zero.

in my opinion your code may not be correct. I beg your pardon, but where there should be a comparison there cannot be an assignment.

APriestman:
The comparison is just this part here: fs->block_count ?
If block_count is non-zero, then set last_block and last_block_act to block_count -1. If it is zero, set last_block and last_block_act to zero.

rcordovano:
fs->last_block_act = fs->last_block = fs->block_count ? fs->block_count - 1 : 0;

is equivalent to:

if (fs->block_count) {
   fs->last_block = fs->block_count - 1; 
} else {
   fs->last_block = 0;
}
fs->last_bloc_act = fs->last_block;

It is a bit hard to read because of the ternary operator and the chained assignment. Parentheses might help humna parsing:

(fs->last_block_act = (fs->last_block = (fs->block_count ? fs->block_count - 1 : 0)));

ihsinme:
I'm sorry I wasted your time.

rcordovano:
@ihsinme, no need to apologize.

PR #2105, opened 2020-12-15
newAnalysisResult() api returns both - the newly added analysis resul…

raman-bt:
…t, as well as the current aggregate score of the object.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/AnalysisResultAdded.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/AnalysisResultAdded.java
@@ -36,7 +36,7 @@ public class AnalysisResultAdded {
 		return analysisResult;
 	}
 
-	public Score getScore() {
+	public Score getAggregateScore() {
 		return score;
 	}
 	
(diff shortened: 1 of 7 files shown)


bindings/java/src/org/sleuthkit/datamodel/AnalysisResultAdded.java
```diff
@@ -0,0 +1,43 @@
...
+	AnalysisResultAdded(AnalysisResult analysisResult, Score score) {
+		this.analysisResult = analysisResult;
+		this.score = score;
+	}
+	
+	public AnalysisResult getAnalysisResult() {
+		return analysisResult;
+	}
+
+	public Score getScore() {
```
bcarrier, line 39: Let's call this getAggregateScore() to make it more clear.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -182,9 +183,11 @@ public AnalysisResult newAnalysisResult(BlackboardArtifact.Type artifactType, lo
 			}
 			
 			// update the final score for the object 
-			caseDb.getScoringManager().updateFinalScore(objId, dataSourceObjId, analysisResult.getScore(), transaction);
+			Score aggregateScore = caseDb.getScoringManager().updateFinalScore(objId, dataSourceObjId, analysisResult.getScore(), transaction);
```
bcarrier, line 186: We should rename updateFileScore to updateAggregrateScore().

rcordovano (approved):

PR #2107, opened 2020-12-16
fix Memory leak in ext2fs_inode_walk() and unix_make_data_run_indirect()

DistantNight:
i find that code exit the ext2fs_inode_walk() without cleanup fs_file,
exit unix_make_data_run_indirect() without cleanup data_run.


bcarrier:
Thanks!  These have already been covered by other PRs at this point.

PR #2114, opened 2020-12-23
7132 other events

gdicristofaro:


bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -453,31 +461,48 @@ String getSQL() {
...
+				insertDescriptionStmt.setLong(3, artifactID);
+			}
+
+			insertDescriptionStmt.setString(4, fullDescription);
+			insertDescriptionStmt.setString(5, medDescription);
+			insertDescriptionStmt.setString(6, shortDescription);
+			insertDescriptionStmt.setInt(7, booleanToInt(hasHashHits));
+			insertDescriptionStmt.setInt(8, booleanToInt(tagged));
+			int row = insertDescriptionStmt.executeUpdate();
+			// if no inserted rows, return null.
```
rcordovano, line 495: I think we need a comment here (and at a few other points where we are dealing with null event description IDs) to explain that we got no rows because the description is already in the database and therefore making the event would be creating a duplicate would be helpful.

rcordovano (changes requested):

rcordovano (approved):

PR #2120, opened 2021-01-12
File addFilesystemFile api updated 

jayaramcs:
- add md5, sha256, mimetype, fileattributes, transaction 
- Add fileattributes to abstractfile constructor 
- updated derived classes  - passing in empty lists 
- [Documentation](https://docs.google.com/document/d/1vm5vx8oJiqXZHjl5xDoJ5rhS_KMa47WmscYnp2xeF9g/edit?ts=5fe27146)


bindings/java/src/org/sleuthkit/datamodel/AbstractAttribute.java
```diff
@@ -0,0 +1,378 @@
...
+	 * added to an appropriate artifact.
+	 *
+	 * @param attributeType The attribute type.
+	 * @param valueInt      The attribute value.
+	 *
+	 * @throws IllegalArgumentException If the value type of the specified
+	 *                                  attribute type is not
+	 *                                  TSK_BLACKBOARD_ATTRIBUTE_VALUE_TYPE.INTEGER.
+	 */
+	public AbstractAttribute(BlackboardAttribute.Type attributeType, int valueInt) throws IllegalArgumentException {
```
rcordovano, line 56: We really don't need the throws clause for a unchecked exception.  I realize this is legacy and it doesn't hurt, I just noticed it.

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -90,6 +94,8 @@
 	private static final ResourceBundle BUNDLE = ResourceBundle.getBundle("org.sleuthkit.datamodel.Bundle");
 	private long dataSourceObjectId;
 	private final String extension;
+	private final List<Attribute> fileAttributesCache = new ArrayList<Attribute>();
+	private final AtomicBoolean loadedAttributesCacheFromDb = new AtomicBoolean(false);
```
rcordovano, line 98: The flag for the cache is thread safe, but the cache itself is not. Which policy do we need?

Discussion to follow via email.

bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -499,8 +510,75 @@ public void setSha256Hash(String sha256Hash) {
...
+			throw new TskCoreException("Error adding file attributes", ex);
+		} finally {
+			if (isLocalTransaction && null != localTransaction) {
+				try {
+					localTransaction.rollback();
+				} catch (TskCoreException ex2) {
+					LOGGER.log(Level.SEVERE, "Failed to rollback transaction after exception", ex2);
+				}
+			}
+		}
```
rcordovano, line 579: The logic for committing or rolling back the transaction is kind of complex here. I think it can be simplified. How about putting the commit in the try block and the rollback in the catch block and eliminating the finally block?

bindings/java/src/org/sleuthkit/datamodel/BlackboardAttribute.java
```diff
@@ -483,33 +304,39 @@ public void addSource(String source) throws TskCoreException {
...
+			return true;
+		} else if (that instanceof BlackboardAttribute) {
+			BlackboardAttribute other = (BlackboardAttribute) that;
+			Object[] thisObject = new Object[]{this.getAttributeType(), this.getValueInt(), this.getValueLong(), this.getValueDouble(),
+				this.getValueString(), this.getValueBytes(), this.getSources()};
+			Object[] otherObject = new Object[]{other.getAttributeType(), other.getValueInt(), other.getValueLong(), other.getValueDouble(),
+				other.getValueString(), other.getValueBytes(), other.getSources()};
+
+			return Objects.deepEquals(thisObject, otherObject);
+		} else {
```
rcordovano, line 332: We have some duplicated code here. Perhaps we can create a protected helper in the super class for the comparison of the attribute type and values? Also, I do not see code to compare the unique fields of this class, i.e., context, sources, artifact ID, etc.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
```diff
@@ -460,6 +461,20 @@ private void createEventTables(Statement stmt) throws SQLException {
...
+		 * include embedded nulls so we have to pay attention to field length.
+		 * http://www.postgresql.org/docs/9.4/static/libpq-example.html
+		 */
+		stmt.execute("CREATE TABLE tsk_file_attributes (obj_id " + dbQueryHelper.getBigIntType() + " NOT NULL, "
+				+ "attribute_type_id " + dbQueryHelper.getBigIntType() + " NOT NULL, "
+				+ "value_type INTEGER NOT NULL, value_byte " + dbQueryHelper.getBlobType() + ", "
+				+ "value_text TEXT, value_int32 INTEGER, value_int64 " + dbQueryHelper.getBigIntType() + ", value_double NUMERIC(20, 10), "
+				+ "FOREIGN KEY(obj_id) REFERENCES tsk_files(obj_id) ON DELETE CASCADE, "
+				+ "FOREIGN KEY(attribute_type_id) REFERENCES blackboard_attribute_types(attribute_type_id))");
+	}
```
rcordovano, line 477: Do we need a PK here?

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/AbstractAttribute.java
```diff
@@ -0,0 +1,378 @@
...
+	 * Gets the value of this attribute. The value is only valid if the
+	 * attribute value type is TSK_BLACKBOARD_ATTRIBUTE_VALUE_TYPE.BYTE.
+	 *
+	 * @return The attribute value.
+	 */
+	public byte[] getValueBytes() {
+		return Arrays.copyOf(valueBytes, valueBytes.length);
+	}
+
+	public long getAttributeOwnerId() {
```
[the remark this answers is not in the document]
bcarrier: Or not make it public so that each implementation can give a more specific name for what it means.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
jayaramcs: @rcordovano  There is not one on BlackboardAttribute. Having one would be good for updates but probably bad since it will get indexed.

bindings/java/src/org/sleuthkit/datamodel/AbstractAttribute.java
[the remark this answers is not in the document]
rcordovano: @jayaramcs and @bcarrier, I like Brian's idea. The method in question can be made protected and superclass methods with less generic names can delegate to it.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
rcordovano: OK

bindings/java/src/org/sleuthkit/datamodel/AbstractAttribute.java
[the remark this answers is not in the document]
jayaramcs: @rcordovano I have already taken care of this.

rcordovano (approved):

PR #2133, opened 2021-01-25
Added a tsk_hosts table and a HostManager to add/get hosts.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/Host.java
```diff
@@ -0,0 +1,128 @@
...
+		this(id, name, HostStatus.ACTIVE);
+	}
+
+	Host(long id, String name, HostStatus status) {
+		this.id = id;
+		this.name = name;
+		this.status = status;
+	}
+
+	public long getId() {
```
rcordovano, line 45: Missing method docs on public methods that are not overrides. I think we should have something for the web page.

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -0,0 +1,170 @@
...
+	 * @param connection Database connection to use.
+	 * 
+	 * @return Optional with host.  Optional.empty if no matching host is found.
+	 * 
+	 * @throws TskCoreException 
+	 */
+	Optional<Host> getHost(String name, CaseDbConnection connection ) throws TskCoreException {
+		
+		String queryString = "SELECT * FROM tsk_hosts"
+							+ " WHERE LOWER(name) = LOWER('" + name + "')";
```
rcordovano, line 96: The use of lower here seems inconsistent with the UNIQUE constraint on name, which will be allow case-sensitive variations.

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -0,0 +1,170 @@
...
+
+		try (Statement s = connection.createStatement();
+				ResultSet rs = connection.executeQuery(s, queryString)) {
+
+			if (!rs.next()) {
+				return Optional.empty();	// no match found
+			} else {
+				return Optional.of(new Host(rs.getLong("id"), rs.getString("name")));
```
rcordovano, line 104: I think we need to pass the status column value to the Host constructor as well.

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -0,0 +1,170 @@
...
+	Set<Host> getHosts() throws TskCoreException {
+		String queryString = "SELECT * FROM tsk_hosts";
+
+		Set<Host> hosts = new HashSet<>();
+		try (CaseDbConnection connection = this.db.getConnection();
+				Statement s = connection.createStatement();
+				ResultSet rs = connection.executeQuery(s, queryString)) {
+
+			while (rs.next()) {
+				hosts.add(new Host(rs.getLong("id"), rs.getString("name")));
```
rcordovano, line 162: I think we need to pass the status column value to the Host constructor as well.

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -0,0 +1,170 @@
...
+
+		CaseDbConnection connection = transaction.getConnection();
+
+		// First search for host by name
+		Optional<Host> host = getHost(name, connection);
+		if (host.isPresent()) {
+			return host.get();
+		}
+
+		// could'nt find it, create a new host
```
rcordovano, line 79: could'nt => couldn't

bindings/java/src/org/sleuthkit/datamodel/Host.java
```diff
@@ -0,0 +1,128 @@
...
+		return id;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public void setName(String name) {
+		this.name = name;
+	}
```
rcordovano, line 55: See previous remark. I believe the name is immutable, once a host has been created. Otherwise, I think we will trip over the UNIQUE constraint.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
raman-bt: Host names are not case sensitive, I don't think we should force them to be.

In terms of implementation, this seemed like the best option - check ignoring the case, but then store what we got. (preserve case).

rcordovano: That sounds logical, but then we had better lower case the host name on INSERT.

bcarrier: This host name can also be whatever the user picks.  So, I think we should preserve whatever they type. IF they want a camel case name, we shoudl allow them too.  The future host addresses work is where actual host names will get stored.

raman-bt: That's what was referring to when i said "store what we get". Windows is a case preserving OS.   Forcing host names to go lower case goes against that.

rcordovano (approved):

PR #2139, opened 2021-01-29
Adding File Attributes table to upgrade code

jayaramcs:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -2299,10 +2299,15 @@ private CaseDbSchemaVersionNumber updateFromSchema8dot6toSchema8dot7(CaseDbSchem
...
+			String dateDataType   = "BIGINT";
+			String bigIntDataType = "BIGINT";
+			String blobDataType   = "BYTEA";
 			String primaryKeyType = "BIGSERIAL";
+
 			if (this.dbType.equals(DbType.SQLITE)) {
-				dateDataType = "INTEGER";
+				dateDataType   = "INTEGER";
+				bigIntDataType = "INTEGER";
+				blobDataType   = "BLOB";
```
rcordovano, line 2310: We have an interface and two implementations for this sort of stuff: SQLHelper. We should use that rather than define these strings here. 

 CaseDatabaseFactory uses SQLHelper.

rcordovano (changes requested):

rcordovano (approved):

APriestman (approved):

PR #2140, opened 2021-01-29
7261 Add support for storing hosts in the data_source_info table.

APriestman:
Added new methods to HostManager.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/HostManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/HostManager.java
@@ -51,6 +51,54 @@ public final class HostManager {
 		this.db = skCase;
 	}
 	
+	/**
+	 * APTODO : using for testing - might be able to remove if desired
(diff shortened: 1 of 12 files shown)


bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -51,6 +51,54 @@
 		this.db = skCase;
 	}
 	
+	/**
+	 * APTODO : using for testing - might be able to remove if desired
```
gdicristofaro, line 55: I think you could probably take this comment out.  I think this method will be useful in autopsy.

APriestman: Yes I think so - I used it in the temporary DSP code, and I'm pretty sure you'll need to do the same.

gdicristofaro (approved):

rcordovano (approved):

PR #2152, opened 2021-02-05
7264 public get hosts

gdicristofaro:


bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -285,22 +290,57 @@ public Host createHost(String name) throws TskCoreException  {
 			throw new TskCoreException(String.format("Error getting hosts"), ex);
 		}
 	}
-	
+
+	/**
+	 * Get all hosts with a particular status.
+	 *
+	 * @param status The status to filter on (if null, this will search for
```
rcordovano, line 297: The status will never be null in the case database. Passing null for host status should cause an IllegalArgumentException.

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -285,22 +290,57 @@ public Host createHost(String name) throws TskCoreException  {
...
+
+			try (ResultSet rs = preparedStatement.executeQuery()) {
+				while (rs.next()) {
+					hosts.add(new Host(rs.getLong("id"), rs.getString("name"), Host.HostStatus.fromID(rs.getInt("status"))));
+				}
+			}
+
+			return hosts;
+		} catch (SQLException ex) {
+			throw new TskCoreException(String.format("Error getting hosts"), ex);
```
rcordovano, line 325: Please include the host status we are querying for as potential debug info.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -208,99 +213,106 @@ public Host createHost(String name) throws TskCoreException  {
...
+	 *
 	 * @return Collection of hosts.
-	 * @throws TskCoreException 
+	 *
+	 * @throws TskCoreException
 	 */
-	public Set<Host> getHosts() throws TskCoreException {
-		String queryString = "SELECT * FROM tsk_hosts";
+	public List<Host> getHosts() throws TskCoreException {
+		String queryString = "SELECT * FROM tsk_hosts WHERE status = ? ";
```
APriestman, line 280: If we're only going to return active hosts, is there any reason to have the status as a parameter? It doesn't seem like it needs to be a prepared statement either way. Also the top comment block should be changed to indicate that only active hosts will be returned.

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -208,70 +213,73 @@ public Host createHost(String name) throws TskCoreException  {
...
-		
+	private Optional<Host> getHost(String name, CaseDbConnection connection) throws TskCoreException {
+
 		String queryString = "SELECT * FROM tsk_hosts"
-							+ " WHERE LOWER(name) = LOWER('" + name + "')";
-		try (
-				Statement s = connection.createStatement();
-				ResultSet rs = connection.executeQuery(s, queryString)) {
+				+ " WHERE LOWER(name) = LOWER(?)";
+		try (PreparedStatement s = connection.getPreparedStatement(queryString, Statement.RETURN_GENERATED_KEYS)) {
```
APriestman, line 256: It looks like this try-with-resources is resulting in the transaction being closed when it shouldn't be.

APriestman (approved):

PR #2154, opened 2021-02-05
7234 tl handle multiple time attributes

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -519,6 +511,44 @@ private long addEventDescription(long dataSourceObjId, long fileObjId, Long arti
...
+			String fullDescription, CaseDbConnection connection) throws TskCoreException {
+
+		String query = "SELECT event_description_id FROM tsk_event_descriptions "
+				+ "WHERE data_source_obj_id = " + dataSourceObjId
+				+ " AND content_obj_id = " + fileObjId
+				+ " AND artifact_id = " + artifactID
+				+ " AND full_description " + (fullDescription != null ? "= '" + fullDescription + "'" : "IS null");
+
+		caseDB.acquireSingleUserCaseReadLock();
+		try (ResultSet resultSet = connection.createStatement().executeQuery(query)) {
```
gdicristofaro, line 538: It might be safer to use a PreparedStatement here.

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -650,7 +680,7 @@ private long addEventDescription(long dataSourceObjId, long fileObjId, Long arti
 							.ifPresent(newEvents::add);
 				} catch (DuplicateException ex) {
 					duplicateExists = true;
-					logger.log(Level.SEVERE, getDuplicateExceptionMessage(artifact, "Attempt to make artifact event duplicate"), ex);
+					logger.log(Level.INFO, getDuplicateExceptionMessage(artifact, "Attempt to make artifact event duplicate"), ex);
```
gdicristofaro, line 683: I think this exception will occur only on double-posting of the same event information, which should only happen if some ingest module erroneously double-posts the same artifact.  Shouldn't that be left as 'SEVERE'?

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
```diff
@@ -659,7 +689,7 @@ private long addEventDescription(long dataSourceObjId, long fileObjId, Long arti
 				try {
 					addOtherEventDesc(artifact).ifPresent(newEvents::add);
 				} catch (DuplicateException ex) {
-					logger.log(Level.SEVERE, getDuplicateExceptionMessage(artifact, "Attempt to make 'other' artifact event duplicate"), ex);
+					logger.log(Level.INFO, getDuplicateExceptionMessage(artifact, "Attempt to make 'other' artifact event duplicate"), ex);
```
gdicristofaro, line 692: Same for this log message.

gdicristofaro (changes requested):

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
kellykelly3: You could make it a warning, but it's not hurting anything.  We are not losing information,   @rcordovano can decide.

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
kellykelly3: The PreparedStatements do not buy us anything unless we are saving them for future use.  Also I tried it with a prepared statement and it didn't work.  This worked.

gdicristofaro: I'm fairly certain a string like "Richard's Phone Call" for a description will cause a failure because of the single quote.

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
rcordovano: We should reserve INFO for things that do not involve errors. In this case, the double posting is an error, so INFO is not appropriate.

Logging a WARNING here is intuitive and would be the right thing to do except for one development process catch. Our regression tests log new SEVERE messages, but not new WARNINGs. So we need to log SEVERE to be sure the programming error for a double posting comes to our attention. This does seem to come up fairly often in log level discussions.

I think that we can consider changing out process. That would require a script change and new "gold." We should write a story about the process change and make this a SEVERE message until we can do the new story, which can reference this code.

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
kellykelly3: While I don't disagree with your point. We control the code the code that creates the description therefore I don't understand how the quotes appeared in the first place.  We have many places in the code with select statements similar to mine that use single quotes.

bindings/java/src/org/sleuthkit/datamodel/TimelineManager.java
kellykelly3: The only reason this is SEVERE is that its the only way to catch that some code is erroneously calling Blackboard.postArtifact multiple times with the same artifact? 

I will change it back to SEVERE.

wschaeferB (approved):

PR #2155, opened 2021-02-07
4039 os account - adapting realms and accounts to the specification doc.

raman-bt:


raman-bt:
@jayaramcs : for your early preview.  This implements the createWindowsRealm() method as specified in the doc, but does not have createWindowsAccount() yet.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -132,14 +131,13 @@ public OsAccount createOsAccount(String uniqueAccountId, String loginName, Strin
...
 	 * @param host            Host for the realm, may be null.
-	 * @param transaction     Transaction to use for database operation.
 	 *
 	 * @return Optional with OsAccount matching the given uniqueId, or
 	 *         realm/login name. Optional.empty if no match is found.
 	 * 
 	 * @throws TskCoreException If there is an error getting the account.
 	 */
-	public Optional<OsAccount> getOsAccount(String uniqueAccountId, String loginName, String realmName, Host host, CaseDbTransaction transaction) throws TskCoreException {
+	public Optional<OsAccount> getOsAccount(String uniqueAccountId, String loginName, String realmName, Host host) throws TskCoreException {
```
bcarrier, line 140: This can go away.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -237,17 +237,17 @@ private OsAccount createOsAccount(String uniqueId, String loginName, OsAccountRe
...
-	 * @param transaction Transaction to use for database connection.
-	 *
 	 *
 	 * @return Optional with OsAccount, Optional.empty if no matching account is
 	 *         found.
 	 *
 	 * @throws TskCoreException If there is an error getting the account.
 	 */
-	public Optional<OsAccount> getOsAccount(String uniqueId, Host host, CaseDbTransaction transaction) throws TskCoreException {
+	public Optional<OsAccount> getOsAccount(String uniqueId, Host host) throws TskCoreException {
```
bcarrier, line 246: This can become private.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -506,24 +505,23 @@ public OsAccount createOsAccountByWindowsSID(String sid, Host host) throws TskCo
...
 	 * 
 	 * @return Optional with OsAccount, Optional.empty if no matching OsAccount is found.
 	 * 
 	 * @throws TskCoreException 
 	 */
-	public Optional<OsAccount> getOsAccountByWindowsSID(String sid, Host host, CaseDbTransaction transaction) throws TskCoreException {
-
-		CaseDbConnection connection = transaction.getConnection();
-
+	public Optional<OsAccount> getOsAccountByWindowsSID(String sid, Host host) throws TskCoreException {
```
bcarrier, line 513: This can go away

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -584,44 +581,40 @@ public OsAccount createOsAccountByLogin(String loginName, String realmName, Host
...
 	 * @param host        Host for the realm.
-	 * @param transaction Transaction to use.
 	 *
 	 * @return Optional with OsAccount, Optional.empty if no matching OS account
 	 *         is found.
 	 *
 	 * @throws TskCoreException
 	 */
-	public Optional<OsAccount> getOsAccountByLogin(String loginName, String realmName, Host host, CaseDbTransaction transaction) throws TskCoreException {
+	public Optional<OsAccount> getOsAccountByLogin(String loginName, String realmName, Host host) throws TskCoreException {
```
bcarrier, line 590: This can become private

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -824,22 +814,21 @@ private OsAccount osAccountFromResultSet(ResultSet rs, OsAccountRealm realm) thr
...
+	 * loginName: if the account has a login name.
 	 *
-	 * @param realm     Account realm
 	 * @param uniqueId  Unique id.
 	 * @param loginName Login name.
 	 *
 	 * @return Account signature.
 	 */
-	private String makeAccountSignature(OsAccountRealm realm, String uniqueId,  String loginName) {
+	private String makeAccountSignature(String uniqueId,  String loginName) {
```
bcarrier, line 825: This could probably be just updateSignature() and refer to the member variables, right?

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealm.java
```diff
@@ -143,4 +221,52 @@ public static RealmNameType fromID(int typeId) {
...
+
+	/**
+	 * Set the realm addr if it is not already set.
+	 *
+	 * @param addr Realm addr to set.
+	 */
+	public void setRealmAddr(String addr) {
+		if (Objects.isNull(this.realmAddr) && Objects.nonNull(addr)) {
+			this.realmAddr = addr;
+			this.signature = addr;	// update signature also, addr is more definitive than name.
```
bcarrier, line 244: Shoudl make an updateSignature() method.

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealm.java
```diff
@@ -143,4 +221,52 @@ public static RealmNameType fromID(int typeId) {
...
 	}
 	
+	/**
+	 * Set the realmName if it is not already set.
+	 * 
+	 * @param name Realm name to set.
+	 */
+	public void setRealmName(String name) {
+		if (Objects.isNull(this.realmName) && Objects.nonNull(name)) {
+			this.realmName = name;
```
bcarrier, line 231: call updateSignature

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealm.java
```diff
@@ -143,4 +221,52 @@ public static RealmNameType fromID(int typeId) {
...
+			this.isDirty = true;
+		}
+	}
+
+	/**
+	 * Set the realm scope host if it is not already set.
+	 *
+	 * @param host Realm scope host to set.
+	 */
+	public void setHost(Host host) {
```
bcarrier, line 254: Let's comment this and the setScopeConfidence() out for now.

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -63,103 +61,126 @@
 	 *
 	 * @throws TskCoreException If there is an error creating the realm.
 	 */
+	
+	// RAMAN TBD: does this go away ??? Confirm with Brian.
 	public OsAccountRealm createRealmByName(String realmName, Host host) throws TskCoreException {
```
bcarrier, line 66: Yes, this can at least be not public or go away if we do not need it internally.

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -63,103 +61,126 @@
...
+		if (!Strings.isNullOrEmpty(accountSid)) {
+			realmAddr = getSubAuthorityId(accountSid);
+			signature = realmAddr;
+		}
+		else {
+			signature = realmName;
 		}
 		
+		// create a realm
+		return createRealm(realmName, realmAddr, signature, scopeHost, scopeConfidence);
```
bcarrier, line 144: I don't think we should pass in signature. createRealm() should do that by calling updateSignature()

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
raman-bt: Gone.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
raman-bt: Umm, signature is a member in OsAccount, this is OsAccountManager.  

We have to write the signature column in the database before we even create a OsAccount object.

It could certainly use a different name now - since there is no more "making"

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealm.java
raman-bt: I intentionally didn't update signature here.

If we are updating the name -  it mean name was earlier null, which means we knew the addr and hence  signature was set to addr.

Do we want to go from addr to name, knowing that addr is more definitive?

bindings/java/src/org/sleuthkit/datamodel/TskCaseDbBridge.java
```diff
@@ -374,35 +375,29 @@ private long addBatchedFilesToDb() {
...
 					}
 
-					// RAMAN TBD: Need to get host by using the data source name and then use that host for creating the OS account below.
-					Host host = null;
-
 					// query the DB to get the owner account
-					Optional<OsAccount> ownerAccount = caseDb.getOsAccountManager().getOsAccount(ownerUid, host, trans);
+					
+					// RAMAN TBD: what should be realm name used here
+					String realmName = "DUMMY";
```
bcarrier, line 381: We should pass in null instead of a made up name.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
```diff
@@ -436,25 +436,26 @@ private void createAccountTables(Statement stmt) throws SQLException {
...
 		stmt.execute("CREATE TABLE tsk_os_account_realms (id " + dbQueryHelper.getPrimaryKey() + " PRIMARY KEY, "
-				+ "name TEXT NOT NULL, "	// realm name - host name or domain name
+				+ "realm_name TEXT DEFAULT NULL, "	// realm name - for a domain realm, may be null
 				+ "realm_addr TEXT DEFAULT NULL, "		// a sid/uid or some some other identifier, may be null
-				+ "host_id " + dbQueryHelper.getBigIntType() + " DEFAULT NULL, " // if the realm just comprises of a single host
-				+ "name_type INTEGER, "	// indicates if the realm name was  was expressed or inferred 
-				+ "UNIQUE(name, host_id), "
-				+ "FOREIGN KEY(host_id) REFERENCES tsk_hosts(id) )");
+				+ "realm_signature TEXT NOT NULL, "	// is either the realm_addr or the realm_addr, one of them must be known.  This is used to prevent duplicates.
+				+ "scope_host_id " + dbQueryHelper.getBigIntType() + " DEFAULT NULL, " // if the realm scope is a single host
```
rcordovano, line 442: I think this comment should say that the signature can be either the realm name or the realm address.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
```diff
@@ -436,25 +436,26 @@ private void createAccountTables(Statement stmt) throws SQLException {
...
+				+ "scope_confidence INTEGER, "	// indicates whether we know for sure the realm scope or if we are inferring it
+				+ "UNIQUE(realm_signature, scope_host_id), "
```
rcordovano, line 444: @raman-bt and @bcarrier, I think we have a problem with this UNIQUE constraint. At least for SQLite, NULL values in the same column are not equivalent. See sqlitetutorial.net/sqlite-unique-constraint/. So these two rows would be possible, even with this constraint:
S-1-5-21, NULL
S-1-5-21, NULL

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
```diff
@@ -245,12 +253,42 @@ public static OsAccountInstanceType fromID(int typeId) {
...
+	 * Sets the account user's full name, such as "John Doe"
 	 *
 	 * @param fullName Full name.
 	 */
 	public void setFullName(String fullName) {
-		this.fullName = fullName;
+		if (this.fullName == null) {
+			this.fullName = fullName;
+			this.isDirty = true;
+		}
```
rcordovano, line 291: @bcarrier and @raman-bt, just want to make sure that you have agreed that we want to silently ignore caller input if login name, unique name, etc. have already been set, either by the current caller or some other caller, possibly on a different machine. This could be confusing for the calling code. Alternatives would be to throw an exception or allow overwrites.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -55,121 +54,123 @@
...
+	OsAccount createOsAccount(String uniqueAccountId,  long realmId) throws TskCoreException {
+
+		// ensure unique id is provided
+		if (Strings.isNullOrEmpty(uniqueAccountId)) {
+			throw new IllegalArgumentException("Cannot create OS account with null uniqueId.");
+		}
+		
+		try (CaseDbConnection connection = this.db.getConnection();) {
+
+			OsAccountRealm realm = db.getOsAccountRealmManager().getRealm(realmId, connection);
```
rcordovano, line 78: @raman-bt and @bcarrier, this may be a fine point, but do we want to create coupling between the OsAccountManager and the OsRealmManager? It seems to be that it would be a cleaner design, at least according to my personal understanding of classical design principles, to have a clear separation of concerns / no coupling.  The client could pass in an OsAccountRealm obtained by making the call themselves. Such an approach would also make error handling easier when all exceptions are TskCoreExceptions. With this code as is, the user does not know whether an exception is coming from the code to look up the realm, or the code to create the account. The client's response may need to be different for each error condition, but that is not possible here.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -55,121 +54,123 @@
...
+	 * @param referringHost Host referring the account.
+	 * @param realmScope    Realm scope.
 	 *
 	 * @return OsAccount.
 	 *
 	 * @throws TskCoreException If there is an error in creating the OSAccount.
 	 *
 	 */
-	public OsAccount createOsAccount(String uniqueAccountId, String loginName, String realmName, Host host) throws TskCoreException {
+	public OsAccount createWindowsAccount(String sid, String loginName, String realmName, Host referringHost, OsAccountRealm.RealmScope realmScope) throws TskCoreException {
```
jayaramcs, line 115: @raman-bt can you please add IllegalArgumentException to the throws part in the documentation? 
There are prob other methods too, but the users of the api will not know without looking at code.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
rcordovano: @raman-bt and @bcarrier, actually, I see that this method is package scope but has no callers. If we think we are going to need it later, perhaps we should comment it out to show our intent? In any case. I think my question in the previous comment still applies, so that I can orient myself better to what our goals are.

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
raman-bt: Addressed uniqueness by making a signature string as follows:
(realmAddr|realmName)_(hostId|"DOMIAN")

Fixed the UNIQUE constraint on table to be only on signature column.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
raman-bt: w.r.t to coupling, A realm is a part of an Account so there is going to be some "coupling" in the form of using apis.

In this instance however I agree that instead of the realmRowID, we can pass the realm itself.  Purely from efficiency point of view, we are wasting cycles - getting the realm, passing the id, then getting the realm again downstream. 

There are other instances where the doc specifies realmRowId as an argument but I replaced it with the  accountRealm.  I shall do the same here.

It is indeed not used from anywhere yet.  I don't know what usecase Brian has in mind.

rcordovano: "Purely from efficiency point of view, we are wasting cycles - getting the realm, passing the id, then getting the realm again downstream."

I'm sorry, I don't follow this argument. I assumed we would pass in an OsAccountRealm object instead of an ID. What am I missing?

raman-bt: I did change it to pass the OsAccountRealm instead of the id.   I'm in  agreement with you albeit for a different reason.  

I meant passing in the id when we already have the object in hand, and then the API fetching the object again is inefficient.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -55,121 +54,125 @@
...
+				// search by loginName
+				if (!Strings.isNullOrEmpty(loginName)) {
+					osAccount = getOsAccountByLoginName(loginName, realm.get());
+					if (osAccount.isPresent()) {
+						return osAccount.get();
+					}
+				}
+
+				// create failed for some other reason, throw an exception
+				throw new TskCoreException(String.format("Error creating OsAccount with loginName = %s", loginName), ex);
```
rcordovano, line 170: Are we throwing away some additional information here that might help debugging or is login name sufficent?

PR #2168, opened 2021-02-12
4121 new data artifact

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/Content.java
```diff
@@ -195,6 +195,21 @@
...
+	 *                       artifact.
+	 * @param osAccount      The OS account associated with the artifact. May be
+	 *                       null.
+	 *
+	 * @return DataArtifact New data artifact.
+	 *
+	 * @throws TskCoreException If a critical error occurred within tsk core.
+	 */
+	public DataArtifact newDataArtifact(BlackboardArtifact.Type artifactType, Collection<BlackboardAttribute> attributesList, OsAccount osAccount) throws TskCoreException;
+
```
rcordovano, line 212: @bcarrier, @jayaramcs and @raman-bt....

Given the move to DTOs, perhaps we do not want to add this convenience API that requires calls through SleuthkitCase?

bindings/java/src/org/sleuthkit/datamodel/DataArtifact.java
```diff
@@ -49,13 +49,14 @@
 	 * @param reviewStatus     The review status of this artifact.
 	 * @param osAccount        OsAccount associated with this artifact, may be
 	 *                         null.
+	 * @param isNew            The object is newly created.
 	 */
-	DataArtifact(SleuthkitCase sleuthkitCase, long artifactID, long sourceObjId, long artifactObjId, long dataSourceObjId, int artifactTypeID, String artifactTypeName, String displayName, ReviewStatus reviewStatus, OsAccount osAccount) {
-		super(sleuthkitCase, artifactID, sourceObjId, artifactObjId, dataSourceObjId, artifactTypeID, artifactTypeName, displayName, reviewStatus);
+	DataArtifact(SleuthkitCase sleuthkitCase, long artifactID, long sourceObjId, long artifactObjId, long dataSourceObjId, int artifactTypeID, String artifactTypeName, String displayName, ReviewStatus reviewStatus, OsAccount osAccount, boolean isNew) {
+		super(sleuthkitCase, artifactID, sourceObjId, artifactObjId, dataSourceObjId, artifactTypeID, artifactTypeName, displayName, reviewStatus, isNew);
```
rcordovano, line 55: @bcarrier, @jayaramcs and @raman-bt...

What is the use case for the  isNew flag?

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -5029,6 +5029,60 @@ BlackboardArtifact newBlackboardArtifact(int artifact_type_id, long obj_id, Stri
...
+	 * @param data_source_obj_id Data source object id.
+	 * @param account            Os Account associated with the artifact. May be
+	 *                           null.
+	 * @param connection         Database connection to use.
+	 *
+	 * @return DataArtifact
+	 *
+	 * @throws TskCoreException
+	 */
+	DataArtifact newDataArtifact(BlackboardArtifact.Type artifactType, long obj_id, long data_source_obj_id, OsAccount account, CaseDbConnection connection) throws TskCoreException {
```
rcordovano, line 5049: @bcarrier, @jayaramcs and @raman-bt...

We have a story on the books about moving artifact related methods from SleuthkitCase (deprecating them, actually, and culling the less useful ones from the refactoring) to Blackboard. Do we want to invest a some time right now in not having a SleuthkitCase.newDataArtifact() method to deprecate in the near future?

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -697,6 +697,9 @@ private boolean attributesMatch(Collection<BlackboardAttribute> fileAttributesLi
 	 * @throws TskCoreException exception thrown if a critical error occurs
 	 *                          within tsk core
 	 */
+	// TODO: this method needs to be removed.  
+	// The caller should instead use the newDataArtifact() api.
+	//
```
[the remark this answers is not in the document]
raman-bt: @jayaramcs had added this method and using it at the moment.  Brian's intent was to put a Todo as reminder  and we could replace it once things have stabilized.  

I think deprecating it now certainly makes sense. 

If all of Jay's changes are already in and this can be replaced right now, I don't mind doing it.

bindings/java/src/org/sleuthkit/datamodel/DataArtifact.java
raman-bt: The isNew flag is not new (no pun intended)   I'd just missed it when I first created the DataArtifact class. 

I think the intent of the flag is to indicate to client that "this artifact was just created and hence has all its attributed loaded in and don't need to go fetch them from the database"

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -698,6 +698,13 @@ public AnalysisResultAdded newAnalysisResult(BlackboardArtifact.Type artifactTyp
 			throw new TskCoreException("Error adding analysis result.", ex);
 		}
 	}
+	
+	@Override
+	public DataArtifact newDataArtifact(BlackboardArtifact.Type artifactType, Collection<BlackboardAttribute> attributesList, OsAccount osAccount) throws TskCoreException {
+
+		return sleuthkitCase.getBlackboard().newDataArtifact(artifactType, this.getObjectID(), this.getDataSource().getId(), attributesList, osAccount);
```
bcarrier, line 705: The current behavior is that we dont' make BlackboardArtifacts as children of BlackboardArtifacts.  I'd propose we'd have similar behavior for DataArtifacts, which means this should throw an exception.

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
bcarrier: Yes, this should go into Blackboard.java

bindings/java/src/org/sleuthkit/datamodel/Content.java
bcarrier: I think Content already has several of these kinds of methods and will need a reference to SleuthkitCase already. I don't see how Content could go to a pure DTO at this point.

bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -696,7 +697,10 @@ private boolean attributesMatch(Collection<BlackboardAttribute> fileAttributesLi
 	 *
 	 * @throws TskCoreException exception thrown if a critical error occurs
 	 *                          within tsk core
+	 * 
+	 * @deprecated Use the newDataArtifact() API to create an artifact with attributes.
 	 */
+	@Deprecated
```
rcordovano, line 703: Can't we replace the body of this method with a call to the newDataArtifact() API?

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
raman-bt: Addressed.

PR #2171, opened 2021-02-16
Added an API to create account instances.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -43,6 +45,7 @@
 
 	private final SleuthkitCase db;
 
+	private final Map<OsAccountInstanceCacheKey, Long> osAccountInstanceCache = new HashMap<>();
```
rcordovano, line 48: See discussion in Slack.

rcordovano (changes requested):

APriestman (approved):

PR #2175, opened 2021-02-17
Add Port to Postgres Connection

dannysmyda:
I was hitting strange connection issues running a Postgres server on port 49213. I tracked down my issues to this class. When the port is omitted in the URL, 5432 is assumed.


dannysmyda:
I also checked the rest of the datamodel looking for uses of `jdbc:postgresql://`. All other occurrences correctly pass the host and port.

APriestman (approved):

PR #2179, opened 2021-02-18
7275 - Updated RA code to create and updated OsAccounts

kellykelly3:


kellykelly3:
@APriestman When you look at the code, the formatting of the method headers in OsAccountAttribute in the github diffs is different when what I see when I look in netbeans at the file.  Let me know if they look OK in netbeans for you.

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
```diff
@@ -512,6 +513,15 @@ public OsAccountStatus getOsAccountStatus() {
 		return Collections.unmodifiableList(osAccountAttributes);
 	}
 	
+	/**
```
raman-bt, line 516: Is this what's been agreed upon with @bcarrier, @rcordovano  & @jayaramcs ?

I've added and deleted this method multiple times now, as we go back & forth between whether we want OsAccount/OsAccountRealm etc. to be POJOs or be more OO-ey classes with methods. 

If we have decided to add the update() method here then perhaps we should also add one to OsAccountRealm to be consistent.

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
```diff
@@ -399,10 +399,11 @@ void resetDirty() {
 	
 	
 	/**
-	 * Adds account attributes to the account. Attributes can be at
+	 * Adds account attributes to the account.Attributes can be at
```
raman-bt, line 402: I think the period is helpful here.

raman-bt (changes requested):

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
kellykelly3: I am not sure what your comment is refering too, but ss mentioned in a comment to @APriestman the header formatting you see on github is not matching the file.

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
kellykelly3: This is what @bcarrier specified in the story.

bindings/java/src/org/sleuthkit/datamodel/OsAccountAttribute.java
```diff
@@ -35,124 +35,118 @@
...
-	 *                       found.
+	 * @param attributeType	Attribute type.
+	 * @param valueInt	    Int value.
+	 * @param osAccount		Account which the attribute pertains to.
+	 * @param host			Host on which the attribute applies to. Pass Null if
+	 *                      it applies across hosts.
+	 * @param sourceObj		Source where the attribute was found.
 	 */
-	public OsAccountAttribute(BlackboardAttribute.Type attributeType, int valueInt, long osAccountObjId, Long hostId, Long sourceObjId) {
+	public OsAccountAttribute(BlackboardAttribute.Type attributeType, int valueInt, OsAccount osAccount, Host host, Content sourceObj) {
```
raman-bt, line 45: In CT Standup today we discussed that 'sourceObj' for an account to be optional  - i.e caller could pass null for it. 

Do you think you can address it as part of this change? Otherwise I can address is separately after your PR is merged.

kellykelly3: I can take care of it.

raman-bt: Thank you.  

In addition to the constructor you  changed I think this also needs:
 - change return value of getSourceObjId() from Long to Optional.
 - handle the null value when inserting in the database (OsAccountManager.addOsAccountAttributes(): line 678)

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
bcarrier: That may have been from a previous design when we had this method on there or I was just being lazy and imprecise when I meant OsAccountManager.update().  Lets remove this method from here.

kellykelly3: @bcarrier It seems a little inconsistent that the user should call OsAccount.addAttributes to update the account list of attributes, but OsManager to updated the reset of the object.

There is an odd inconsistent behavior here.

raman-bt (approved):

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -673,9 +669,13 @@ void addOsAccountAttributes(OsAccount account, Set<OsAccountAttribute> accountAt
 				if (accountAttribute.getHostId().isPresent()) {
 					preparedStatement.setLong(2, accountAttribute.getHostId().get());
 				} else {
-					preparedStatement.setNull(2, java.sql.Types.BIGINT);
+					preparedStatement.setNull(2, java.sql.Types.NULL);
+				}
+				if(accountAttribute.getSourceObjId().isPresent()) {
+					preparedStatement.setLong(3, accountAttribute.getSourceObjId().get());
+				}else {
+					preparedStatement.setNull(4, java.sql.Types.NULL);
```
jayaramcs, line 677: I think there is a typo here. setNull should be on position 3 instead of 4.

raman-bt (approved):

PR #2183, opened 2021-02-22
7291 update host

gdicristofaro:


bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -133,15 +133,116 @@ Host createHost(String name, CaseDbTransaction trans) throws TskCoreException {
...
+		if (name == null) {
+			throw new IllegalArgumentException("Name provided must be non-null");
+		}
+
+		String queryString = "SELECT h.id FROM data_source_info d INNER JOIN tsk_hosts h ON d.host_id = h.id WHERE LOWER(h.name)=LOWER(?)";
+		String deleteString = "DELETE FROM tsk_hosts WHERE LOWER(name) = LOWER(?)";
+
+		CaseDbTransaction trans = this.db.beginTransaction();
+		try {
+			// check if host has any child data sources.  if so, don't delete and throw exception.
```
APriestman, line 207: Four other tables have foreign keys referencing tsk_hosts:
tsk_os_account_realms
tsk_os_account_attributes
tsk_os_account_instances
tsk_host_address_map
It's unlikely but possible that these could exist for a host that is no longer associated with a data source (for example, if new rows were created during/right after a merge).

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
```diff
@@ -133,15 +133,116 @@ Host createHost(String name, CaseDbTransaction trans) throws TskCoreException {
...
+	 * @throws IllegalArgumentException
+	 */
+	public Host updateHost(Host newHost) throws TskCoreException, IllegalArgumentException {
+		if (newHost == null) {
+			throw new IllegalArgumentException("No host argument provided.");
+		} else if (newHost.getName() == null) {
+			throw new IllegalArgumentException(String.format("Host with id %d has no name", newHost.getId()));
+		}
+
+		CaseDbConnection connection = this.db.getConnection();
```
APriestman, line 163: Need to close this or move to the try()

gdicristofaro: Ok.  Should be in a try-with-resources now.

bindings/java/src/org/sleuthkit/datamodel/HostManager.java
gdicristofaro: Ok.  I'm checking if the host to be deleted is referenced in any of these tables, and if so, I don't delete the host.

APriestman (approved):

PR #2187, opened 2021-02-23
4140: OsAccount Attributes are not loaded from db

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -727,6 +727,57 @@ void addOsAccountAttributes(OsAccount account, List<OsAccountAttribute> accountA
...
+		String queryString = "SELECT attributes.os_account_obj_id as os_account_obj_id, attributes.host_id as host_id, attributes.source_obj_id as source_obj_id, "
+				+ " attributes.attribute_type_id as attribute_type_id,  attributes.value_type as value_type, attributes.value_byte as value_byte, "
+				+ " attributes.value_text as value_text, attributes.value_int32 as value_int32, attributes.value_int64 as value_int64, attributes.value_double as value_double, "
+				+ " hosts.id, hosts.name as host_name, hosts.status as host_status "
+				+ " FROM tsk_os_account_attributes as attributes"
+				+ "		LEFT JOIN tsk_hosts as hosts "
+				+ " ON attributes.host_id = hosts.id "
+				+ " WHERE os_account_obj_id = " + account.getId();
+
+		try (CaseDbConnection connection = this.db.getConnection();
```
APriestman, line 750: Get the single user read lock before the connection

APriestman (changes requested):

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
raman-bt: Added a read lock.

APriestman (approved):

PR #2188, opened 2021-02-23
Search for OsAccountRealm by name if SID not found.

APriestman:


bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -195,15 +195,26 @@ public OsAccountRealm createWindowsRealm(String accountSid, String realmName, Ho
...
-			return this.getRealmByAddr(realmAddr, referringHost, connection);
+			Optional<OsAccountRealm> realm = getRealmByAddr(realmAddr, referringHost, connection);
+			if (realm.isPresent()) {
+				return realm;
+			}
 		}
 
-		// No realm addr, Search  by name	
-		return this.getRealmByName(realmName, referringHost, connection);
+		// No realm addr so search by name	.
```
raman-bt, line 208: is that a tab before the period?

raman-bt (approved):

kellykelly3 (approved):

bcarrier (approved):

PR #2190, opened 2021-02-24
4150:  Added SQLite read locks to get methods in Host/Realms/Accounts etc.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -92,6 +92,7 @@
 			queryString += " AND address_type = " + type.getId();
 		}
 
+		db.acquireSingleUserCaseReadLock();
```
APriestman, line 95: We shouldn't get the lock on an already open connection - the assumption is that we will get it when we create the connection.  Also beginTransaction gets the write lock, so we don't need to do any locking with a transaction.

raman-bt: I realized belatedly that out convention is to get the lock when we open the connection.  

I think it makes sense to be consistent with it. I'll change it, and ensure its consistent elsewhere.

APriestman: It looks like in SleuthkitCase we get the lock we need right before the actual query even when given a connection, so I think your changes make sense. Just take out the unnecessary locks when we have a transaction.

APriestman (approved):

PR #2194, opened 2021-02-25
4160: Log a warning if a thread requests a SQLite connection while it…

raman-bt:
… already holds an uncommitted transaction.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
@@ -12335,6 +12335,10 @@ public class SleuthkitCase {
 
 		@Override
 		public CaseDbConnection getPooledConnection() throws SQLException {
+			// If the requesting thread already has an open transaction, the new connection may get SQLITE_BUSY errors. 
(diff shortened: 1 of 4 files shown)


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -12932,8 +12954,11 @@ public void rollback() throws TskCoreException {
 		 *
 		 */
 		void close() {
-			this.connection.close();
-			sleuthkitCase.releaseSingleUserCaseWriteLock();
+			synchronized(threadsWithOpenTransactionLock) {
+				this.connection.close();
+				sleuthkitCase.releaseSingleUserCaseWriteLock();
+				threadsWithOpenTransaction.remove(Thread.currentThread().getId());
+			}
```
rcordovano, line 12961: Can the scope of this synchronized block be reduced to just the call into the Map protected by the lock Object?

rcordovano (changes requested):

raman-bt:
@rcordovano : I addressed the change you requested.  Could you please take another look and merge it if it looks ok?

rcordovano (approved):

APriestman (approved):

PR #2201, opened 2021-03-01
4162: add version based optimistic row locking to tsk_aggregate_score…

raman-bt:
… table.


bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
```diff
@@ -174,22 +182,38 @@ private void setAggregateScore(long objId, long dataSourceObjectId, Score score,
...
+
+					// register score change in the transaction.
+					transaction.registerScoreChange(new ScoreChange(objId, dataSourceObjectId, currentScore, resultScore));
+					return resultScore;
+				} else {
+					// return the current score
+					return currentScore;
+				}
+			} catch (SQLException ex) {
+				LOGGER.log(Level.WARNING, String.format("Error trying to update aggregate score for objId = %d, data source object id = %d.  Will try again. ", objId, dataSourceObjectId), ex);
```
APriestman, line 212: You probably need to make a checkpoint and rollback the transaction here. I ran into errors with postgresql when I tried to continue using a transaction after an error. See https://www.progress.com/blogs/solution-to-current-transaction-is-aborted-error-for-postgresql-compatibility-with-odbc-jdbc

raman-bt: Hmm.  The transaction is given to us by the calling client.  I don't think we can/should just roll it back  - it probably creates more trouble for the client than what we are trying to address.

@bcarrier, your thoughts?

raman-bt: Ahh, I see what you mean - a savepoint for the transaction.  Yea - let me look into it.

APriestman: I did it in createHost() in HostManager.java

bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
```diff
@@ -114,47 +118,52 @@ private Score getAggregateScore(long objId, CaseDbConnection connection) throws
...
-	private void setAggregateScore(long objId, long dataSourceObjectId, Score score, CaseDbConnection connection) throws TskCoreException {
+	private void setAggregateScore(long objId, long dataSourceObjectId, Score score, int version, CaseDbConnection connection) throws TskCoreException, SQLException {
 
+		// insert or update score
 		String query = String.format("INSERT INTO tsk_aggregate_score (obj_id, data_source_obj_id, significance , confidence) VALUES (%d, %d, %d, %d)"
-				+ " ON CONFLICT (obj_id) DO UPDATE SET significance = %d, confidence = %d",
-				objId, dataSourceObjectId, score.getSignificance().getId(), score.getConfidence().getId(), score.getSignificance().getId(), score.getConfidence().getId() );
+				+ " ON CONFLICT (obj_id) DO UPDATE SET version = %d, significance = %d, confidence = %d WHERE version = %d",
+				objId, dataSourceObjectId, score.getSignificance().getId(), score.getConfidence().getId(), 
+				version + 1, score.getSignificance().getId(), score.getConfidence().getId(), version );
```
APriestman, line 155: In postgresql I got this error:
org.postgresql.util.PSQLException: ERROR: column reference "version" is ambiguous

bcarrier:
Instead of opportunistic locking, we are going to lock the table. We will keep this branch around though.

bcarrier:
I was going to merge this into a branch for safe keeping, but it has merge conflicts that need to be resolved.

raman-bt:
Its on my todo list to fix the conflicts.  I will do it right after finishing up account update.

PR #2207, opened 2021-03-02
4169: Allow null data source for OsAccounts and Host Addresses

raman-bt:


kellykelly3:
@raman-bt I would recommend doing a usage search on getDataSourceObjectID() I think there is some potential for some NPEs.  For example BlackboardAttribute.getDisplayString. 

Also AccountFileInstance.getDataSourceObjectID needs to be modified to return a Long

kellykelly3 (approved):

PR #2214, opened 2021-03-05
HostAddress api updates 

jayaramcs:
## New API 
`Optional<long> existsHostAddress(HostAddress.HostAddressType type, String address) throws TskCoreException `

Returns ObjId of HostAddress if it exists. 

`boolean existsHostNameAndIpMapping(long addr) throws TskCoreException `

Returns true if addr is used as either IP or host name

## Rename API 
`addHostNameToIpMapping` to `addHostNameAndIpMapping` because 'To' seems to imply too much directionality  

## Performance 
Both exists methods could have a Guava cache that is populated by:
- Calls to exists
- Calls to 'add' or 'get' methods

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
@@ -360,7 +360,7 @@ public class HostAddressManager {
 	public void addHostNameAndIpMapping(HostAddress dnsNameAddress, HostAddress ipAddress, Long time, Content source, final SleuthkitCase.CaseDbTransaction caseDbTransaction) throws TskCoreException {
 
 		if (Objects.isNull(caseDbTransaction)) {
-			throw new IllegalArgumentException("null caseDbTransaction passed to addAttributes");
+			throw new IllegalArgumentException("null caseDbTransaction passed to addHostNameAndIpMapping");
 		}
 		try {
 			addHostNameAndIpMapping(dnsNameAddress, ipAddress, time, source, caseDbTransaction.getConnection());
@@ -470,7 +470,7 @@ public class HostAddressManager {
 	 */
 	public Optional<Long> existsHostAddress(HostAddress.HostAddressType type, String address) throws TskCoreException {
 		
-		long id = recenHostAddresstCache.getIfPresent(type.getId()+"#"+address.toLowerCase());
+		Long id = recenHostAddresstCache.getIfPresent(type.getId()+"#"+address.toLowerCase());
 		if(Objects.nonNull(id)){
 			return Optional.of(id);
 		}
--- a/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
@@ -389,7 +389,7 @@ public class HostAddressManager {
 			throw new IllegalArgumentException("An IPv4/IPv6 address is expected.");
 		}
 		if (Objects.isNull(connection)) {
-			throw new IllegalArgumentException("null caseDbTransaction passed to addHostNameToIpMapping");
+			throw new IllegalArgumentException("null connection passed to addHostNameAndIpMapping");
 		}
 
 		String insertSQL = db.getInsertOrIgnoreSQL(" INTO tsk_host_address_dns_ip_map(dns_address_id, ip_address_id, source_obj_id, time) "
--- a/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
@@ -50,17 +50,17 @@ public class HostAddressManager {
 	/**
 	 * An HostAddress Object Id entry is maintained in this cache when a
 	 * hostaddress and ip mapping is added. This is here to improve the
-	 * performance of {@link #existsHostNameAndIpMapping(long) } check.
+	 * performance of {@link #hostNameAndIpMappingExists(long) } check.
 	 */
 	private final Cache<Long, Byte> recentHostNameAndIpMappingCache = CacheBuilder.newBuilder().maximumSize(200000).build();
 
 	/**
 	 * Recently added or accessed Host Address Object Ids are cached. This is
 	 * here to improve performance of the
-	 * {@link #existsHostAddress(org.sleuthkit.datamodel.HostAddress.HostAddressType, java.lang.String)}
+	 * {@link #hostAddressExists(org.sleuthkit.datamodel.HostAddress.HostAddressType, java.lang.String)}
 	 * check.
 	 */
-	private final Cache<String, Long> recenHostAddresstCache = CacheBuilder.newBuilder().maximumSize(200000).build();
+	private final Cache<String, Long> recentHostAddresstCache = CacheBuilder.newBuilder().maximumSize(200000).build();
 
 	
 	/**
@@ -193,7 +193,7 @@ public class HostAddressManager {
 			preparedStatement.setString(3, address.toLowerCase());
 
 			connection.executeUpdate(preparedStatement);
-			recenHostAddresstCache.put(addressType.getId()+"#"+address.toLowerCase(), objId);
+			recentHostAddresstCache.put(addressType.getId()+"#"+address.toLowerCase(), objId);
 			return new HostAddress(db, objId, addressType, address);
 		} catch (SQLException ex) {
 			throw new TskCoreException(String.format("Error adding host address of type = %s, with address = %s", type.getName(), address), ex);
@@ -312,7 +312,7 @@ public class HostAddressManager {
 				long objId = rs.getLong("id");
 				int type = rs.getInt("address_type");
 				String address =  rs.getString("address");
-				recenHostAddresstCache.put(type+"#"+address, objId);
+				recentHostAddresstCache.put(type+"#"+address, objId);
 				return new HostAddress(db, objId, HostAddress.HostAddressType.fromID(type),address);
 			}
 		} catch (SQLException ex) {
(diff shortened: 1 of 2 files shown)


bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -308,38 +332,177 @@ private HostAddress getHostAddress(long id, CaseDbConnection connection) throws
...
+	 * @param type
+	 * @param address
+	 *
+	 * @return
+	 *
+	 * @throws TskCoreException
+	 */
+	public Optional<Long> existsHostAddress(HostAddress.HostAddressType type, String address) throws TskCoreException {
+		
+		long id = recenHostAddresstCache.getIfPresent(type.getId()+"#"+address.toLowerCase());
```
raman-bt, line 473: Does id need to be type Long here?

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -308,38 +332,177 @@ private HostAddress getHostAddress(long id, CaseDbConnection connection) throws
...
+	 * @param caseDbTransaction The transaction in the scope of which the
+	 *                          operation is to be performed, managed by the
+	 *                          caller. Null is not permitted.
+	 *
+	 * @throws TskCoreException
+	 */
+	public void addHostNameAndIpMapping(HostAddress dnsNameAddress, HostAddress ipAddress, Long time, Content source, final SleuthkitCase.CaseDbTransaction caseDbTransaction) throws TskCoreException {
+
+		if (Objects.isNull(caseDbTransaction)) {
+			throw new IllegalArgumentException("null caseDbTransaction passed to addAttributes");
```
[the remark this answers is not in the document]
jayaramcs: changed.

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
jayaramcs: Yes . changed.

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -43,7 +46,23 @@
...
+	 * An HostAddress Object Id entry is maintained in this cache when a
+	 * hostaddress and ip mapping is added. This is here to improve the
+	 * performance of {@link #existsHostNameAndIpMapping(long) } check.
+	 */
+	private final Cache<Long, Byte> recentHostNameAndIpMappingCache = CacheBuilder.newBuilder().maximumSize(200000).build();
 
+	/**
+	 * Recently added or accessed Host Address Object Ids are cached. This is
+	 * here to improve performance of the
+	 * {@link #existsHostAddress(org.sleuthkit.datamodel.HostAddress.HostAddressType, java.lang.String)}
```
rcordovano, line 60: @APriestman, see previous question.

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -308,38 +332,177 @@ private HostAddress getHostAddress(long id, CaseDbConnection connection) throws
...
-				preparedStatement.setNull(4, java.sql.Types.BIGINT);
+	/**
+	 * Returns true if addressObjectId is used as either IP or host name
+	 * <br>
+	 * <b>Note:</b> This api call uses a database connection. Do not invoke within a transaction.
+	 * 
+	 * @param address
+	 * @return 
+	 */
+	public boolean existsHostNameAndIpMapping(long addressObjectId) throws TskCoreException {
```
rcordovano, line 422: I think it would be more natural if this method were named hostNameAndIpMappingExists().

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -308,38 +332,177 @@ private HostAddress getHostAddress(long id, CaseDbConnection connection) throws
...
+	 * within a transaction.
+	 *
+	 * @param type
+	 * @param address
+	 *
+	 * @return
+	 *
+	 * @throws TskCoreException
+	 */
+	public Optional<Long> existsHostAddress(HostAddress.HostAddressType type, String address) throws TskCoreException {
```
rcordovano, line 471: I think it would be more natural if this method were named hostAddressExists().

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -352,7 +515,7 @@ public void addHostNameToIpMapping(HostAddress dnsNameAddress, HostAddress ipAdd
 	 *
 	 * @throws TskCoreException
 	 */
-	List<HostAddress> getIp(String hostname) throws TskCoreException {
+	public List<HostAddress> getIp(String hostname) throws TskCoreException {
```
rcordovano, line 518: Perhaps getIpAddress() to be more exact?

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -43,7 +46,23 @@
 	private static final Logger LOGGER = Logger.getLogger(HostAddressManager.class.getName());
 
 	private final SleuthkitCase db;
+	
+	/**
+	 * An HostAddress Object Id entry is maintained in this cache when a
+	 * hostaddress and ip mapping is added. This is here to improve the
+	 * performance of {@link #existsHostNameAndIpMapping(long) } check.
```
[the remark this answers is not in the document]
APriestman: I don't see any errors for this line when I run doxygen

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -43,7 +46,23 @@
...
+	 */
+	private final Cache<Long, Byte> recentHostNameAndIpMappingCache = CacheBuilder.newBuilder().maximumSize(200000).build();
 
+	/**
+	 * Recently added or accessed Host Address Object Ids are cached. This is
+	 * here to improve performance of the
+	 * {@link #existsHostAddress(org.sleuthkit.datamodel.HostAddress.HostAddressType, java.lang.String)}
+	 * check.
+	 */
+	private final Cache<String, Long> recenHostAddresstCache = CacheBuilder.newBuilder().maximumSize(200000).build();
```
APriestman, line 63: The 't' seems misplaced (should be recentHostAddressCache, right?)

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -308,38 +332,177 @@ private HostAddress getHostAddress(long id, CaseDbConnection connection) throws
...
-			if (time != null) {
-				preparedStatement.setLong(4, time);
-			} else {
-				preparedStatement.setNull(4, java.sql.Types.BIGINT);
+	/**
+	 * Returns true if addressObjectId is used as either IP or host name
+	 * <br>
+	 * <b>Note:</b> This api call uses a database connection. Do not invoke within a transaction.
+	 * 
+	 * @param address
```
APriestman, line 419: Incorrect parameter name

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
APriestman: Seems to be fine

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -308,38 +332,177 @@ private HostAddress getHostAddress(long id, CaseDbConnection connection) throws
...
 			throw new IllegalArgumentException("An IPv4/IPv6 address is expected.");
 		}
+		if (Objects.isNull(connection)) {
+			throw new IllegalArgumentException("null connection passed to addHostNameAndIpMapping");
+		}
 
 		String insertSQL = db.getInsertOrIgnoreSQL(" INTO tsk_host_address_dns_ip_map(dns_address_id, ip_address_id, source_obj_id, time) "
 				+ " VALUES(?, ?, ?, ?) ");
 
-		db.acquireSingleUserCaseWriteLock();
```
APriestman, line 323: I think we should either pass in the transaction (so we know we already have the write lock) or get the write lock (possibly for the second time)

jayaramcs: @APriestman There are two public variants of this method, one that takes a transaction and another that provides the connection.  Are you suggesting that I reacquire the lock write lock regardless of the manner in which the call gets here?

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
jayaramcs: renamed

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
jayaramcs: renamed

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
jayaramcs: renamed

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
jayaramcs: tough audience. Fixed

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
jayaramcs: fixed

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
APriestman: So I think maybe addHostNameAndIpMapping() should not get the lock - I'm under the impression that for these new classes we've gone with acquiring the lock only around the actual SQL call. It'd only be redundant for the transaction case. If you're against that at least make sure there's a prominent comment about having the write lock before calling that method.

jayaramcs: @APriestman I have moved the lock closer to the statements.

APriestman (approved):

PR #2216, opened 2021-03-05
7354 OsAccountInstance handling

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
```diff
@@ -553,6 +488,21 @@ public OsAccountDbStatus getOsAccountDbStatus() {
 		return Collections.unmodifiableList(osAccountAttributes);
 	}
 	
+	/**
+	 * Get get account instances.
```
APriestman, line 492: typo

bindings/java/src/org/sleuthkit/datamodel/OsAccountInstance.java
```diff
@@ -0,0 +1,222 @@
...
+			}
+		}
+
+		return dataSource;
+	}
+
+	/**
+	 * Returns the type for this OsAccount instance.
+	 *
+	 * @return
```
APriestman, line 102: Missing return value (also missing on the next method)

APriestman (approved):

PR #2220, opened 2021-03-08
4189: update account instance types.

raman-bt:


kellykelly3:
@raman-bt & @rcordovano 
This is going to cause interesting issues with my PR.  I moved this enum out of OsAccount into a new OsAccountInstance class.  I am not sure what the best way to handle this is?  Should I approve this RP and let it go in then fix my PR? https://github.com/sleuthkit/sleuthkit/pull/2216

kellykelly3 (approved):

PR #2224, opened 2021-03-10
7316 fix retrieval of osaccountid for files

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -9342,6 +9342,9 @@ void deleteDataSource(long dataSourceObjectId) throws TskCoreException {
 					}
 					
 					Long osAccountObjId = rs.getLong("os_account_obj_id");
+					if(rs.wasNull()) {
```
raman-bt, line 9345: Not a big deal but for consistency sake, I think a space between if and ( is our norm

(fix all instances)

raman-bt (changes requested):

raman-bt (approved):

PR #2233, opened 2021-03-11
4200: address a deadlock - acquiring a new connection while already h…

raman-bt:
…olding a transaction.


raman-bt:
Closing this PR.  Will create new separate ones later.

PR #2237, opened 2021-03-12
Add Locking to CaseDbAccessManager

dannysmyda:


bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -115,6 +115,7 @@ public boolean columnExists(String tableName, String columnName, CaseDbTransacti
         Statement statement = null;
 		ResultSet resultSet = null;
         try {
+			tskDB.acquireSingleUserCaseReadLock();
```
APriestman, line 118: This shouldn't be necessary since the transaction gets the write lock. If you're seeing issue with transactions, it may be that we need to move acquiring the write lock in the CaseDbTransaction constructor to the beginning, and maybe not call getConnection() before calling the constructor and getting the lock.

dannysmyda: I actually didn't see issues with this method in particular, I preemptively added it, but I can strip this PR down to just changes with the methods I was using directly.

dannysmyda:
Actually, I reran my code without this PR and it works just fine. I'm going to close all together.

PR #2240, opened 2021-03-16
4216: Update the TSK Score Enums.

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/Score.java
```diff
@@ -38,42 +39,47 @@
...
 	 */
 	public enum Significance {
 
-		NONE(0, "None"),		//< Item is Good and has no (bad) significance
-		UNKNOWN(10, "Unknown"), //< no analysis has been performed to ascertain significance.
-		LOW(20, "Low"),
-		MEDIUM(30, "Medium"), //< Suspicious.  Could be good or bad. 
-		HIGH(40, "High");	//< Bad & notable
-
+		UNKNOWN(0, bundle.getString("Significance.Unknown.text")),				// no analysis has been performed to ascertain significance.
```
bcarrier, line 77: I don't think we need UNKNOWN anymore.

raman-bt: We define a SCORE_UNKNOWN as new Score(Significance.UNKNOWN, Confidence.NONE)

This is returned  by getAggregateScore() for anything that does *not* have any  analysis  result or real score.

raman-bt:
Closing this PR.  We've decided to rename Confidence to something else as it doesn't really reflect confidence any more.

PR #2246, opened 2021-03-18
4227: stricter validation in OsAccountRealm and OsAccount set() methods.

raman-bt:
Also make isDirty() methods public.


APriestman (approved):

raman-bt:
@rcordovano : could you merge this please.

PR #2263, opened 2021-03-24
OsAccount stores only ID to prevent data from getting stale. Cleaned …

bcarrier:
…up from when realm was optional.  An additional Autopsy PR is needed.


APriestman (approved):

APriestman:
Made temp fix in Autopsy PR https://github.com/sleuthkit/autopsy/pull/6821
Story to do it right in JIRA-7451

PR #2266, opened 2021-03-25
Changed delete events so that they include only ID and not objects with data no longer in DB.

bcarrier:
Will need corresponding Autopsy PR.


APriestman:
Please resolve conflicts

APriestman (approved):

PR #2275, opened 2021-03-29
Lock aggregate_score table so that only one client can update at a time

bcarrier:


APriestman:
It looks like Blackboard.deleteAnalysisResult() calls setAggregateScore() without going through the locking. Is this an issue?

APriestman (approved):

PR #2281, opened 2021-03-30
7454 - Moved the tsk events to TskEvent

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TskEvent.java
```diff
@@ -18,10 +18,262 @@
...
+		 */
+		public List<Long> getOsAcountObjectIds() {
+			return Collections.unmodifiableList(accountObjectIds);
+		}
+	}
+
+	/**
+	 * Base event for all person events
+	 */
+	static class PersonTskEvent {
```
rcordovano, line 212: PersonsTskEvent seems like a more consistent name for this class.

rcordovano (changes requested):

rcordovano (approved):

PR #2283, opened 2021-03-31
4251: Reconcile different OsAccount update philosophies

raman-bt:
This PR does not address updating an account in case of a merge.  Therefore I'm marking it as draft for now.


bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -1024,8 +1032,12 @@ void addOsAccountAttributes(OsAccount account, List<OsAccountAttribute> accountA
 				}
 
 				connection.executeUpdate(preparedStatement);
-
+				
 			}
+
```
jayaramcs, line 1037: Please revert this. My pr takes care of this.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -959,12 +967,12 @@ private void mergeOsAccountObjects(OsAccount sourceAccount, OsAccount destAccoun
...
-	 * @param account	         Account for which the attributes is being added.
-	 * @param accountAttribute List of attributes to add.
+	 * @param account           Account for which the attributes is being added.
+	 * @param accountAttributes List of attributes to add.
 	 *
-	 * @throws TskCoreException,
+	 * @throws TskCoreException
 	 */
-	void addOsAccountAttributes(OsAccount account, List<OsAccountAttribute> accountAttributes) throws TskCoreException {
+	public void addOsAccountAttributes(OsAccount account, List<OsAccountAttribute> accountAttributes) throws TskCoreException {
```
jayaramcs, line 975: Please revert this. My pr takes care of this.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -826,29 +809,58 @@ private String makeOsAccountUpdateQuery(String tableName, OsAccount sourceAccoun
...
+		// update the dest account core 
+		AccountUpdateStatus updateStatus = this.updateOsAccountCore(destAccount, destAddr, destLoginName, trans);
+		
+		if (updateStatus.getUpdateStatusCode() == AccountUpdateStatusEnum.UPDATED && updateStatus.getUpdatedAccount().isPresent()) {
+			mergedDestAccount = updateStatus.getUpdatedAccount().get();
+		}
+		
+		String destFullName = null;
+		Long destCreationTime = null;
 		if (!destAccount.getFullName().isPresent() && sourceAccount.getFullName().isPresent()) {
```
APriestman, line 848: It appears that updateOsAccountProperties will not change the field in destAccount if it exists, so the first half of this check may not longer be necessary. This is likely true of several other conditionals in this method as well.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -1125,101 +1137,322 @@ public void addOsAccountAttributes(OsAccount account, List<OsAccountAttribute> a
 
 		return instanceList;
 	}
+	
+	
+	/**
+	 * Updates the properties of the specified account in the
+	 * database.
+	 *
+	 * A column is updated only if it's current value is null and a non-null
```
APriestman, line 1146: it's -> its

APriestman (approved):

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
raman-bt: We've changed updateOsAccountProperties () to set any non-null value it is given.

APriestman (changes requested):

APriestman (approved):

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -220,51 +220,134 @@ public OsAccountRealm createWindowsRealm(String accountSid, String realmName, Ho
...
+	
 	/**
-	 * Updates the specified realm in the database.
-     * NOTE: This will not merge two realms if the updated information exists
+	 * Get a windows realm by the account SID, or the domain name. The input SID
+	 * is an user/group account SID. The domain SID is extracted from this
+	 * incoming SID.
+	 *
+	 * If a realm is found but is missing either the addr or the realmName, then
+	 * the realm is updated.
```
rcordovano, line 230: I'm not sure a reader will know about the equivalency of the account SID and the address when they read this.

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -220,51 +220,134 @@ public OsAccountRealm createWindowsRealm(String accountSid, String realmName, Ho
...
+		
+		return realmOptional; // return the found realm as is, if any
+	}
+	
+	
+	/**
+	 * Updates the address and/or name of specified realm in the database.
+     *
+	 * Updates the realm address and/or name, if a non blank address/name is
+	 * specified and the current address/name is blank.
```
rcordovano, line 267: Are these two "Updates the..." statements redundant?

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -220,51 +220,134 @@ public OsAccountRealm createWindowsRealm(String accountSid, String realmName, Ho
...
 	}
 		
 	/**
-	 * Updates the specified realm in the database.
-	 * 
-	 * @param realm Realm to update.
+	 * Updates the address and/or name of specified realm in the database.
+	 *
+	 * Updates the realm address and/or name, if a non blank address/name is
+	 * specified and the current address/name is blank.
```
rcordovano, line 296: Same question as above, are these two "Updates the..." sentences redundant?

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -220,51 +220,134 @@ public OsAccountRealm createWindowsRealm(String accountSid, String realmName, Ho
...
+		} else {
+			newRealmName = currRealmName;
+		}
+		
+		// make new signature
+		String newSignature = makeRealmSignature(newRealmAddr, newRealmName, realm.getScopeHost().orElse(null));
+		
+		// if nothing is to be changed, return
+		if ( updateStatusCode == RealmUpdateStatusEnum.NO_CHANGE) {
+			return new RealmUpdateStatus(updateStatusCode, null);
```
rcordovano, line 347: I believe we need to always return an OsAccountRealm here, even if it is the one passed in.

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -775,11 +861,9 @@ void mergeRealms(OsAccountRealm sourceRealm, OsAccountRealm destRealm, CaseDbTra
 		
 		// Update the destination realm if it doesn't have the name or addr set and the source realm does
 		if (!destRealm.getRealmAddr().isPresent() && sourceRealm.getRealmAddr().isPresent()) {
-			destRealm.setRealmAddr(sourceRealm.getRealmAddr().get());
-			updateRealm(destRealm, trans.getConnection());
+			updateRealm(destRealm, sourceRealm.getRealmAddr().get(), null, trans.getConnection());
 		} else if (destRealm.getRealmNames().isEmpty() && !sourceRealm.getRealmNames().isEmpty()) {
-			destRealm.addRealmName(sourceRealm.getRealmNames().get(0));
-			updateRealm(destRealm, trans.getConnection());
+			updateRealm(destRealm, null, sourceRealm.getRealmNames().get(0), trans.getConnection());
```
rcordovano, line 866: Do these lines mean that we are updating the signature twice? If so, can we combine operations by using more conditional clauses?

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -812,4 +896,37 @@ void mergeRealms(OsAccountRealm sourceRealm, OsAccountRealm destRealm, CaseDbTra
...
+	}
+	
+	/**
+	 * Container to encapsulate the status returned by the realm update api, and
+	 * the updated realm.
+	 */
+	public final static class RealmUpdateStatus {
+		
+		private final RealmUpdateStatusEnum updateStatusCode;
+		private final OsAccountRealm updatedRealm; // may be null if there is no update.
```
rcordovano, line 917: I believe it has been agreed that an OsAccountRealm object will always be returned. Also, we should use the name changes: RealmUpdateStatus => OsRealmUpdateResult and RealmUpdateStatusEnum => OsRealmUpdateStatus.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
APriestman: We're actually only updating the signature up above this (makeMergedRealmSignature()). In updateRealm() it will stay the same because the status isn't ACTIVE. The general idea is that we enforce in the SQL update statement that the signature won't change once the status is MERGED instead of going by anything in the realm object to try to avoid timing issues.

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
```diff
@@ -481,7 +395,7 @@ public OsAccountDbStatus getOsAccountDbStatus() {
 	 *
 	 * @throws TskCoreException
 	 */
-	public synchronized List<OsAccountAttribute> getOsAccountAttributes() throws TskCoreException {
+	public synchronized List<OsAccountAttribute> getExtendedOsAccountAttributes() throws TskCoreException {
```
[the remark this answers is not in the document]
raman-bt: I believe @jayaramcs has changes to address thread safety while adding attributes  (this method won't be public any more) - he is waiting this PR to get merged.

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
raman-bt: Changed it to"... missing either the SID or the realmName, ..."

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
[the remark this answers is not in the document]
jayaramcs: All my PRs have been merged in. I do not have any outstanding code for tsk locally either. 
This method was made synchronized based on review comments. https://github.com/sleuthkit/sleuthkit/pull/2265/commits/fbec6ff681cd2040be4b28e11374e001637fe524

PR #2293, opened 2021-04-07
4262 4267 account getters setters

raman-bt:
4262: getOsAccountType() & getOsAccountStatus() should return Optional.
4267: Remove setters for fullname, account type, and creation time.
4274: Make OsAccount thread safe.


APriestman (approved):

bindings/java/src/org/sleuthkit/datamodel/OsAccount.java
```diff
@@ -409,7 +362,7 @@ public OsAccountDbStatus getOsAccountDbStatus() {
 	 *
 	 * @throws TskCoreException
 	 */
-	public List<OsAccountInstance> getOsAccountInstances() throws TskCoreException {
+	public synchronized List<OsAccountInstance> getOsAccountInstances() throws TskCoreException {
```
rcordovano, line 365: It looks like we don't have a way to set OS account instances, either via a setter or the constructor. Perhaps we need to add a synchronized setInstancesInternal() method?

rcordovano (changes requested):

PR #2298, opened 2021-04-07
7480 - Updated timeline events and artifact catalog

kellykelly3:


bindings/java/doxygen/artifact_catalog.dox
```diff
@@ -286,9 +286,7 @@ A contact book entry in an application file or database.
 Programs that have been deleted from the system.
 
 ### REQUIRED ATTRIBUTES
-- TSK_DATETIME_DELETED (Date/Time the program was deleted)
-     or 
-  TSK_DATETIME
+- TSK_DATETIME
```
rcordovano, line 289: Please carry over the dropped parenthetical explanation of the attribute.

bindings/java/doxygen/artifact_catalog.dox
```diff
@@ -459,8 +457,7 @@ Details about an installed program.
 - TSK_PROG_NAME (Name of the installed program)
 
 ### OPTIONAL ATTRIBUTES
-- TSK_DATETIME (A date and time associated with the installed program, e.g., the last modified time, in seconds since 1970-01-01T00:00:00Z)
-- TSK_DATETIME_CREATED (When the program was installed, in seconds since 1970-01-01T00:00:00Z)
+- TSK_DATETIME (When the program was installed, in seconds since 1970-01-01T00:00:00Z)
```
rcordovano, line 460: @bcarrier, should this be optional or required?

bindings/java/src/org/sleuthkit/datamodel/Bundle.properties
```diff
@@ -429,7 +429,7 @@ TimelineEventType.UserDeviceEventEnd.txt=User Device Event End
 TimelineEventType.ServiceAccount.txt=Service Account
 TimelineEventType.WIFINetwork.txt=Wifi Network
 TimelineEventType.WebCache.text=Web Cache
-TimelineEventType.InstalledProgram.txt=Installed Program
+TimelineEventType.InstalledProgram.txt=Program Installed
```
rcordovano, line 432: We have two "Program Installed" strings. That doesn't seem right to me.

bindings/java/src/org/sleuthkit/datamodel/TimelineEventType.java
```diff
@@ -231,10 +231,10 @@ public int compare(TimelineEventType o1, TimelineEventType o2) {
 					METADATA_LAST_PRINTED, METADATA_LAST_SAVED, METADATA_CREATED, PROGRAM_EXECUTION,
 					RECENT_DOCUMENTS, REGISTRY, BACKUP_EVENT_START, BACKUP_EVENT, BACKUP_EVENT_END,
 					BLUETOOTH_PAIRING, CALENDAR_ENTRY_START, CALENDAR_ENTRY_END,
-					DELETE_PROGRAM, DELETE_PROGRAM_DELETED,
+					DELETE_PROGRAM,
```
rcordovano, line 234: Could we make the names of these constants consistent, e.g., PROGRAM_DELETED instead of DELETE_PROGRAM?

rcordovano (changes requested):

bindings/java/doxygen/artifact_catalog.dox
bcarrier: Optional. We may not have it.

gdicristofaro (approved):

PR #2300, opened 2021-04-09
4283: Make sure OsAccount Signature is accurate

raman-bt:


bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -1421,6 +1421,7 @@ private OsAccountUpdateResult updateOsAccountCore(OsAccount osAccount, String ad
 		OsAccountUpdateStatus updateStatusCode = OsAccountUpdateStatus.NO_CHANGE;
 		OsAccount updatedAccount;
 
+		db.acquireSingleUserCaseWriteLock();
```
APriestman, line 1424: Unnecessary since a transaction is already open

bindings/java/src/org/sleuthkit/datamodel/OsAccountRealmManager.java
```diff
@@ -303,74 +305,108 @@ public OsRealmUpdateResult updateRealm(OsAccountRealm realm, String realmAddr, S
...
+	 * @throws SQLException If there is an error updating the database.
+	 * @throws TskCoreException  If the value type is not handled.
+	 */
+	private <T> void updateRealmColumn(long realmId, String colName, T colValue, CaseDbConnection connection) throws SQLException, TskCoreException {
+
+		String updateSQL = "UPDATE tsk_os_account_realms "
+				+ " SET " + colName + " = ? "
+				+ " WHERE id = ?";
+
+		PreparedStatement preparedStatement = connection.getPreparedStatement(updateSQL, Statement.NO_GENERATED_KEYS);
```
APriestman, line 390: We probably should get the write lock here since we don't have a transaction

APriestman (approved):

PR #2327, opened 2021-04-20
Changes to prevent allocation of inline ext4 data that exceeds maximum and fixed memory leak on error path

joachimmetz:


joachimmetz:
@rcordovano (others) PTAL might be good to include in the 4.11 release

APriestman (approved):

tsk/fs/ext2fs.c
```diff
@@ -654,11 +654,15 @@ ext4_load_attrs_inline(TSK_FS_FILE *fs_file, const uint8_t * ea_buf, size_t ea_b
         }
     }
 
+    if ((fs_meta->size == 0) || (fs_meta->size > EXT2_INLINE_MAX_DATA_LEN)) {
```
APriestman, line 657: I don't believe this is correct. I have sample data where the file size is just above the inline max and it reads the extra data from the extended attribute. I added a print statement here to display these file sizes:
$ ./fls.exe ext4_with_inline_data_various_sizes.raw
d/d 11: lost+found
r/r 12: fileA
r/r 13: fileB
### metaSize: 75, max inline len: 60
r/r 14: fileC
### metaSize: 95, max inline len: 60
r/r 15: fileD
### metaSize: 115, max inline len: 60
r/r 16: fileE
r/r 17: fileF
V/V 1025:       $OrphanFiles

With your PR, if I try to run icat on one of those files I get an error:
$ ./icat.exe ext4_with_inline_data_various_sizes.raw 14
### metaSize: 75, max inline len: 60
Invalid API argument (tsk_fs_attrlist_get: Null list pointer)

With the current develop code, the contents are printed correctly:
$ ./icat.exe ext4_with_inline_data_various_sizes.raw 14
File C 012345678901234567890123456789012345678901234567890123456789 File C

joachimmetz: 1. There are no unit tests that cover the scenario you describe.
2. I don't have access to the test data you describe, so there are no means for me to validate what you claim makes sense from a data format perspective or not
3. line 671 does a max check with the same constant? that either means fs_meta->size is not the size of the resident data ?

APriestman: This is the script I used - modified version of yours. https://pastebin.com/4jYAjgYn

And yes - it appeared in testing that when the file length is just a little bit longer than would fit in the inline data, it would put it in an extended attribute right after the inline data section. That's what's happening around line 673. fs_meta->size is the true file size and can be slightly larger than the max inline length.

joachimmetz: thx I'll have a look at your version of the script and test files to determine what you are trying to tell me here in terms of edge cases.

But then let's change `EXT2_INLINE_MAX_DATA_LEN` to some other upper bound. The problem with the current code is there is no bounds check so the malloc on line 664 allocates what ever `fs_meta->size` is. If `fs_meta->size` is the actual file size this could be very large in case of inline file content with a huge sparse extent following it.

PS. FYI the script is licensed under CC-BY-4.0 License, redistribution on putting on pastebin without attribution violates the license.

APriestman: No, EXT2_INLINE_MAX_DATA_LEN is the length to read from the resident attribute. We shouldn't change that.

Yes I am aware that we could end up with a huge allocated file if we have a sparse extent but a lot of general attribute-handling code would have to be changed to be able to deal with it differently. With resident data the code will always assume the data is all in a buffer to be read.

APriestman: It's an option, but then attempting to read the file will cause errors so it seems like we're just creating a different problem. Are files with huge sparse extents common in ext4?

joachimmetz: this can happen with malformed data, and can be considered a denial-of-service weakness. Since it can exhaust resource. 

So `fs_meta->size` should not be larger than "inode size - offset of resident data", where is "inode size" tracked?

joachimmetz: So maximum inode size supported by mke2fs is 1024 and offset of resident data is 156, so using 868 as upper bound for the resident data size looks reasonable.

joachimmetz: @APriestman made some changes

joachimmetz:
@APriestman @bcarrier @rcordovano what is the status of this review? This is still an issue in 4.11.0

rcordovano:
@joachimmetz, @APriestman passed this PR on to @bcarrier for further review (observe that Brian is listed with Ann as a reviewer) with the following remark; "The one is under discussion from a while back. Basically a question of supporting files with large sparse extents vs. protecting against allocating large amounts of memory due to error."

joachimmetz:
@rcordovano thx for the update, since you're trying to do recovery in libtsk as well, in which you can encounter all kinds of data, I would recommend to take an approach that can handle large sparse extents with limited memory overhead.

Issue #1355, opened 2018-10-26
Memory Leak (118491754)


Google-Autofuzz:
Hello sleuthkit team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
sleuthkit (tested with revision * develop eb24f79d7ce06cb017c73f0e7d2e4abd1eab6e2e).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

Instructions:
`unzip artifacts_118491754.zip`
`docker build --build-arg SANITIZER=address --tag=autofuzz-sleuthkit-118491754 autofuzz_118491754`
`docker run --entrypoint /fuzzing/repro.sh --cap-add=SYS_PTRACE -v $PWD/autofuzz_118491754/poc-840b951250b11e50f6a1a6aa2b7c31fa5c50bddf018e3b1e96e64b6c5f517f80_min:/tmp/poc autofuzz-sleuthkit-118491754 "fls ext" /tmp/poc`
`docker run --cap-add=SYS_PTRACE -v $PWD/autofuzz_118491754/poc-840b951250b11e50f6a1a6aa2b7c31fa5c50bddf018e3b1e96e64b6c5f517f80_min:/tmp/poc -it autofuzz-sleuthkit-118491754`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
fls -- ext

=================================================================
==9==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x4dc158 in __interceptor_malloc (/fuzzing/sleuthkit/tools/fstools/fls+0x4dc158)
    #1 0x5ba15e in tsk_malloc /fuzzing/sleuthkit/tsk/base/mymalloc.c:32:16
    #2 0x521962 in tsk_fs_file_alloc /fuzzing/sleuthkit/tsk/fs/fs_file.c:31:31
    #3 0x5bfa74 in ext2fs_inode_walk /fuzzing/sleuthkit/tsk/fs/ext2fs.c:1007:20
    #4 0x520685 in tsk_fs_dir_find_orphans /fuzzing/sleuthkit/tsk/fs/fs_dir.c:1205:9
[...]

Issue #1189, opened 2018-03-30
Overflow / Invalid Shift (77149877)


Google-Autofuzz:
Hello sleuthkit team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
sleuthkit (tested with revision * develop f0fa8f8a020338de72c3e6d95d324db43cf641ba).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

Instructions:
`unzip artifacts_77149877.zip`
`docker build --build-arg SANITIZER=undefined --tag=autofuzz-sleuthkit-77149877 autofuzz_77149877`
`docker run --entrypoint /fuzzing/repro.sh --cap-add=SYS_PTRACE -v $PWD//tmp/autofuzz-triageBbRnWB/poc-2d3c7fb26b565bb108f8e6fbbe9d3fbbc97ebde3a422c0eeb4430c9a3b7eefcf_min:/tmp/poc autofuzz-sleuthkit-77149877 "fls ext" /tmp/poc`
`docker run --cap-add=SYS_PTRACE -v $PWD//tmp/autofuzz-triageBbRnWB/poc-2d3c7fb26b565bb108f8e6fbbe9d3fbbc97ebde3a422c0eeb4430c9a3b7eefcf_min:/tmp/poc -it autofuzz-sleuthkit-77149877`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
fls -- ext
ext2fs_dent.c:152:21: runtime error: left shift of 241 by 24 places cannot be represented in type 'int'
    #0 0x4c4079 in ext2fs_dent_parse_block /fuzzing/sleuthkit/tsk/fs/ext2fs_dent.c:152:21
    #1 0x4c36f9 in ext2fs_dir_open_meta /fuzzing/sleuthkit/tsk/fs/ext2fs_dent.c:340:13
    #2 0x430305 in tsk_fs_dir_open_meta /fuzzing/sleuthkit/tsk/fs/fs_dir.c:290:14
    #3 0x430cb2 in tsk_fs_dir_walk_lcl /fuzzing/sleuthkit/tsk/fs/fs_dir.c:556:19
[...]

PR #2328, opened 2021-04-20
Fixed memory leaks on error path #1355

joachimmetz:


joachimmetz:
@rcordovano (others) PTAL might be good to include in the 4.11 release

tsk/fs/ext2fs.c
```diff
@@ -1174,7 +1178,11 @@ ext2fs_inode_walk(TSK_FS_INFO * fs, TSK_INUM_T start_inum,
     size =
         ext2fs->inode_size >
         sizeof(ext2fs_inode) ? ext2fs->inode_size : sizeof(ext2fs_inode);
-    if ((dino_buf = (ext2fs_inode *) tsk_malloc(size)) == NULL) {
+
+    dino_buf = (ext2fs_inode *) tsk_malloc(size);
+    if (dino_buf == NULL) {
+        free(fs_file->meta);
```
bcarrier, line 1184: Hi @joachimmetz .  Does calling tsk_fs_file_close() not work here instead of the two calls to free fs_file and fs_file->meta?

joachimmetz: maybe, I find the current naming/calling convention confusing, alloc to "open" the fs_file and "close" to free it? 

if that is the case I would recommend creating tsk_fs_file_open()

I would also recommend using `goto on_error` for a more clear and cleaner error path

joachimmetz: for now I've mainly addressed the leak itself

joachimmetz:
This PR addresses https://github.com/sleuthkit/sleuthkit/issues/1355

APriestman (approved):

joachimmetz:
The leaks can be observed with poc of https://github.com/sleuthkit/sleuthkit/issues/1189

PR #2332, opened 2021-04-21
Changes to prevent OOB and uninitialized memory reads when reading ext extents

joachimmetz:


joachimmetz:
@APriestman (others) PTAL

joachimmetz:
@APriestman made some changes since this caused some tests to fail. It would really be beneficial for this project to extend the unit tests (`make check`) with some test images to test regression as part of the CI testing

joachimmetz:
added more bounds checks

APriestman (approved):

PR #2334, opened 2021-04-21
Fixed memory leaks on error path

joachimmetz:


joachimmetz:
@bcarrier PTAL, IMHO another examples of where a clear and cleaner error path would be beneficial

Issue #1858, opened 2020-03-20
ASSERT (141878693)


Google-Autofuzz:
Hello sleuthkit team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
sleuthkit (tested with revision * develop 7e33d5d270a539eef60f711a327aa3cefd2c6623).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

Instructions:
`unzip artifacts_141878693.zip`
`docker build --build-arg SANITIZER=address --tag=autofuzz-sleuthkit-141878693 autofuzz_141878693`
`docker run --entrypoint /fuzzing/repro.sh --cap-add=SYS_PTRACE -v $PWD/autofuzz_141878693/autofuzz_141878693:/tmp/poc autofuzz-sleuthkit-141878693 "fls fat" /tmp/poc`
`docker run --cap-add=SYS_PTRACE -v $PWD/autofuzz_141878693/autofuzz_141878693:/tmp/poc -it autofuzz-sleuthkit-141878693`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
fls -- fat
fls: exfatfs_meta.c:58: int8_t exfatfs_is_cluster_alloc(FATFS_INFO *, TSK_DADDR_T): Assertion `(a_cluster_addr >= FATFS_FIRST_CLUSTER_ADDR) && (a_cluster_addr <= a_fatfs->lastclust)' failed.
AddressSanitizer:DEADLYSIGNAL
=================================================================
==9==ERROR: AddressSanitizer: ABRT on unknown address 0x000000000009 (pc 0x7f3ea105f7bb bp 0x7f3ea11acee0 sp 0x7fff0830d250 T0)
    #0 0x7f3ea105f7ba in gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x377ba)
    #1 0x7f3ea104a534 in abort (/lib/x86_64-linux-gnu/libc.so.6+0x22534)
    #2 0x7f3ea104a40e in __tls_get_addr (/lib/x86_64-linux-gnu/libc.so.6+0x2240e)
    #3 0x7f3ea1058101 in __assert_fail (/lib/x86_64-linux-gnu/libc.so.6+0x30101)
[...]

PR #2347, opened 2021-04-23
Removed redundant asserts for exfat #1858

joachimmetz:


tsk/fs/exfatfs_meta.c
```diff
@@ -50,12 +50,10 @@ exfatfs_is_cluster_alloc(FATFS_INFO *a_fatfs, TSK_DADDR_T a_cluster_addr)
     uint8_t bitmap_byte;
     ssize_t bytes_read = 0;
 
-    assert(a_fatfs != NULL);
     if (fatfs_ptr_arg_is_null(a_fatfs, "a_fatfs", func_name)) {
         return -1;
     }
 
-    assert((a_cluster_addr >= FATFS_FIRST_CLUSTER_ADDR) && (a_cluster_addr <= a_fatfs->lastclust));
     if ((a_cluster_addr < FATFS_FIRST_CLUSTER_ADDR) || (a_cluster_addr > a_fatfs->lastclust)) {
```
micrictor, line 57: These conditions are slightly different - is there a reason that the assertion here catches if `a_cluster_addr == FATFS_FIRST_CLUSTER_ADDR == a_fatfs->lastclust`? I think this might guard against an OOB read if the cluster address is the address of the last cluster.

joachimmetz: > These conditions are slightly different 

Yes but aren't they each-others inverse logic ? Lower bound `FATFS_FIRST_CLUSTER_ADDR` upper bound `a_fatfs->lastclust`. The assert will raise if the condition is false, the if will enter the block if the condition is true.

micrictor: You're right - brain fart, ignore me!

LGTM

micrictor (approved):

APriestman (approved):

Issue #1859, opened 2020-03-20
Stack-overflow (148232034)


Google-Autofuzz:
Hello sleuthkit team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
sleuthkit (tested with revision * develop 62dd0b0437c664f3d12726174cb63c1b2f776b76).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

Instructions:
`unzip artifacts_148232034.zip`
`docker build --build-arg SANITIZER=address --tag=autofuzz-sleuthkit-148232034 autofuzz_148232034`
`docker run --entrypoint /fuzzing/repro.sh --cap-add=SYS_PTRACE -v $PWD/autofuzz_148232034/autofuzz_148232034:/tmp/poc autofuzz-sleuthkit-148232034 "fls fat" /tmp/poc`
`docker run --cap-add=SYS_PTRACE -v $PWD/autofuzz_148232034/autofuzz_148232034:/tmp/poc -it autofuzz-sleuthkit-148232034`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
fls -- fat
AddressSanitizer:DEADLYSIGNAL
=================================================================
==9==ERROR: AddressSanitizer: stack-overflow on address 0x7fff260e87f8 (pc 0x0000004c86fe bp 0x7fff260e9050 sp 0x7fff260e8800 T0)
    #0 0x4c86fd in __asan_memset (/fuzzing/sleuthkit/tools/fstools/fls+0x4c86fd)
    #1 0x52f2bc in tsk_fs_dir_walk /fuzzing/sleuthkit/tsk/fs/fs_dir.c:795:5
    #2 0x638503 in fatxxfs_dent_parse_buf /fuzzing/sleuthkit/tsk/fs/fatxxfs_dent.c:373:29
    #3 0x62e190 in fatfs_dir_open_meta /fuzzing/sleuthkit/tsk/fs/fatfs_dent.cpp:347:14
    #4 0x52e7b9 in tsk_fs_dir_open_meta /fuzzing/sleuthkit/tsk/fs/fs_dir.c:290:14
[...]

PR #2349, opened 2021-04-26
Added tsk_fs_dir_walk recursion depth to prevent stack-overflow #1859

joachimmetz:


joachimmetz:
For reviewer this is not an elegant solution but means to address a more serious issue namely #1859

joachimmetz:
@bcarrier @APriestman can you please review this is quite a serious weakness and has been left unaddressed since at least Mar 20, 2020

bcarrier:
Thanks! I merged this in, but increased the max depth to 256.  32 seemed too low. I don't think I personally have 32 folder deep structures on my system, but I'd hate for someone to not be able to see into one of those folders if a drive had that many. 

I also renamed the methods for consistency:
- Moved "Internal" to the end
- Renamed existing "_lcl" method to "_internal".
-

bcarrier:
I think my approach confused github. I made changes in a local branch and pushed it up, but this PR isn't linking to b9c38b87b, which was into release-4.11.0. I'm going to manually close this.

joachimmetz:
@bcarrier this issue is not fixed, please reopen

* Since I could not cleanly compare diffs between the develop branch with patch and the release-4.11.0 brach. I checked out and build release-4.11.0 with ASAN and ran `fls -r autofuzz_148232034/autofuzz_148232034` and the issue still persists

```
/tools/fstools/fls -r /tmp/autofuzz_148232034/autofuzz_148232034
AddressSanitizer:DEADLYSIGNAL
=================================================================
==830913==ERROR: AddressSanitizer: stack-overflow on address 0x7ffe0fb42800 (pc 0x7efc628cd5d4 bp 0x7ffe0fb43040 sp 0x7ffe0fb42800 T0)
    #0 0x7efc628cd5d4 in free (/lib64/libasan.so.6+0xae5d4)
    #1 0x7efc62213b78 in tzset_internal /usr/src/debug/glibc-2.33-20.fc34.x86_64/time/tzset.c:401
...
    #247 0x55db4a in fatxxfs_dent_parse_buf sleuthkit/tsk/fs/fatxxfs_dent.c:373
    #248 0x5519b3 in fatfs_dir_open_meta sleuthkit/tsk/fs/fatfs_dent.cpp:347

SUMMARY: AddressSanitizer: stack-overflow (/lib64/libasan.so.6+0xae5d4) in free
==830913==ABORTING
```

* so these changes are currently not merged into the develop branch, which could lead to regression.
* Is there any documentation which branch is for what? and how these are managed (e.g. which patches are merged/cherry picked from where).

bcarrier:
Looking at #1589, it looks like there were 248 levels of recursion when the overflow happened.  So, perhaps my increasing the max levels to 256 was too much. 111ba769f changes it to 128.

bcarrier:
Any changes made to just this brach will get merged in when 4.11.0 goes out, which should be monday.  I re-opened up the release for this stuff.

joachimmetz:
ack, I'll apply the patch and see what the tests tell me

joachimmetz:
The stack-overflow has been addressed with https://github.com/sleuthkit/sleuthkit/commit/111ba769f4d43be45d90811f0d5fed91f81b6be1. Note that there a (possible infinite) loop with the POC

```
tsk_fs_dir_walk_internal: Loop detected with address 181fatfs_getFAT: contents of entry 4 too large - resetting
fatfs_getFAT: contents of entry 4 too large - resetting
fatfs_getFAT: contents of entry 4 too large - resetting
tsk_fs_dir_walk_internal: Loop detected with address 181
```

bcarrier:
I just added 49b72951a because the check was returning the wrong return value.  I need to look into this though because we already some recursion checks in that code (that are also at 128 levels). I'm not sure why they were not triggered. We now have two concepts of recursion detection in there...

bcarrier:
Looking at the original problem some more:
- This is unique to FAT file systems because we have special code in there to resolve the '.' entry to the parent metadata address. If we don't know its address, we start at the root folder to find it. 
- It looks like the original bug is from data that causes the scan and then during the scan, the FAT code thinks it needs to resolve the '.' entry again and it goes on forever.
- Seems like the cleaner solution here is to have the FAT code know that it should just recurse down and not try to further resolve.  If we had that, then all of these API changes to all of the "_open" calls could go back. But, I'm not sure what that does with multiple threads...  
 
This is now much more work than I had hoped for on a Saturday morning.  Given that the bug is fixed and this isn't public API, I'll add some comments to the code and maybe change it later.

joachimmetz:
@bcarrier thx for still taking the time to fix the issue on a Saturday morning. Just realize that this PR had been pending since Apr 26. My suggestion would be to take a critical look at the current development / test process, maybe some changes can be made to not have make these merges just before a release deadline ?

PR #2353, opened 2021-04-27
Fixed HFS Catalog key OOB read

joachimmetz:


tsk/fs/hfs.c
```diff
@@ -838,7 +838,8 @@ hfs_cat_traverse(HFS_INFO * hfs,
                 key = (hfs_btree_key_cat *) & node[rec_off];
                 keylen = 2 + tsk_getu16(hfs->fs_info.endian, key->key_len);
 
-                if (keylen > nodesize - rec_off) {
+                // Want a key of at least 6 bytes
+                if ((keylen < 6) || (keylen > nodesize - rec_off)) {
```
APriestman, line 842: The displayed error won't be accurate if the key was too short.
Please add a short note on where the '6' comes from (I think the first two fields in hfs_btree_key_cat ?)

joachimmetz: Correct the first 6 members of hfs_btree_key_cat are 6 bytes, this is because later the second member is checked without a bounds check, but since keylen is not passed the check needs to be added here.

joachimmetz: I've made some updates

APriestman (approved):

PR #2354, opened 2021-04-27
Fixed HFS B-Tree node OOB read

joachimmetz:


tsk/fs/hfs.c
```diff
@@ -3854,6 +3854,13 @@ hfs_load_extended_attrs(TSK_FS_FILE * fs_file,
         return 0;
     }
 
+    // Want a node of at least 14 bytes
```
APriestman, line 3857: For clarity, could you add that this is the size of the hfs_btree_node struct?

joachimmetz: I've changed it to `sizeof(hfs_btree_node)`

Issue #1401, opened 2018-12-26
Heap Out-Of-Bounds Memory Access (73577591)


Google-Autofuzz:
Hello sleuthkit team,

As part of our fuzzing efforts at Google, we have identified an issue affecting
sleuthkit (tested with revision * develop 493cf4e70d1438bb60708ef51c62eeb4c11da98c).

To reproduce, we are attaching a Dockerfile which compiles the project with
LLVM, taking advantage of the sanitizers that it offers. More information about
how to use the attached Dockerfile can be found here:
https://docs.docker.com/engine/reference/builder/

Instructions:
`unzip artifacts_73577591.zip`
`docker build --build-arg SANITIZER=address --tag=autofuzz-sleuthkit-73577591 autofuzz_73577591`
`docker run --entrypoint /fuzzing/repro.sh --cap-add=SYS_PTRACE -v $PWD/autofuzz_73577591/poc-e20ea35480b1e132c186746566268898f9615e98d7f966af2e3c94112a1cabf4_min:/tmp/poc autofuzz-sleuthkit-73577591 "fls hfs" /tmp/poc`
`docker run --cap-add=SYS_PTRACE -v $PWD/autofuzz_73577591/poc-e20ea35480b1e132c186746566268898f9615e98d7f966af2e3c94112a1cabf4_min:/tmp/poc -it autofuzz-sleuthkit-73577591`

Alternatively, and depending on the bug, you could use gcc, valgrind or other
instrumentation tools to aid in the investigation. The sanitizer error that we
encountered is here:

```
fls -- hfs
=================================================================
==9==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6210000014ff at pc 0x000000528d61 bp 0x7ffe13d72340 sp 0x7ffe13d72338
READ of size 1 at 0x6210000014ff thread T0
    #0 0x528d60 in hfs_cat_traverse /fuzzing/sleuthkit/tsk/fs/hfs.c:944:21
    #1 0x52c058 in hfs_cat_get_record_offset /fuzzing/sleuthkit/tsk/fs/hfs.c:1182:9
    #2 0x52b088 in hfs_cat_file_lookup /fuzzing/sleuthkit/tsk/fs/hfs.c:1604:11
    #3 0x52cd19 in hfs_inode_lookup /fuzzing/sleuthkit/tsk/fs/hfs.c:2593:9
[...]

PR #2356, opened 2021-04-28
Fixed OOB reads in hfs_cat_traverse #1401

joachimmetz:


tsk/fs/hfs.c
```diff
@@ -554,11 +554,20 @@ hfs_ext_find_extent_record_attr(HFS_INFO * hfs, uint32_t cnid,
                 int keylen;
                 TSK_FS_ATTR_RUN *attr_run;
 
+                // Make sure node is large enough
+                if ((rec + 1) * 2 > (int) nodesize - 2) {
```
APriestman, line 558: Can you explain this conditional? It seems like it should be testing that (nodesize - (rec+1)*2) is greater than or equal to zero and that it is less than or equal to (nodesize-2), ensuring that the offset into the node array is not negative and there's space to read two bytes.

joachimmetz: * so `(nodesize - (rec+1)*2)` can cause an int wrap seeing `uint16 - ( (int cast of uint16) + 1 ) * 2 )` hence that the condition is split as `(int cast of uint16) + 1 ) * 2 > (int cast of uint16)`
* and we need to test for `(nodesize - (rec+1)*2) > 2` since we want to read 2 bytes

Does that help explain it ? If not let me know happy to provide more context.

joachimmetz: and on second thought this likely needs to be `(rec + 1) * 2 >= (int) nodesize - 2` since `(rec + 1) * 2` is used as a byte array offset

APriestman: I'm not sure why "and we need to test for (nodesize - (rec+1)*2) > 2 since we want to read 2 bytes" is true - aren't we reading two bytes starting at (nodesize - (rec+1)*2)? Why would that need to be greater than 2?

joachimmetz: A general recommendation would be to clean up the code and put `(rec+1)*2` into a variable with a more descriptive name

APriestman: But that's never even used on its own? We could do "offset = nodesize - (rec + 1) * 2", making line 560 "&node[offset]" and the testing beforehand would be that offset >= 0 and offset <= nodesize - 2. That's not the same as your current conditional though.

APriestman: And we really don't need the second check. rec will never be negative so nodesize - (rec + 1) * 2 is at most nodesize-2, leaving room to read two bytes

joachimmetz: > But that's never even used on its own?

Not sure if I understand your question/concern here. Does that matter? The compiler will optimize the code anyway, and if it make the code more readable, why not?

> We could do "offset = nodesize - (rec + 1) * 2"

that could work if you make sure nodesize does not wrap:

```
(int) offset = (int) nodesize - (rec + 1) * 2
```

> That's not the same as your current conditional though.

The current statement combines the bounds check and the offset calculation

```
int node_offset_from_end = (rec + 1) * 2
int node_offset_from_start = nodesize - node_offset_from_end;

// this will allow for a 1-byte OOB read, e.g. if node_offset_from_start = 9 and node_size = 10
if (! ((node_offset_from_start >= 0) && (node_offset_from_start <= nodesize - 2))) {
  error
}

// Without 1-byte OOB read
if (! ((node_offset_from_start >= 0) && (node_offset_from_start < nodesize - 2))) {
  error
}

// Change of the logic to remove the negation
if ((node_offset_from_start < 0) || (node_offset_from_start >= nodesize - 2)) {
  error
}

// Replacing node_offset_from_start by nodesize - node_offset_from_end;
if (((nodesize - node_offset_from_end) < 0) || ((nodesize - node_offset_from_end) >= nodesize - 2)) {
  error
}

if ((node_offset_from_end > nodesize) || (node_offset_from_end < 2))  {
  error
}

// Replacing node_offset_from_end with (rec + 1) * 2

if (((rec + 1) * 2 > nodesize) || ((rec + 1) * 2 < 2)) {
  error
}

//  (rec + 1) * 2 < 2 will only be true if rec < 0 which will not happen unless a bit flip
// So we are left with, but that does not account for nodesize having 2 bytes

if ((rec + 1) * 2 > nodesize) {
  error
}

```

joachimmetz:
```
9 < ( 3 + 1 ) * 2 [9 < 8 false] but we have only 1 byte left
```

APriestman: Your last line there is exactly what I think the error conditional should be, but in your code you have 
if ((rec + 1) * 2 > (int) nodesize - 2) { error handling}

joachimmetz: hope that helps

joachimmetz: However I would strongly encourage us to write this in such a way that it clear, if that is using additional variables to clarify I'm fine with that

joachimmetz: > Your last line there is exactly what I think the error conditional should be, but in your code you have

this does not account for `9 < ( 3 + 1 ) * 2 [9 < 8 false]`

APriestman: "9 < ( 3 + 1 ) * 2 [9 < 8 false] but we have only 1 byte left, hence the additional - 2"
But in this case the offset we're reading from is  9 - 8 =  1, so there's plenty of room to read two bytes

joachimmetz: ah, good point let me recheck the logic

joachimmetz: my bad, good catch. I considered this an offset from the start in terms of the size correction but not an offset from end. So I agree the `-2` is not needed. Though I just noticed that there another instance of this in `hfs_ext_find_extent_record`. Making some change.

joachimmetz:
@APriestman I made some change PTAL

tsk/fs/hfs.c
APriestman: Looks great now - thanks for adding the additional comments

APriestman (approved):

joachimmetz:
Great, feel free to close #1401 as fixed as well

PR #2359, opened 2021-04-28
7553 data model event changes

rcordovano:
- Make all new data model event classes static inner classes of TskEvent
- Make all new data model event classes capable of batching data model objects
- Make all new data model object deletion events work in terms of IDs, not objects


bindings/java/src/org/sleuthkit/datamodel/AggregateScoresChangedEvent.java
```diff
@@ -1,54 +0,0 @@
-/*
```
gdicristofaro, line 1: Will any of these refactorings or renames affect downstream clients of this code?

PR #2364, opened 2021-05-01
Fixed OOB reads in hfs_cat_traverse

joachimmetz:


tsk/fs/hfs.c
```diff
@@ -483,7 +483,7 @@ hfs_ext_find_extent_record_attr(HFS_INFO * hfs, uint32_t cnid,
                 rec_off =
                     tsk_getu16(fs->endian,
                     &node[nodesize - (rec + 1) * 2]);
-                if (rec_off + sizeof(hfs_btree_key_ext) > nodesize) {
+                if (rec_off >= nodesize - sizeof(hfs_btree_key_ext)) {
```
APriestman, line 486: Can you explain why the > turned into >= ?

joachimmetz: Because line 495 rec_off is used as an offset and `&( node[rec_off] ) + sizeof(hfs_btree_key_ext)` would be OOB when `rec_off == nodesize - sizeof(hfs_btree_key_ext)` (seeing rec_off is count from 0)

joachimmetz: e.g. `4 + 8 == 12` but `node[ 12 ]` would be out of bounds

APriestman: But would we access node[12]? Suppose hfs_btree_key_ext was just the first part: "uint8_t key_len[2]". Then sizeof(hfs_btree_key_ext) would be 2. If nodesize was 12 and rec_off was 10, we'd be fine, wouldn't we? key_len[0] would be node[10] and key_len[1] would be node[11].

joachimmetz: `sizeof(hfs_btree_key_ext)` would be whatever the size of this structure is defined. If the full structure or not is accessed I cannot tell from the original check. Then I recommend making the check more clear what bounds is actually checked. Note that the changes did address an OOB read flagged by a fuzzer.

joachimmetz: Note the challenge of the code is that pointer are casted back and forth so it is very hard to see what struct members are actually used

joachimmetz: This should be the corresponding POC file https://oss-fuzz.com/testcase-detail/5640270227046400

APriestman: I got the file from Brian. It looks like the OOB read is taken care of by the change at line 861. Brian and I both believe that this line should be kept as > not >=.

joachimmetz: as I said it might be good to change the test then to a bound that is more representative for what data is actually accessed or add a comment. What part of  https://github.com/sleuthkit/sleuthkit/blob/9cabe33ee9c6f1098e7866c0d05217d948e0b19c/tsk/fs/tsk_hfs.h#L463 is used and what part not? 

Note that you have edit access on this PR, so don't hesitate to tweak the check.

APriestman (approved):

PR #2367, opened 2021-05-03
7572 Add AFF4 nuget library

APriestman:


kellykelly3:
Closing this PR because of an inability to test.

@APriestman indicated that although the changes probably work it was never tested due to issues installing the aff4 library on Linux.

PR #2375, opened 2021-05-05
Changes to prevent self-copy of orphan sub directory name

joachimmetz:


joachimmetz:
Likely related https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24899&q=sleuthkit&can=2

joachimmetz:
@bcarrier @rcordovano what is the status of this review? This is a security weakness that has been reported since Aug 14, 2020

APriestman (approved):

PR #2379, opened 2021-05-06
Fixed leaks in fatfs_make_data_runs

joachimmetz:


joachimmetz:
Related to https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24892&q=sleuthkit&can=2

APriestman (approved):

PR #2380, opened 2021-05-06
Fixed leak in dos_load_ext_table

joachimmetz:


joachimmetz:
Related to https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24894&q=sleuthkit&can=2

APriestman (approved):

PR #2381, opened 2021-05-06
Fixed leaks in iso9660_dir_open_meta

joachimmetz:


joachimmetz:
Related to https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24949&q=sleuthkit&can=2

APriestman (approved):

PR #2382, opened 2021-05-07
Fixed OOB read in tsk_printf due to missing end-of-string character

joachimmetz:


joachimmetz:
Surfaced with POC of https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=25580&q=sleuthkit&can=2

APriestman (approved):

PR #2403, opened 2021-05-18
Adding caching for HostAddressManager

jayaramcs:
- addUsage
- getHostAddress

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
@@ -62,9 +62,10 @@ public class HostAddressManager {
 	/**
 	 * Recently added host address usage is cached. This is intended to improve 
 	 * the performance of {@link #addUsage(org.sleuthkit.datamodel.Content, org.sleuthkit.datamodel.HostAddress) }
-	 * Key: DatasourceId # Host Id. Value : Content Id
+	 * Key: DatasourceId # Host Id # Content Id. Value has no significance. it will be set to true if there is 
+	 * a value in cache for the key.
 	 */
-	private final Cache<String, Long> hostAddressUsageCache = CacheBuilder.newBuilder().maximumSize(200000).build();
+	private final Cache<String, Boolean> hostAddressUsageCache = CacheBuilder.newBuilder().maximumSize(200000).build();
 
 	/**
 	 * Construct a HostAddressManager for the given SleuthkitCase.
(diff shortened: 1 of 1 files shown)


bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -52,12 +52,19 @@
...
 	/**
-	 * Recently added or accessed Host Address Object Ids are cached. This is
+	 * Recently added or accessed Host Address Objects are cached. This is
 	 * here to improve performance of the
 	 * {@link #hostAddressExists(org.sleuthkit.datamodel.HostAddress.HostAddressType, java.lang.String)}
-	 * check.
+	 * check as well as the {@link #getHostAddress(org.sleuthkit.datamodel.HostAddress.HostAddressType, java.lang.String) }
 	 */
-	private final Cache<String, Long> recentHostAddresstCache = CacheBuilder.newBuilder().maximumSize(200000).build();
+	private final Cache<String, HostAddress> recentHostAddresstCache = CacheBuilder.newBuilder().maximumSize(200000).build();
```
APriestman, line 60: Can you fix the existing typo here Addresst->Address

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -98,6 +105,12 @@
 	 * @throws TskCoreException
 	 */
 	private Optional<HostAddress> getHostAddress(HostAddress.HostAddressType type, String address, CaseDbConnection connection) throws TskCoreException {
+		
+		HostAddress hostAddress = recentHostAddresstCache.getIfPresent(type.getId() + "#" + address.toLowerCase());
```
APriestman, line 109: Can we put the "type.getId() + "#" + address.toLowerCase()" part into a method like "createRecentHostAddressKey()"?

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -587,13 +604,21 @@ public boolean hostNameAndIpMappingExists(long addressObjectId) throws TskCoreEx
 	 * @param hostAddress The host address.
 	 */
 	public void addUsage(Content content, HostAddress hostAddress) throws TskCoreException {
+		
+		String key = content.getDataSource().getId() + "#" + hostAddress.getId();
+		Long cachedValue = hostAddressUsageCache.getIfPresent(key);
+		if (null != cachedValue && cachedValue == content.getId()) {
```
APriestman, line 610: Is there any reason to do the second part of this check?

jayaramcs: @APriestman Yes; usage is associated with both a data source as well as the content. The key is on datasourceId.

APriestman: Ok I see that now, but then why wouldn't the key be on all three fields? Isn't the point just to check if you've already added (hostAddrId, contentId, datasourceId) to the database?

jayaramcs: I have updated it with all 3 now

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
jayaramcs: @APriestman changes are in

APriestman (approved):

PR #2415, opened 2021-05-27
Adding API. getAggregateScores for list of object ids

jayaramcs:
This is a performance optimization to lookup scores in batch

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
@@ -65,8 +65,11 @@ public class ScoringManager {
 	 * @throws TskCoreException
 	 */
 	public Score getAggregateScore(long objId) throws TskCoreException {
+		db.acquireSingleUserCaseReadLock();
 		try (CaseDbConnection connection = db.getConnection()) {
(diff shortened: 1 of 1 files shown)


bindings/java/src/org/sleuthkit/datamodel/ScoringManager.java
```diff
@@ -66,6 +70,46 @@ public Score getAggregateScore(long objId) throws TskCoreException {
...
+			return Collections.emptyMap();
+		}
+
+		String queryString = "SELECT obj_id, significance, method_category FROM tsk_aggregate_score WHERE obj_id in "
+				+ objIds.stream().map(l -> l.toString()).collect(Collectors.joining(",", "(", ")"));
+
+		Map<Long, Score> results = objIds.stream().collect(Collectors.toMap( key -> key, key -> Score.SCORE_UNKNOWN));
+		try (CaseDbConnection connection = db.getConnection()) {
+
+			db.acquireSingleUserCaseReadLock();
```
APriestman, line 95: Get the read lock first

jayaramcs: There were a bunch more in that class.

APriestman (approved):

Issue #2429, opened 2021-05-31
Integrate OSS-Fuzz fuzz targets directly into Sleuthkit


joachimmetz:
Per https://github.com/google/oss-fuzz/pull/5682#pullrequestreview-646832018

```
... but it would be nice if we could move the fuzzers and build files upstream.
```

joachimmetz:
Changes to integrate fuzz targets into Sleuthkit code base https://github.com/sleuthkit/sleuthkit/pull/2430

PR #2430, opened 2021-05-31
Added OSS-Fuzz fuzz targets #2429

joachimmetz:
This PR incorporates files that were copied from https://github.com/google/oss-fuzz/tree/80d39ec418fda8ea9ee6e3e91d7d9341affaf421/projects/sleuthkit under Apache License 2

### DIFF
--- a/.github/workflows/build_ossfuzz.yml
+++ b/.github/workflows/build_ossfuzz.yml
@@ -1,0 +1,39 @@
+# Build OSSFuzz fuzz targets from source.
+name: build_ossfuzz
+on:
+  push:
+    branches:
+      - '**'
+  pull_request:
+    branches:
+      - main
+      - develop
+permissions: read-all
+jobs:
+  build_ossfuzz:
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        include:
+        - architecture: 'x64'
+          compiler: 'gcc'
+          configure_options: ''
+    steps:
+    - name: Install build dependencies
+      run: |
+        sudo apt-get -y install git
+    - uses: actions/checkout@v4
+      with:
+        repository: google/oss-fuzz
+        path: oss-fuzz
+    - name: Build OSSFuzz fuzz targets
+      working-directory: oss-fuzz
+      run: |
+        # TODO: update ossfuzz script and then remove this work-around
+        # Work around hardcoded -Werror flags https://github.com/sleuthkit/sleuthkit/issues/3012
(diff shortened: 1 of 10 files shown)


joachimmetz:
@bcarrier @rcordovano what is the status of the review of this PR ?

joachimmetz:
@gdicristofaro any questions / updates?

joachimmetz:
@simsong if you're looking at changes to improve CI and testing, PTAL

simsong:
We've moved to a single Makefile, because 'recursive make is considered harmful.' But if you want to keep this as a different makefile, I'm okay with that.

joachimmetz:
> We've moved to a single Makefile, because 'recursive make is considered harmful.' 

Any additional context what is "harmful" about it?

> But if you want to keep this as a different makefile, I'm okay with that.

I've rebased the PR, kept the separate Makefile.am for now, but can move it into the main one

simsong:
> > We've moved to a single Makefile, because 'recursive make is considered harmful.'
> 
> Any additional context what is "harmful" about it?

Yes!

https://accu.org/journals/overload/14/71/miller_2004/

joachimmetz:
> https://accu.org/journals/overload/14/71/miller_2004/

thanks I'll take a closer read, but at first glance looks like this is specifically about "vanilla make" not about make in combination with autotools

simsong:
It is specifically about auto tools! We used it for years at sandstorm as a
guiding principle.

joachimmetz:
> It is specifically about auto tools! 

I might be missing the point then, given technically there is nothing preventing automake from generating a single Makefile.

simsong:
That’s what we are doing!

joachimmetz:
> That’s what we are doing!

I meant more as automake generating a single Makefile based on recursive Makefile.am configuration.

the automake link I added gave me a bit more context from the autotools perspective what are the tradeoffs.

Let me know if the ossfuzz Makefile.am should be moved into the main Makefile.am given it's specific use case might not be providing a lot of benefit

simsong:
Ah. Yes. We would prefer it to be in the root Makefile.am.

joachimmetz:
I'll take a look tomorrow, to move it, should not be too involved.

joachimmetz:
@simsong PTAL, moved ossfuzz/Makefile.am into Makefile.am and added a CI test to test the ossfuzz build

simsong (approved):
Thanks for the fuzzer!

PR #2431, opened 2021-06-01
7670 - Rename method category to priority in score

bcarrier:


bindings/java/src/org/sleuthkit/datamodel/Score.java
```diff
@@ -209,32 +205,32 @@ public String toString() {
...
-	public static final Score SCORE_UNKNOWN = new Score(Significance.UNKNOWN, MethodCategory.AUTO);
+	public static final Score SCORE_UNKNOWN = new Score(Significance.UNKNOWN, Priority.NORMAL);
 	
 	// Score is a combination of significance and method category.
 	private final Significance significance;
-	private final MethodCategory methodCategory;
+	private final Priority priority;
 
-	public Score(Significance significance, MethodCategory methodCategory) {
+	public Score(Significance significance, Priority methodCategory) {
```
APriestman, line 214: Can we rename this parameter?

bcarrier: should be fixed.

APriestman (approved):

PR #2435, opened 2021-06-02
7671 added constants

gdicristofaro:


bindings/java/src/org/sleuthkit/datamodel/Score.java
```diff
@@ -205,6 +205,11 @@ public String toString() {
 		}
 	}
 
+	public static final Score SCORE_NOTABLE = new Score(Significance.NOTABLE, Priority.NORMAL);
```
rcordovano, line 208: Given that users will be typing Score.SCORE_LIKELY_NOTABLE, etc., perhaps we can save a few keystrokes with names like Score.LIKELY_NOTABLE, since "Score" is already part of the constant's name?

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/Score.java
gdicristofaro: Sure.  We could do that.  I did that to mirror `SCORE_UNKNOWN`.  Do we want to change that one as well?

rcordovano: Never mind.

PR #2437, opened 2021-06-03
Assign correct device block size in HFS

### LINKED ISSUE #2436: Inconsistent device block size assigned in hfs.c
hfs.c assigns device block size (dev_bsize) as follows:

```
    /* this isn't really accurate; fs->block_size reports only the size
       of the allocation block; the size of the device block has to be
       found from the device (allocation block size should always be
       larger than device block size and an even multiple of the device
       block size) */
    fs->dev_bsize = fs->block_size =
        tsk_getu32(fs->endian, hfs->fs->blk_sz);
```

All other filesystems use the value already stored in TSK_IMG_INFO::sector_size

InvisiblePhil:
Use TSK_IMG_INFO::sector_size, as this may differ from block_size.
All other filesystems use this value 

Closes #2436


simsong (approved):
Looks better, although I would really like to have a test image to confirm this.

simsong:
@bcarrier  — I accepted this. Please let me know if you have any concerns.

PR #2440, opened 2021-06-03
7643 - Added getUniquePath to LocalFile

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -1402,6 +1404,24 @@ public void save(CaseDbTransaction transaction) throws TskCoreException {
 	public Optional<Long> getOsAccountObjectId() {
 		return Optional.ofNullable(osAccountObjId);
 	}
+	
+	@Override
+	public String getUniquePath() throws TskCoreException {
+
+		if (uniquePath == null) {
+			if (getDataSource() instanceof LocalFilesDataSource) {
```
APriestman, line 1412: getDataSource() should end up using the frequentlyUsedContentMap instead of doing a db lookup each time, but it might be better to only call getDataSource() once and store the result to make this slightly more efficient

APriestman (approved):

PR #2450, opened 2021-06-16
7701 Add extension to findFiles search

APriestman:


eugene7646 (approved):

bindings/java/src/org/sleuthkit/datamodel/FileManager.java
```diff
@@ -91,14 +91,17 @@
 			normalizedPath = normalizedPath + "/";
 		}
 		
-		String query = "SELECT tsk_files.* FROM tsk_files JOIN tsk_objects ON tsk_objects.obj_id = tsk_files.obj_id WHERE parent_path = ? AND name = ? AND data_source_obj_id = ?";
+		String ext = SleuthkitCase.extractExtension(name);
```
bcarrier, line 94: We should probably not include 'ext' in they query if there isn't an extension. And we shoudl add a comment that this is being done as an optimization

APriestman:
@eugene7646 I just realized this is still out there. Do you want to look the changes real quick?

eugene7646 (approved):

PR #2451, opened 2021-06-17
7698 Make getArtifactType public

APriestman:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -4789,7 +4789,7 @@ String addSourceToArtifactAttribute(BlackboardAttribute attr, String source) thr
 	 *						    or no value is found.
 	 *
 	 */
-	BlackboardArtifact.Type getArtifactType(int artTypeId) throws TskCoreException {
+	public BlackboardArtifact.Type getArtifactType(int artTypeId) throws TskCoreException {
```
rcordovano, line 4792: I'd prefer to move this method to Blackboard and make it public there rather than adding to the public API of SleuthkitCase. There is actually an extant Jira story about moving methods from SleuthkitCase to Blackboard. Can do?

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
APriestman: Shouldn't be a problem. Unfortunately I'm going to need a new version addLocalFile() to copy the OS account info

rcordovano: Maybe that belongs in FIleManager?

PR #2459, opened 2021-06-25
7760 case uco getters

gdicristofaro:


case-uco/java/test/org/sleuthkit/caseuco/FacetDeserializerTests.java
```diff
@@ -0,0 +1,179 @@
+/*
```
eugene7646, line 1: @gdicristofaro Shouldn't we have our copyright here?

case-uco/java/test/org/sleuthkit/caseuco/TestSuite.java
```diff
@@ -0,0 +1,22 @@
+/*
```
eugene7646, line 1: @gdicristofaro Shouldn't we have our copyright here?

case-uco/java/test/org/sleuthkit/caseuco/TestSuite.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Runs all regression tests and contains utility methods for the tests The
+ * default ant target sets properties for the various folders.
+ */
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+    FacetDeserializerTests.class
+})
+public class TestSuite {
+
```
eugene7646, line 21: Is this class supposed to be empty?

gdicristofaro: I was mimicking the test suite in TSK proper with this except I just needed it for an entry point so that it runs the test suites marked in the annotation.  I think the `DataModelTestSuite` has some static helper methods, but I didn't need them for the CASE UCO code.

eugene7646 (approved):

PR #2496, opened 2021-07-23
7834 - Added TagName and TagSet events

kellykelly3:


wschaeferB (approved):

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -12922,7 +12899,7 @@ long getId() {
 				+ "FROM tsk_objects INNER JOIN blackboard_artifacts " //NON-NLS
 				+ "ON tsk_objects.obj_id=blackboard_artifacts.obj_id " //NON-NLS
 				+ "WHERE (tsk_objects.par_obj_id = ?)"),
-		INSERT_OR_UPDATE_TAG_NAME("INSERT INTO tag_names (display_name, description, color, knownStatus) VALUES (?, ?, ?, ?) ON CONFLICT (display_name) DO UPDATE SET description = ?, color = ?, knownStatus = ?"),
+//		INSERT_OR_UPDATE_TAG_NAME("INSERT INTO tag_names (display_name, description, color, knownStatus) VALUES (?, ?, ?, ?) ON CONFLICT (display_name) DO UPDATE SET description = ?, color = ?, knownStatus = ?"),
```
rcordovano, line 12902: Can we just remove this code instead of commenting it out?

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
kellykelly3: Of Course.  That was an accidental left behind.

PR #2507, opened 2021-07-30
Fix OOB write in ntfs_proc_compunit and integer overflow in ntfs_uncompress_setup

joachimmetz:


joachimmetz:
@bcarrier can you give this some priority, an OOB write is a pretty serious defect also see https://cwe.mitre.org/data/definitions/787.html

joachimmetz:
https://oss-fuzz.com/testcase-detail/5764903091699712 if you need the POC, access to this is ACL'ed

bcarrier:
Manually applied to release branch in 3fe8fc91d

PR #2517, opened 2021-08-08
Speed up tsk_fs_path2inum by not loading attributes when not needed

Eran-YT:
Currently tsk_fs_path2inum uses tsk_fs_dir_get which loads the attributes for the directory it gets, which tsk_fs_path2inum doesn't actually need in most cases, and this is very expensive (They are read from disk).
This PR changes this behavior to only load the attributes if needed (only a very specific case involving NTFS), thus speeding up the function.
By my benchmarks the speed up is about by a factor of >50.
This also reduces the disk I/O.


simsong (approved):
Nice.

PR #2518, opened 2021-08-09
Refactor and optimize NTFS parent map

Eran-YT:
Use std::unordered_map (hashmap) instead of std::map (binary tree)
some refactors and optimizations


simsong:
@uckelman-sf — What do you think about this? You're the C++ wizard...

uckelman-sf:
The changes look ok. The problems I see were already there in the original code.  `std::unordered_map` is usually faster than `std::map`, so is preferable when you're doing lookups but don't need in-order traversal.

PR #2524, opened 2021-08-13
Fixed OOB read and write in ext4_load_attrs_inline

joachimmetz:


joachimmetz:
@bcarrier @APriestman PTAL unsure why the original code calculates the size as `inode_data_len + ea_inline_data_len` and then writes the buffer starting at `inode_data_len`. This makes me wonder if the original code is doing the correct thing.

joachimmetz:
This also address an OOB write https://cwe.mitre.org/data/definitions/787.html so please give this some urgency.

APriestman (approved):

PR #2532, opened 2021-09-02
fixes #2423: use Datetime::TimeZone for timezone handling

### LINKED ISSUE #2423: mactime -z does not work?
What is the expected behavior of `mactime -z` ?

```
mactime -b bodyfile -d -h

Sat Jan 05 2019 12:39:59,0,macb,r/rrwxrwxrwx,0,0,34,"\emptyfile"
Sat Jan 05 2019 12:39:59,84,macb,r/rrwxrwxrwx,0,0,34,"\emptyfile ($FILE_NAME)"
```

```
mactime -b bodyfile -d -h -z CET

Sat Jan 05 2019 12:39:59,0,macb,r/rrwxrwxrwx,0,0,34,"\emptyfile"
Sat Jan 05 2019 12:39:59,84,macb,r/rrwxrwxrwx,0,0,34,"\emptyfile ($FILE_NAME)"
```

```
mactime -b bodyfile -d -h -z Europe/Amsterdam

Sat Jan 05 2019 12:39:59,0,macb,r/rrwxrwxrwx,0,0,34,"\emptyfile"
Sat Jan 05 2019 12:39:59,84,macb,r/rrwxrwxrwx,0,0,34,"\emptyfile ($FILE_NAME)"
```

janstarke:


simsong:
Hi. We need a test case for this. Can you supply one?

janstarke:
While working on test data, I stumbled across the `mactime` help text

```
-z: Specify the timezone the data came from (in the local system format) (does not work with -y)
```

and I'm wondering what this could be. According to <https://wiki.sleuthkit.org/index.php?title=Body_file>,

> The times are reported in UNIX time format.

UNIX time

> measures time by the number of non-leap seconds that have elapsed since 00:00:00 UTC on 1st January 1970, the Unix [epoch](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap03.html#tag_21_03_00_20).

# Option 1

So, every *creator* of a bodyfile output should make sure that the timestamps are related to UTC, no matter where the data originally came from.

So, in my eyes, the flag `-z` could (and should) be removed, as the origin timezone is always UTC. I suggest to reject this PR and remove the `-z` flag.

However, it could be interesting to map the timestamps to some local time, in order to correlate them to some other timestamps which you already have.

# Option 2

The specification of the bodyfile format should be changed in a way that the origin timezone format is explicitly undefined. I this case, I'd continue to work on test cases

simsong:
Hi. Are you aware that the `-z` option is for specifying a timezone when importing times from a FAT12/16/32 file system, which puts timestamps in local time, until NTFS and other file systems, which store timestamps in absolute time?

janstarke:
Yes, I am. I also know that tools like `fls` are not timestamp aware.

However, as I already stated, the bodyfile format does not allow to store timestamps from a timezone different than UTC, which seems to be a conflict to me. I'm not happy with working on a feature which violates the file format specification :disappointed:

simsong:
Great. I think that we can't change the file format. So my understanding is that the `-z` flag lets the FAT file systems have their timestamps stored in UTC. 

We *can't* remove the `-z` flag because that will break every script that  specifies it. We can the program change the flag so that it is ignored under certain conditions, but we can't remove it.

We *can* change the documentation to state that the origin timezone is undefined if the `-z` flag is not specified.  

How would you like to proceed?

janstarke:
Sounds goods. In my eyes, `UTC` would be a plausible default for `-z`, so we could change the documentation to 

```
-z: Specify the timezone the data came from (in the local system format) (defaults to 'UTC' if not specified; does not work with -y)
```

This should break no existing scripts and we had a defined behavior of `mactime`. In addition, I now know how my test cases could look like.

simsong:
> Sounds goods. In my eyes, `UTC` would be a plausible default for `-z`, so we could change the documentation to
> 
> ```
> -z: Specify the timezone the data came from (in the local system format) (defaults to 'UTC' if not specified; does not work with -y)
> ```
> 
> This should break no existing scripts and we had a defined behavior of `mactime`. In addition, I now know how my test cases could look like.

Great! I look forward to the test case. I have now implemented the catch2 test rig in unit_tests, so if you want to add a catch2 test that runs mactimes as a subprocess, that would be neat.

simsong:
@Taha-Ebrahim — Can you take a look at this and let me know what you think?

simsong:
This needs to wait for https://github.com/sleuthkit/sleuthkit/issues/3392

simsong:
@janstarke — what do you think of the copilot review?

PR #2533, opened 2021-09-03
687: IPv6 address regex don't handle zone specifiers

raman-bt:


eugene7646:
@raman-bt How should I test this PR? What story is this for?

raman-bt:
> @raman-bt How should I test this PR? What story is this for?

TSK-687 is the story.  I added some unit tests.  And perhaps make sure it doesn't break anything for Autopsy.

eugene7646 (approved):

bindings/java/src/org/sleuthkit/datamodel/HostAddressManager.java
```diff
@@ -774,4 +782,44 @@ private static boolean isIPv6(String ipAddress) {
...
 	}
+	
+	/**
+	 * Normalizes an address.
+	 *
+	 * @param address
+	 *
+	 * @return Normalized address.
+	 */
+	private static String getNormalizedAddress(String address) {
```
jayaramcs, line 793: @raman-bt  shouldn't this also do the toLower() since it returns a normalized form?

raman-bt: It intentionally does NOT convert to lowercase so that we can preserve the case and leave it to caller to convert if necessary.  There's a comment to that effect on getNormalizedIPV6Address() just a few lines below.  

I will add the comment here as well.

jayaramcs: Ok. that makes sense. I did not notice the other comment. sorry.

crayy8:
This PR is good to go. I've verified that everything is working as expected with a disk image I created with both regular IPv6 addresses as well as IPv6 with zone IDs. 

\\fstore\forensics\cybertriage\users\chris\xp-sp3-v4-ipv6.img. 

Found in Johns NTUser.DAT (Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\MountPoints2)
##fe80--a45c-65bd-9e02-6d06s11.ipv6-literal.net#c$
##0000--0001.ipv6-literal.net#MountTest

Found in Johns NTUser.DAT (Software\Microsoft\Terminal Server Client\Servers)
fe80::a45c:65bd:9e02:6d06%11
fe80::33:65bd:9e02:3

crayy8 (approved):

PR #2536, opened 2021-09-08
Remove persona artifact

APriestman:


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -1880,13 +1879,7 @@ public int hashCode() {
...
 		 */
 		TSK_PREVIOUSLY_NOTABLE(71, "TSK_PREVIOUSLY_NOTABLE",
-				bundle.getString("BlackboardArtifact.tskPreviouslyNotable.text"), Category.ANALYSIS_RESULT),
-		
-		/**
-		 * *TEMPORARY* Indicates that the artifact is associated with a persona.
-		 */
-		TSK_MATCHING_PERSONA(72, "TSK_MATCHING_PERSONA",
-				bundle.getString("BlackboardArtifact.tskMatchingPersona.text"), Category.ANALYSIS_RESULT);
+				bundle.getString("BlackboardArtifact.tskPreviouslyNotable.text"), Category.ANALYSIS_RESULT);
```
eugene7646, line 1882: @APriestman Based on https://github.com/sleuthkit/sleuthkit/pull/2506/files , I think "public static final Type TSK_MATCHING_PERSONA" on line 1300 also needs to be removed, as well as Bundle.properties and Bundle.properties-MERGED

eugene7646 (approved):

PR #2545, opened 2021-09-18
Fixed integer overflow in ext4_load_attrs_inline leading to OOB-read

joachimmetz:


joachimmetz:
If not checked this can lead to an OOB-read in https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ext2fs.c#L677

APriestman (approved):

joachimmetz:
https://osv.dev/vulnerability/OSV-2021-1309 though this assessment does not appear to be covering the full extent of the issue

PR #1276, opened 2018-07-04
Fix several NTFS decompression bugs (II)

uckelman-sf:
This PR fixes decompression errors in NTFS compressed files that have an initialized size which differs from the logical size, and also allows decompression of an NTFS compressed file to continue in the event that a  corrupted compression unit is encountered.

This PR includes the patches from PR #1183, and an additional patch for the remaining problem noted there.

PR #2547, opened 2021-09-21
Refactor NTFS Decompression / initsize checks

bcarrier:


bcarrier:
@uckelman-sf : Can you take a quick look at this. It should have the same behavior as PR #1276 that got merged in, but with more explicit behavior (at the expense of potentially unneeded memsets) about making sure the uncompressed buffer is 0s past initsize.

PR #2549, opened 2021-09-22
8015, 8027- Updated getShortDescription

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -233,44 +232,108 @@ public String getDisplayName() {
...
+				case TSK_WIFI_NETWORK:
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_SSID));
+					break;
+				case TSK_REMOTE_DRIVE:
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_REMOTE_PATH));
+					break;
+				case TSK_SERVICE_ACCOUNT:
+				case TSK_SCREEN_SHOTS:
+				case TSK_DELETED_PROG:
+				case TSK_METADATA:
```
rcordovano, line 249: TSK_PROG_NAME is optional for TSK_METADATA.

Given that we do not enforce the artifact catalog requirements in any way, we should actually be prepared for all cases where the expected attribute is not present. Basically, we need to check for a null return value from getAttributes() in all cases.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -233,44 +232,108 @@ public String getDisplayName() {
...
+				case TSK_OS_INFO:
+				case TSK_PROG_NOTIFICATIONS:
+				case TSK_PROG_RUN:
+				case TSK_RECENT_OBJECT:
```
rcordovano, line 253: TSK_PROG_NAME is optional for TSK_RECENT_OBJECT.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -233,44 +232,108 @@ public String getDisplayName() {
...
+				case TSK_USER_DEVICE_EVENT:
+				case TSK_WEB_SEARCH_QUERY:
```
rcordovano, line 255: TSK_PROG_NAME is optional for TSK_WEB_SEARCH_QUERY.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -233,44 +232,108 @@ public String getDisplayName() {
...
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_PROG_NAME));
+					break;
+				case TSK_BLUETOOTH_PAIRING:
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_DEVICE_NAME));
+					break;
+				case TSK_ACCOUNT:
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_ACCOUNT_TYPE));
+					break;
+				case TSK_WEB_CATEGORIZATION:
```
rcordovano, line 264: TSK_WEB_CATEGORIZATION is not in the artifacts catalog.

I'm not going to go over every item here...I'll leave that to you :)

Overall comment is that we need to check that we are:
* Dealing with missing required attributes
* Not depending on optional attributes...we may need to change required/optional, but that could be a separate story.
* Make sure that all non-deprecated types are in the artifacts catalog, that could be a sepaarate story.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -233,44 +232,108 @@ public String getDisplayName() {
...
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_MAC_ADDRESS));
+					break;
+				case TSK_WIFI_NETWORK:
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_SSID));
+					break;
+				case TSK_REMOTE_DRIVE:
+					attr = getAttribute(new BlackboardAttribute.Type(ATTRIBUTE_TYPE.TSK_REMOTE_PATH));
+					break;
+				case TSK_SERVICE_ACCOUNT:
+				case TSK_SCREEN_SHOTS:
```
rcordovano, line 247: TSK_SCREEN_SHOTS is not deprecated, but it's not in the artifacts catalog. We either can address that as part of this story or we can make a list of missing entries and make a new story.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
```diff
@@ -233,44 +232,108 @@ public String getDisplayName() {
...
+				case TSK_DELETED_PROG:
+				case TSK_METADATA:
+				case TSK_OS_INFO:
+				case TSK_PROG_NOTIFICATIONS:
+				case TSK_PROG_RUN:
+				case TSK_RECENT_OBJECT:
+				case TSK_USER_DEVICE_EVENT:
```
rcordovano, line 254: I don't know what a TSK_USER_DEVICE_EVENT is. It's not in the artifact catalog.

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
kellykelly3: @rcordovano The code does.  See the end of the method.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
kellykelly3: Correct. Which would you prefer instead?  If PROG_NAME is not available it will default to getDisplayName()

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
kellykelly3: Correct. Which would you prefer instead?  If PROG_NAME is not available it will default to getDisplayName()

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
kellykelly3: @rcordovano 
1. If TSK_WEB_CATEGORIZATION is not in the catalog then it probably predates my changes.  What would you like done in this case?
2. The code is dealing with missing attributes.
3. Then what should be used?  Some of this artifacts have things list TSK_TEXT as the only required attribute I used my best judgement with the assumption if the attribute wasn't there then it will default the BlackboardArtifact.getName.
4. That's a maintenance nightmare.  The expectation that this method should be updated with every artifact change is a bit much.  Also what is the issue with support deprecated artifacts.  The code could still come across them.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
kellykelly3: @rcordovano I only added artifacts that are in the catalog.  I did not make changes to the existing code.

bindings/java/src/org/sleuthkit/datamodel/BlackboardArtifact.java
kellykelly3: @rcordovano I didn't pull artifacts out of the sky, someone else put this here in the original version.

wschaeferB (approved):

PR #2550, opened 2021-09-27
Ct4635 - account instance type filters

raman-bt:
- Added instance_type to uniqueness constraint for tsk_os_account_instances table
- updated the newOsAccountInstance() method to to add an instance row if the new instance is more significant than all the instances already in the table.
- updated the getOsAccountInstance() method to return the most significant instance 
- added code to schema for tsk_os_account_instances  on upgrades.


raman-bt:
Changed this to draft.  

The newOsAccountInstance() is being changed to return OsAccountInstance (in a smaller separate PR).  This should be merged following that...

kellykelly3:
@raman-bt You have a merge conflict.  Could you please fix?

raman-bt:
Fixed merge conflict.

kellykelly3:
@raman-bt 
_updated the getOsAccountInstance() method to return the most significant instance_

How might this impact other methods that use the instance table to get accounts?  For example OsAccountManager.getIOsAccount(Host), OsAccountManger.getOsAccountsByDataSourceObjId()

I see you did not modify any of the merging code.  How will this affect the merging code?  How will your changes affect mergeOSAccount when there are multiple instances for the same Os Account\Data Source?  I wasn't involved in the merge code so maybe @APriestman can speak to this concern?

APriestman:
> I see you did not modify any of the merging code. How will this affect the merging code? How will your changes affect mergeOSAccount when there are multiple instances for the same Os Account\Data Source? I wasn't involved in the merge code so maybe @APriestman can speak to this concern?

Yes I believe there is a slight problem with the merge code. In OsAccountManager.mergeOsAccounts(), I think we need to add a check on the type to the DELETE FROM tsk_os_account_instances query.

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -1210,18 +1257,26 @@ public void addExtendedOsAccountAttributes(OsAccount account, List<OsAccountAttr
 	 */
 	private List<OsAccountInstance> getOsAccountInstances(String whereClause) throws TskCoreException {
 		List<OsAccountInstance> osAcctInstances = new ArrayList<>();
-		String querySQL = "SELECT * FROM tsk_os_account_instances WHERE " + whereClause;
+		
+		// this query returns only the most significant instance type (least ordinal) for each instance,
+		// that matches the specified WHERE clause.
```
APriestman, line 1262: Can you add this to the method comment block?

APriestman (approved):

kellykelly3 (approved):

PR #2560, opened 2021-09-29
Removed Timeline event WEB_COOKIE_END

kellykelly3:


bindings/java/src/org/sleuthkit/datamodel/TimelineEventType.java
```diff
@@ -657,12 +656,7 @@ public int compare(TimelineEventType o1, TimelineEventType o2) {
 			new Type(TSK_DATETIME_ACCESSED),
 			new Type(TSK_URL));
 
-	TimelineEventType WEB_COOKIE_END = new URLArtifactEventType(42,
```
gdicristofaro, line 660: I think `TSK_DATETIME_END` needs to be removed from https://github.com/sleuthkit/sleuthkit/blob/develop/bindings/java/test/org/sleuthkit/datamodel/TimelineEventTypesTest.java#L121 too.

gdicristofaro (changes requested):

PR #2566, opened 2021-10-02
Fixed multiple bounds issues in tsk_UTF16toUTF8 and iso9660_load_inodes_dir

joachimmetz:


joachimmetz:
One of the issues here is that a uint8 buffer is casted to uint16 but the length/size of the buffer is not even before it is passed to tsk_UTF16toUTF8. (And tsk_UTF16toUTF8 casts it back to a uint8 when reading 2 bytes ???). So it is possible this issue in more places where tsk_UTF16toUTF8 is called

PR #2567, opened 2021-10-02
Fixed OOB-read in ntfs_dent_copy

joachimmetz:


bcarrier:
Thanks! This was fixed in another PR as well.

PR #2590, opened 2021-10-22
7953 interesting item replace file and artifact

wschaeferB:


bindings/java/src/org/sleuthkit/datamodel/BlackboardAttribute.java
```diff
@@ -455,7 +455,7 @@ String getSourcesCSV() {
 		public static final Type TSK_ENTROPY = new Type(29, "TSK_ENTROPY", bundle.getString("BlackboardAttribute.tskEntropy.text"), TSK_BLACKBOARD_ATTRIBUTE_VALUE_TYPE.DOUBLE);
 
 		// TSK_HASHSET_NAME (id: 30) has been deprecated.  Please use TSK_SET_NAME instead.
-		// TSK_INTERESTING_FILE (id: 31) has been deprecated.  Please use TSK_INTERESTING_FILE_HIT instead.
+		// TSK_INTERESTING_FILE (id: 31) has been deprecated.  Please use TSK_INTERESTING_ITEM instead.
```
rcordovano, line 458: Let's say: "Please use TSK_INTERESTING_ITEM analysis result instead." On line 457, we are replacing an attribute type with an attribute type, but here we are referring the developer to an artifact type, so let's make that a little more explicit."

rcordovano (changes requested):

bindings/java/src/org/sleuthkit/datamodel/BlackboardAttribute.java
wschaeferB: I will change the wording as I believe understand the intent of this comment. I just would like to clarify so I know I am not missing something else which is going on here. I replaced an artifact type with an artifact type here, correct?

rcordovano (approved):

bindings/java/src/org/sleuthkit/datamodel/BlackboardAttribute.java
rcordovano: No, @wschaeferB, you are not replacing an artifact type with an artifact type. You are replacing an attribute type that used to be attached to something else (TSK_GEN_INFO?) with a dedicated analysis result. Here is the corresponding code where the attribute type was deprecated in BlackboardAttribute.ATTRIBUTE_TYPE:

		/**
		 * @deprecated Use a TSK_INTERESTING_FILE_HIT artifact instead.
		 */
		@Deprecated
		TSK_INTERESTING_FILE(31, "TSK_INTERESTING_FILE", //NON-NLS
				bundle.getString("BlackboardAttribute.tskInterestingFile.text"),
				TSK_BLACKBOARD_ATTRIBUTE_VALUE_TYPE.LONG), 

What you are doing here is supplying the deprecation of the corresponding BlackboardAttribute.Type constant.  IN all truth, the Type constant should never have been created - TSK_INTERESTING_FILE attributes have been deprecated for many years. I would be fine with just deleting the BlackboardAttribute.Type constant, but I don't mind leaving it as is, either.

rcordovano (approved):

PR #2596, opened 2021-11-02
8145 prepared statements

gdicristofaro:


bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -675,5 +723,226 @@ private void validateSQL(String sql) throws TskCoreException {
...
+		private final CaseDbConnection connection;
+		private final PreparedStatement preparedStatement;
+
+		/**
+		 * Main constructor.
+		 *
+		 * @param connection        The db connection.
+		 * @param preparedStatement The delegate prepared statement.
+		 */
+		CasePreparedStatement(CaseDbConnection connection, PreparedStatement preparedStatement) {
```
rcordovano, line 743: It looks to me like this constructor should be private. I believe only the enclosing class needs to create instances of CasePreparedStatement objects.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -675,5 +723,226 @@ private void validateSQL(String sql) throws TskCoreException {
 		 * TODO (JIRA-5950): Need SQL injection defense in CaseDbAccessManager 
 		 */
 	}
+	
+	/**
+	 * A wrapper around a PreparedStatement to execute queries against the
+	 * database.
+	 */
+	@Beta
+	public static class CasePreparedStatement implements AutoCloseable {
```
rcordovano, line 732: For consistency with CaseDbAccessManager and CaseDbConnection, and perhaps for clarity, I would rename this class as CaseDbPreparedStatement.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -602,6 +607,49 @@ public void select(final String sql, final CaseDbAccessQueryCallback queryCallba
...
+	 *
+	 * @throws TskCoreException
+	 */
+	@Beta
+	public CasePreparedStatement prepareSelect(String sql) throws TskCoreException {
+		String selectSQL = "SELECT " + sql; // NON-NLS
+		try {
+			CaseDbConnection connection = tskDB.getConnection();
+			PreparedStatement statement = connection.getPreparedStatement(selectSQL, Statement.NO_GENERATED_KEYS);
+			return new CasePreparedStatement(connection, statement);
```
rcordovano, line 627: If we are passing the CaseDbConnection and the PreparedStatement into the CasePreparedStatement constructor, why not make the constructor do the work? The argument to the constructor could then be the SQL instead.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -602,6 +607,49 @@ public void select(final String sql, final CaseDbAccessQueryCallback queryCallba
...
+	 *
+	 * @throws TskCoreException
+	 */
+	@Beta
+	public void select(CasePreparedStatement preparedStatement, CaseDbAccessQueryCallback queryCallback) throws TskCoreException {
+		tskDB.acquireSingleUserCaseReadLock();
+		try (ResultSet resultSet = preparedStatement.getStatement().executeQuery()) {
+			queryCallback.process(resultSet);
+		} catch (SQLException ex) {
+			throw new TskCoreException("Error running SELECT query.", ex);
```
rcordovano, line 647: We should make debugging easier by including the SQL that failed to execute in the error message. This might require the CasePreparedStatement to provide a getter for its SQL, but see previous comment where I suggest that we pass the SQL to the constructor for that class.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -602,6 +607,49 @@ public void select(final String sql, final CaseDbAccessQueryCallback queryCallba
...
+	 */
+	@Beta
+	public CasePreparedStatement prepareSelect(String sql) throws TskCoreException {
+		String selectSQL = "SELECT " + sql; // NON-NLS
+		try {
+			CaseDbConnection connection = tskDB.getConnection();
+			PreparedStatement statement = connection.getPreparedStatement(selectSQL, Statement.NO_GENERATED_KEYS);
+			return new CasePreparedStatement(connection, statement);
+		} catch (SQLException ex) {
+			throw new TskCoreException("Error creating select prepared statement.", ex);
```
rcordovano, line 629: We should make debugging easier by including the SQL that failed to be prepared in the error message.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -675,5 +723,226 @@ private void validateSQL(String sql) throws TskCoreException {
...
+			this.connection = connection;
+			this.preparedStatement = preparedStatement;
+		}
+
+		/**
+		 * Returns the delegate prepared statement.
+		 *
+		 * @return The delegate prepared statement.
+		 */
+		PreparedStatement getStatement() {
```
rcordovano, line 753: I think that this can be private.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -675,5 +723,226 @@ private void validateSQL(String sql) throws TskCoreException {
...
+			return preparedStatement;
+		}
+
+		/**
+		 * Returns the connection to use the prepared statement with.
+		 *
+		 * @return The connection to use the prepared statement with.
+		 */
+		CaseDbConnection getConnection() {
```
rcordovano, line 762: I don't think this method is needed.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -675,5 +723,226 @@ private void validateSQL(String sql) throws TskCoreException {
...
+				preparedStatement.setObject(parameterIndex, x);
+			} catch (SQLException ex) {
+				throw new TskCoreException(MessageFormat.format("There was an error setting the value at index: {0} to {1}", parameterIndex, x), ex);
+			}
+		}
+
+		@Override
+		public void close() throws SQLException {
+			connection.close();
+			preparedStatement.close();
```
rcordovano, line 944: Doesn't closing the JDBC connection close the JDBC prepared statement? In any case, if we are going to close both explicitly, I think that we need to close the prepared statement first.

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -675,5 +723,226 @@ private void validateSQL(String sql) throws TskCoreException {
...
+		/**
+		 * Resets the parameters in the prepared statement.
+		 *
+		 * @throws TskCoreException
+		 */
+		public void reset() throws TskCoreException {
+			try {
+				preparedStatement.clearParameters();
+			} catch (SQLException ex) {
+				throw new TskCoreException("An error occurred while clearing parameters.", ex);
```
rcordovano, line 775: Some of the exception messages use periods, some do not. It's a trivial point, but consistency is good. I recommend omitting the periods, since it encourages short error messages and I think the periods look funny in the logs, but that's strictly personal preference, of course.

rcordovano (changes requested):

rcordovano (approved):

bindings/java/src/org/sleuthkit/datamodel/CaseDbAccessManager.java
```diff
@@ -675,5 +723,256 @@ private void validateSQL(String sql) throws TskCoreException {
...
+	 * A wrapper around a PreparedStatement to execute queries against the
+	 * database.
+	 */
+	@Beta
+	public class CaseDbPreparedStatement implements AutoCloseable {
+
+		private final CaseDbConnection connection;
+		private final PreparedStatement preparedStatement;
+		private final String originalSql;
+		private final boolean isWriteLockRequired;
```
APriestman, line 738: Tiny thing, but I think the code would be more readable if this was named something like "hasWriteLock" (keep the name for the parameter). You could set it immediately after getting the lock to make things very clear.

APriestman (approved):

PR #2597, opened 2021-11-03
8131 rename trace and bundleproperty in case/uco code

markmckinnon:


case-uco/java/src/org/sleuthkit/caseuco/CaseUcoExporter.java
```diff
@@ -590,13 +590,13 @@ private void assembleGenInfo(String uuid, BlackboardArtifact artifact, List<Json
     }
 
     private void assembleWebHistory(String uuid, BlackboardArtifact artifact, List<JsonElement> output) throws TskCoreException {
-        Trace userNameNode = new BlankTraceNode();
+        ObservableObject userNameNode = new BlankTraceNode();
```
bcarrier, line 593: @markmckinnon We should probably rename BlankTraceNode to BlankObservableObject.

case-uco/java/src/org/sleuthkit/caseuco/UcoObject.java
```diff
@@ -27,6 +27,8 @@
  */
 abstract class UcoObject {
 
+    static final String UCO_OBJECT = "uco-observable:";
```
bcarrier, line 30: I'd name this something like UCO_OBSERV.  I think we're going to have several of these prefixes.

case-uco/java/src/org/sleuthkit/caseuco/CaseUcoExporter.java
markmckinnon: Changed and commited.

case-uco/java/src/org/sleuthkit/caseuco/UcoObject.java
markmckinnon: Changed and commited

rcordovano (approved):

PR #2608, opened 2021-11-15
Support for searching artifacts that contain specific attribute values (8122)

eugene7646:


bindings/java/src/org/sleuthkit/datamodel/Blackboard.java
```diff
@@ -1661,6 +1679,60 @@ public long getArtifactsCount(int artifactTypeID) throws TskCoreException {
...
+		String query = " AND artifacts.artifact_type_id = " + artifactType.getTypeID() //NON-NLS 
+				+ " AND attributes.attribute_type_id = " + attributeType.getTypeID() //NON-NLS
+				+ ((value == null || value.isEmpty()) ? "" : " AND attributes.value_text = '" + value + "'") //NON-NLS
+				+ (showRejected ? "" : " AND artifacts.review_status_id != " + BlackboardArtifact.ReviewStatus.REJECTED.getID()) //NON-NLS
+				+ (dataSourceObjId != null ? " AND artifacts.data_source_obj_id = " + dataSourceObjId : ""); //NON-NLS
+	
+		List<BlackboardArtifact> artifacts = new ArrayList<>();
+		caseDb.acquireSingleUserCaseReadLock();
+		try (CaseDbConnection connection = caseDb.getConnection()) {
+			if (artifactType.getCategory() == BlackboardArtifact.Category.ANALYSIS_RESULT) {
```
kellykelly3, line 1711: You could probably clean up the redudancy in this if else if you use a var for the query and set based on the type, they have one try with the connection, statement and resultset.

eugene7646: @kellykelly3 Done.

kellykelly3 (approved):

PR #2631, opened 2021-12-10
CT-4851 (fix UTF-16 conversion)

crayy8:
- Sanatize UTF-16 data and replace invalid characters with the UTF-16 replace char
 - Cast data to wstring instead of using from_bytes()

### DIFF
--- a/tsk/base/tsk_unicode.c
+++ b/tsk/base/tsk_unicode.c
@@ -580,18 +580,16 @@ tsk_cleanupUTF8(char *source, const char replacement)
  * @param replacement Character to insert into source as needed.
  */
 void
-tsk_cleanupUTF16(TSK_ENDIAN_ENUM endian, wchar_t *source, const wchar_t replacement) {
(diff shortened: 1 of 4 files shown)


tsk/base/tsk_unicode.c
```diff
@@ -580,18 +580,16 @@ tsk_cleanupUTF8(char *source, const char replacement)
  * @param replacement Character to insert into source as needed.
  */
 void
-tsk_cleanupUTF16(TSK_ENDIAN_ENUM endian, wchar_t *source, const wchar_t replacement) {
+tsk_cleanupUTF16(TSK_ENDIAN_ENUM endian, wchar_t *source, size_t wstrlen, const wchar_t replacement) {
```
bcarrier, line 583: I'd call this source_len instead of wstrlen. That looks too much like a function name.

crayy8: Resolved in commit https://github.com/sleuthkit/sleuthkit/pull/2631/commits/d96828ac84f28bf78675ad312f338097fc2b039b

APriestman (approved):

PR #2637, opened 2021-12-19
Fixed integer overflow in ext2fs_dir_open_meta

joachimmetz:


joachimmetz:
Note that the roundup macro requires increment size to be `> 1` and value `< INT64_MAX / value`

APriestman:
I realize it should never come up with real data, but it looks like roundup(x, 1) will return x, which seems valid.

joachimmetz:
@APriestman the integer overflow found by a fuzzer, so realistic data or not, there is a weakness in the code here handling this edge case

bcarrier:
Hi @joachimmetz.  Thanks for this.  We are trying to figure out if there are other uses of roundup() that we need to address and if we should instead make a tsk_roundup() method that does the validation and is more typesafe than the macros. 

Can you expand a bit more on the 'value < INT64_MAX / value' requirement. I'm not following on how this check is related to the problem? What were the fuzzing inputs that caused this to fail?

Thanks.

joachimmetz:
@bcarrier you should have access to the fuzzer POCs in https://bugs.chromium.org/p/oss-fuzz/issues/list?q=sleuthkit&can=2 just look for an integer overflow found by sleuthkit:sleuthkit_fls_ext_fuzzer. 

So `roundup` not `tsk_roundup` is defined here https://github.com/sleuthkit/sleuthkit/blob/0239c5934e348699d0be38f694fb6320252a91fc/tsk/base/tsk_os.h#L23

`roundup(x, y) ( ( ((x)+((y) - 1)) / (y)) * (y) )` as you can see

* if `y == 0` then you have a divide by zero
* if `y == 1` then the result is x (and no need to so a roundup)
* if `y < 0` the result is 0 (due to integer arithmetic) or overflow if `y < -INT64_MAX` (assuming y is 64-bit)
* if `y > INT64_MAX / y` the divider `/ (y)) * (y) )` overflows (assuming y is 64-bit)

joachimmetz:
the fuzzer POC file (ignore the extension github is being picky about file extension that can be uploaded)
[clusterfuzz-testcase-minimized-fls_ext_fuzzer-4675556499390464.txt](https://github.com/sleuthkit/sleuthkit/files/7820600/clusterfuzz-testcase-minimized-fls_ext_fuzzer-4675556499390464.txt)

APriestman:
It's still unclear where the "fs_dir->fs_file->meta->size > INT64_MAX / fs_dir->fs_file->meta->size" part of the check is coming from. In the attached test file, the issue is that the (x + (y - 1)) part exceeds INT64_MAX (x(meta->size) = 0x7fffffffffffff00 and y (block_size) = 0x400). After this, we divide by y and then multiply by that same y, so I don't think that can cause issues as long as y is positive. So I believe we should throw an error if:
- fs_dir->fs_file->meta->size <= 0
- a_fs->block_size <= 0 (using 1 is pointless but not an error)
- INT64_MAX - (a_fs->block_size - 1) < fs_dir->fs_file->meta->size

joachimmetz:
> a_fs->block_size <= 0 (using 1 is pointless but not an error)

technically the extfs format defines block size in increments of 1024 (supported: 1024 2048 4096) anything else should be invalid

> It's still unclear where the "fs_dir->fs_file->meta->size > INT64_MAX / fs_dir->fs_file->meta->size" part of the check is coming from.

Don't recall at the moment, could have been an error on my side at the time. Feel free to adjust the changes where you see fit.

PR #2640, opened 2021-12-21
8258 create accounts with attributes

APriestman:


bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -248,6 +248,7 @@ SleuthkitCase getSleuthkitCase() {
 	 *                        address).
 	 * @param moduleName      The module creating the account.
 	 * @param sourceFile      The source file the account was found in.
+	 * @param attributes      List of blackboard attributes to add to the data artifact.
```
rcordovano: We might want to mention that this argument could be null or the empty list.

bindings/java/src/org/sleuthkit/datamodel/CommunicationsManager.java
```diff
@@ -522,17 +525,20 @@ private Account getOrCreateAccount(Account.Type accountType, String accountUniqu
...
 
 		BlackboardArtifact accountArtifact = getAccountFileInstanceArtifact(accountType, accountUniqueID, sourceFile);
 		if (accountArtifact == null) {
-			List<BlackboardAttribute> attributes = Arrays.asList(
-					new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ACCOUNT_TYPE, moduleName, accountType.getTypeName()),
-					new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ID, moduleName, accountUniqueID)
-			);
+			List<BlackboardAttribute> attributes = new ArrayList<>();
+			attributes.add(new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ACCOUNT_TYPE, moduleName, accountType.getTypeName()));
+			attributes.add(new BlackboardAttribute(BlackboardAttribute.ATTRIBUTE_TYPE.TSK_ID, moduleName, accountUniqueID));
```
rcordovano, line 538: It would be nice to replace the use of BlackboardAttribute.ATTRIBUTE_TYPE here with the appropriate BlackboardAttribute.Type constants.

rcordovano (changes requested):

rcordovano (approved):

PR #2682, opened 2022-02-07
CT-4992: Handle well known SIDs and accounts. 

raman-bt:


APriestman (approved):

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -115,40 +122,128 @@
...
 			.build();
 			
 	
 	// Looks for security identifier prefixes of the form S-<number>-<number>-<number>
-	//More information on security identifier architecture can be found at: 
+	// More information on security identifier architecture can be found at: 
 	// https://docs.microsoft.com/en-us/windows/security/identity-protection/access-control/security-identifiers
-	private static final Pattern WINDOWS_SPECIAL_ACCOUNT_REGEX = Pattern.compile("^\\s*S\\-\\d*\\-\\d*\\-(\\d*)");
+	// A number of accounts in the range S-1-5-80-* to S-1-5-111-* are special. 
+	private static final Pattern WINDOWS_SPECIAL_ACCOUNT_PREFIX_REGEX = Pattern.compile("^\\s*S\\-1\\-5\\-(\\d*)\\-");
```
crayy8, line 195: - Do we want this to be matching if the SID string has spaces in the front?
- This seems to only match if the S is capital. We should change it to [sS] to account for both cases

raman-bt: Addressed.

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -115,40 +122,128 @@
...
-			.put("S-1-5-96", "Font Driver Host Virtual Account")
+
+	// These SID prefixes indicate well known windows accounts.
+	//  - We can fill in the login names for these SID, as well as account user description.
+	private static final Map<String, WellKnownSidInfo> SPECIAL_SID_PREFIXES_MAP = ImmutableMap.<String, WellKnownSidInfo>builder() 
+			.put("S-1-5-80", new WellKnownSidInfo(false, "S-1-5-80", "NT Service", "", "NT Service Virtual Account"))
+			.put("S-1-5-82", new WellKnownSidInfo(false, "S-1-5-82", "IIS AppPool", "", "IIS AppPool Virtual Account"))
+			.put("S-1-5-83", new WellKnownSidInfo(false, "S-1-5-83", "NT Virtual Machine", "", "Virtual Machine Virtual Account") )
+			.put("S-1-5-90", new WellKnownSidInfo(false, "S-1-5-90", "Window Manager", "", "Windows Manager Virtual Account"))
+			.put("S-1-5-94", new WellKnownSidInfo(false, "S-1-5-94", "Windows Remoting Virtual Users", "", "Windows Remoting Virtual Account"))
```
crayy8, line 186: The realm should be "WinRM Virtual Users"

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -158,7 +253,7 @@ static boolean isWindowsSpecialSid(String sid) {
 			}
```
crayy8, line 253: The above logic where we look if the sid starts with a prefix seems redundant since the following step where we check if its s-1-5-80 - s-5-1-111 will catch these as well.

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -171,26 +266,173 @@ static boolean isWindowsSpecialSid(String sid) {
...
+		}
+		
+		for (Entry<String, WellKnownSidInfo> specialPrefixEntry : SPECIAL_SID_PREFIXES_MAP.entrySet()) {
+			if (tempSID.startsWith(specialPrefixEntry.getKey())) {
+				return specialPrefixEntry.getValue().getRealmAddr();
+			}
+		}
+
+		Matcher match = WINDOWS_SPECIAL_ACCOUNT_PREFIX_REGEX.matcher(tempSID);
+		if (match.find()) {
```
crayy8, line 309: Is the idea here that if we come across one of the special sids say s-1-5-100 that we currently don't have a realm for we will put in a real of NT AUTHORITY-100 as a placeholder? This would get updated later if the account ever came in with the real domain?

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -192,7 +192,7 @@ public String getDescription() {
 	// More information on security identifier architecture can be found at: 
 	// https://docs.microsoft.com/en-us/windows/security/identity-protection/access-control/security-identifiers
 	// A number of accounts in the range S-1-5-80-* to S-1-5-111-* are special. 
-	private static final Pattern WINDOWS_SPECIAL_ACCOUNT_PREFIX_REGEX = Pattern.compile("^\\s*S\\-1\\-5\\-(\\d*)\\-");
+	private static final Pattern WINDOWS_SPECIAL_ACCOUNT_PREFIX_REGEX = Pattern.compile("^[sS]\\-1\\-5\\-(\\d*)\\-");
```
crayy8, line 195: \\d* should be \\d+

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -171,26 +266,173 @@ static boolean isWindowsSpecialSid(String sid) {
...
+	static String getWindowsWellKnownSidRealmName(String sid) {
+		
+		if (StringUtils.isNotBlank(sid) && sid.equals(NTAUTHORITY_SID_PREFIX)) {
+			return NTAUTHORITY_REALM_NAME;
+		}
+		
+		WellKnownSidInfo wellKnownSidInfo = getWindowsWellKnownInfo(sid);
+		return Objects.nonNull(wellKnownSidInfo) 
+				? wellKnownSidInfo.getRealmName() 
+				: null;
```
crayy8, line 369: should this be returning empty string per the functions comments?

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -115,40 +122,128 @@
...
-			.put("S-1-5-80", "Service Virtual Account")
-			.put("S-1-5-82", "IIS AppPool Virtual Account")
-			.put("S-1-5-83", "Virtual Machine Virtual Account")
-			.put("S-1-5-90", "Window Manager Virtual Account")
-			.put("S-1-5-94", "WinRM Virtual accountt")
-			.put("S-1-5-96", "Font Driver Host Virtual Account")
+
+	// These SID prefixes indicate well known windows accounts.
+	//  - We can fill in the login names for these SID, as well as account user description.
+	private static final Map<String, WellKnownSidInfo> SPECIAL_SID_PREFIXES_MAP = ImmutableMap.<String, WellKnownSidInfo>builder()
```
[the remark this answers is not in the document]
raman-bt: Yes.  The name comparison is case insensitive. 

I have seen these as display names both in camel case and upper case. 
https://docs.microsoft.com/en-us/windows/security/identity-protection/access-control/security-identifiers

These are what we display to in the UI.  If you prefer that all caps be displayed, then we can change them.

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
raman-bt: changed.

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
raman-bt: We are returning the realm address (SID) here.  So we will return "S-1-5-100" in this case.

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
```diff
@@ -171,26 +266,173 @@ static boolean isWindowsSpecialSid(String sid) {
...
+	 * @param sid SID to check.
+	 * 
+	 * @return Login Name for Windows special SID, an empty string if the SID is not a known special SID. 
+	 */
+	static String getWindowsWellKnownSidLoginName(String sid) {
+		
+		WellKnownSidInfo wellKnownSidInfo = getWindowsWellKnownInfo(sid);
+		return Objects.nonNull(wellKnownSidInfo) 
+				? wellKnownSidInfo.getLoginName()
+				: null;
```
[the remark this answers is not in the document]
raman-bt: Fixed comment.

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
raman-bt: fixed comment.

bindings/java/src/org/sleuthkit/datamodel/WindowsAccountUtils.java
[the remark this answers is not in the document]
raman-bt: I've plugged in these names exactly - first 3 in uppercase and last three in mixed case.

crayy8 (approved):

PR #2705, opened 2022-04-18
CT5409 icat support for logical directories

APriestman:


APriestman:
Replaced by https://github.com/sleuthkit/sleuthkit/pull/2706

PR #2709, opened 2022-05-06
Updating sqlite-jdbc and postgresql. Vulnerability mitigation

jayaramcs:
Postgresql : [CVE-2022-21724](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-21724)


APriestman:
This is going to break Autopsy - we at least need to change the lines here: https://github.com/sleuthkit/autopsy/blob/671f5a3961a8a441f42914e7163ce42addecaca6/Core/build.xml#L137
I can't test a fix because I now have the wrong version of Java.

APriestman:
Autopsy PR https://github.com/sleuthkit/autopsy/pull/7622

PR #2713, opened 2022-05-17
CT5429 Final logical directory fixes

APriestman:
Remove timestamps.
Limit cache file length.
Add default name if conversion fails.


tsk/fs/logical_fs.cpp
```diff
@@ -190,7 +193,11 @@ convert_wide_string_to_utf8(const wchar_t *source) {
 			&utf8[maxUTF8len], TSKlenientConversion);
 
 	if (retVal != TSKconversionOK) {
-		return NULL; // TODO - revisit failed conversion handling
+		// If the conversion failed, use a default name
+		if (tsk_verbose)
+			tsk_fprintf(stderr,
+				"convert_wide_string_to_utf8: error converting logical file name to UTF-8\n");
+		strncpy(dest, invalidName, strlen(invalidName));
```
crayy8, line 200: Should we use strncpy_s?

Is it possible that dest will be smaller than invalidName len? Ie. if we fail on a single character name?

APriestman: char *dest = (char*)tsk_malloc(maxUTF8len);

crayy8 (approved):

PR #2717, opened 2022-06-07
Fix string format specifier warnings

uckelman-sf:
* Use the correct string format specifiers in calls to TFPRINTF.
* Use standards-compliant %ls instead of legacy MSVC %S.

### DIFF
--- a/tests/fs_thread_test.cpp
+++ b/tests/fs_thread_test.cpp
@@ -154,7 +154,7 @@ static const TSK_TCHAR *progname;
 static void
 usage()
 {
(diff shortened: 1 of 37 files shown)


bcarrier:
Hi @uckelman-sf .  I'm getting a bunch of compiler issues when I build this branch.  I think it's because all of the calls from %s were changed to " PRIttocTSK " and they need to instead have an explicit '%' in them. I.e. '%"PRIttocTSK"'.  

Otherwise, I think it looks OK. I had to read this post for context: https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823

I had never seen warnings/errors from using the Visual-studio style formatting.

kellykelly3 (approved):

PR #2729, opened 2022-07-22
Fixes heap use after free read

sashashura:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36024 and
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36098 and
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36021 and
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=36122 and
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=38365

The `val_data` points to invalid memory because of invalid `t.val_offset`.


sashashura:
@APriestman I saw you have merged some of my pull requests but not others. Do you think this fix is no good? Please let me know.

kellykelly3 (approved):

PR #2730, opened 2022-07-22
Fixes Heap-buffer-overflow read in inode_walk_file_act

sashashura:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=35648

The root cause is in call to in `(size_t) ((a_fs->block_count + 7) / 8))` `fatfs_inode_walk`. When `a_fs->block_count` is max 64bit int it leads to integer overflow and allocation of a small buffer.

### DIFF
--- a/tsk/fs/fatfs_meta.c
+++ b/tsk/fs/fatfs_meta.c
@@ -1224,8 +1224,7 @@ fatfs_inode_walk(TSK_FS_INFO *a_fs, TSK_INUM_T a_start_inum,
     /* Allocate a bitmap to keep track of which sectors are allocated to
      * directories. */
     if ((dir_sectors_bitmap =
-            (uint8_t*)tsk_malloc((size_t) ((a_fs->block_count +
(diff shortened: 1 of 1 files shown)


sashashura:
More explanation: when `a_fs->block_count` is max 64 bit uint value or close to it, `a_fs->block_count + 7` overflows (wraps around) and becomes for instance `6`. It breaks the calculation as the result of `(size_t) ((a_fs->block_count + 7) / 8))` becomes `0`.

As I understand `+ 7) / 8` is done to account for possible reminder from division by `8`. Say `block_count` is 15. Then, adding extra 7 before the division gives `2 == (15 + 7) / 8)` instead of 1.

The fix is to avoid the overflow: divide the `block_count` first and add one if needed.

sashashura:
Is anything unclear with this one? All others have been merged.

bcarrier:
Thanks.  This fix was also addressed by another PR that added a 128MB limit on alloc size.

PR #2732, opened 2022-07-22
Fix sleuthkit build on Linux

sashashura:
The `shouldTreatAsDirectory` function uses WIN32 specific definitions.


sashashura:
@APriestman?

sashashura:
Fixed with another commit

PR #2737, opened 2022-08-04
DO NOT MERGE Save folder names and sizes when realloc fails

APriestman:


simsong:
@APriestman  - Should we close this?

Issue #1148, opened 2018-02-27
LVM Decoding


crssantos:
I realized that sleuthkit doesn´t recognize Linux Logical Volume Manager (LVM) partitions. 

How can I solve this problem?

Thanks in advance.

gilsonm112:
Linux Volume Manager and I have stopped using Sleuthkit because it does not support them, it would be nice to include this support

lfcnassif:
Up

igheid:
Up

Issue #2748, opened 2022-08-27
Where is the pool layer documentation?


joachimmetz:
Where is the TSK pool layer documented?

http://wiki.sleuthkit.org/index.php?title=TSK_Tool_Overview does not mention the concept of pools

A third party paper seems to mention some details https://www.researchgate.net/publication/318925723_Extending_The_Sleuth_Kit_and_its_underlying_model_for_pooled_storage_file_system_forensic_analysis

joachimmetz:
Questions to be answered:

* what is the role of the CompatPool versus the "regular" Pool object?
* Why is GUID a member of the base Pool object? https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/pool/tsk_pool.hpp#L41. What if the identifier of a pool is not a GUID/UUID?

PR #2751, opened 2022-08-28
Added Linux LVM support using libbfio and libvslvm #1148

joachimmetz:


joachimmetz:
@bcarrier PTAL per separate mail conversion. I'll need to some support to get this cross-platform and inline with design ideas behind the pool layer seeing there is very little developers information that I was able to find.

joachimmetz:
@bcarrier any help from the maintainers side? The documentation for the pool layer is mostly missing https://github.com/sleuthkit/sleuthkit/issues/2748

tsk/pool/pool_open.cpp
```diff
@@ -3,16 +3,19 @@
  *
  * Brian Carrier [carrier <at> sleuthkit [dot] org]
  * Copyright (c) 2018-2019 BlackBag Technologies.  All Rights reserved
+ * Copyright (c) 2022 Joachim Metz <joachim.metz@gmail.com>
  *
  * This software is distributed under the Common Public License 1.0
  */
-#include "../fs/tsk_apfs.hpp"
+#include "tsk/tsk_config.h"
```
APriestman, line 10: This breaks the Visual Studio build since the file is not found.

joachimmetz: what is not found? can't this be addressed with an additional include path in the VS configuration?

APriestman: tsk_config.h does not exist.

joachimmetz: ah so it is not generated for VS, any conditional imports I should use? what is the recommended way of dealing with this in this codebase?

APriestman: You could try importing tsk_base_i.h instead - it has a check on HAVE_CONFIG_H before trying to include it.

joachimmetz: I'll have a look at that (tomorrow local time), the VS configuration has to define HAVE_LIBBFIO and HAVE_LIBVSLVM and ideally LIBVSLVM_HAVE_BFIO as well

joachimmetz: Made changes to headers

APriestman (approved):

typeryougishiki:
Can I use this supply to get fs info from lvm volume？

markmckinnon:
As long as you are willing to build sleuthkit from source on Linux and make the proper libraries available you can see that information

leean518:
Does LVM support not work when developing on the windows operating systems?

leean518:
@joachimmetz I believe I did. But just to confirm, can you specify how to compile with LVM support on Windows?

joachimmetz:
> But just to confirm, can you specify how to compile with LVM support on Windows?

do you see DLLs for libvslvm and libbfio ? if not you likely did not compile with LVM support. I don't maintain the Visual Studio configuration of this project, so I don't know.

leean518:
> 

I had to add the libraries myself as they are not included. The main problem I am having is how to integrate it with the sleuth kit so that the library is able to detect where it is.

Issue #2495, opened 2021-07-22
`--enable-static` links tools statically only to libtsk


uckelman-sf:
Do this:
```
$ ./configure --disable-shared --enable-static
$ make
```
Check the linkage of some tool in `tools/`:
```
$ ldd img_cat
	linux-vdso.so.1 (0x00007fff095fb000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007fe03f28d000)
	libsqlite3.so.0 => /lib64/libsqlite3.so.0 (0x00007fe03f156000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007fe03f14f000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fe03f12d000)
	libz.so.1 => /lib64/libz.so.1 (0x00007fe03f113000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fe03efce000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007fe03efb1000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fe03ede7000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fe03f4bd000)
```
We've successfully statically linked with libtsk... but not with anything else. 

These are forensics tools, for which a common use case is running them on a machine where we have to bring everything with us; it would be better if the tools could be entirely statically linked, not just statically linked to libtsk.

uckelman-sf:
I'll open a PR adding an option for full static linking of the tools soon.

simsong:
...  by your wonderful patch!

uckelman-sf:
I do, but I need to pull in the test we've been using from elsewhere, so it won't be today.

simsong:
That's fine.

uckelman-sf:
Working on static link checking in PR #2968.

uckelman-sf:
This was completed in #2968. Closing.

PR #2755, opened 2022-09-08
Statically link tools for static builds

uckelman-sf:
Tools should also be statically linked when doing static builds. Fixes Issue #2495.


uckelman-sf:
@simsong Ready for review.

simsong (approved):
Thanks for the help on the configure.ac file.  It's slowly being modernized!

PR #2756, opened 2022-09-08
Don't add /usr/local/* to CFLAGS or LDFLAGS.

uckelman-sf:
Don't add /usr/local/* to CFLAGS or LDFLAGS. It's surprising,
nonstandard behavior and leads to wasted troubleshooting time. If the user wants that, the user should set it when invoking configure.


uckelman-sf:
Done as part of PR #2942.

PR #2770, opened 2022-10-21
Added a unit test for account update/merge scenario.

raman-bt:


raman-bt:
@bcarrier or @APriestman 

Could one of you please merge this into TSK>  It's an update to the Unit test for realm/account merge.

PR #2774, opened 2022-10-25
Fix bound check on key_data

APriestman:


bcarrier:
Fix to bug introduced via #2729

PR #2781, opened 2022-11-18
openssl support and minor crypto.hpp correction

arisjr:
This correction enables again libssl and the APFS patch, correcting its build on the project.

It was made a small adjustment to add openssl support and an equally small, but important, correction on crypto.hpp to enable the correct linkage/build of TSK binaries.


arisjr:
Just to complement this pull request, I have to say it was tested using libssl1.1.

PR #2787, opened 2022-12-22
6356 Bail out of processing very large folders

APriestman:


sf-jonstewart:
This is concerning. The Sleuthkit can't just bail out of processing directories with a million files. This is forensics. The point is to do so in a way that's not accidentally quadratic.

sf-jonstewart:
Additionally, extreme slowness happens at thresholds orders of magnitude below 1M files. This is not at all a fix.

bcarrier:
> This is not at all a fix.

It's not the final fix on this topic, but it certainly creates better behavior than without it. It now gives an error message that the situation has occurred versus exhausting memory and taking an application down with it. 

For better or worse, there is an inherent assumption in TSK that all file names in a folder can be kept in memory.  We do this so that we can de-dupe and not show lots of duplicate delete files names. 

It's a substantial change to both de-dupe and not keep everything in memory.

uckelman-sf:
How is having to recompile in order to process an image _better behavior_?

lfcnassif:
This affected our users recently. If we send a PR making that hard coded limit configurable using env vars, will it be reviewed?

jonstewart:
Hi Luis,

Maybe? I am going to review this in next couple days and see whether it's possible to eliminate the quadratic behavior entirely — it would still require memory, but most people have enough memory to dedupe out a million filenames given the right data structures.

PR #2790, opened 2023-01-06
6415 Add support for long file names

APriestman:


tsk/fs/logical_fs.cpp
```diff
@@ -274,19 +274,57 @@ populate_fs_file_from_win_find_data(const WIN32_FIND_DATA* fd, TSK_FS_FILE * a_f
...
+*
+* @param base_path The path to the directory to open
+*
+* @return The search path with wildcard appended (must be freed by caller)
+*/
+TSK_TCHAR * create_search_path_long_path(const TSK_TCHAR *base_path) {
+#ifdef TSK_WIN32
+
+	// First convert the base path to an absolute path
+	TCHAR absPath[LOGICAL_MAX_PATH_UNICODE];
```
crayy8, line 307: do we need to add +1 here for the null byte?

APriestman: I don't believe so? https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamew

tsk/fs/logical_fs.cpp
```diff
@@ -1011,6 +1081,7 @@ logicalfs_dir_open_meta(TSK_FS_INFO *a_fs, TSK_FS_DIR ** a_fs_dir,
 	vector<wstring> file_names;
 	vector<wstring> dir_names;
 	if (TSK_OK != load_dir_and_file_lists_win(path, file_names, dir_names, LOGICALFS_LOAD_ALL)) {
+		fprintf(stderr, "Error in load_dir_and_file_lists_win\n");
```
bcarrier, line 1084: Should this go in the error string vs printf?

APriestman: Oops I think that was a debug thing. The error should already be recorded

PR #2799, opened 2023-02-09
Fire event when os accounts are merged

msilva-basis:


bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -959,6 +959,7 @@ private void mergeOsAccounts(OsAccount sourceAccount, OsAccount destAccount, Cas
 
 			
 			// TBD: We need to emit another event which tells CT that two accounts are being merged so it can updates other dedicated tables
```
bcarrier, line 961: This comment should change.

msilva-basis: The comment has now been updated.

raman-bt (approved):
Looks good to me.

bindings/java/src/org/sleuthkit/datamodel/TskEvent.java
```diff
@@ -306,6 +306,52 @@ public List<Long> getOsAccountObjectIds() {
...
+		 * 
+		 * @return 
+		 */
+		public List<MergedAccountsPair> getMergedAccountPairs() {
+			return getDataModelObjects();
+		}
+
+	}
+
+	public final static class MergedAccountsPair {
```
jayaramcs, line 336: we should add comments to this class and its members

msilva-basis: I have now added the comments.

Issue #2802, opened 2023-02-11
Heap-buffer-overflow in APFSJObject::add_entry


joachimmetz:
Reported by OSSFuzz in https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52764

joachimmetz:
https://osv.dev/vulnerability/OSV-2022-1106

joachimmetz:
For assessment of impacted SleuthKit versions https://github.com/sleuthkit/sleuthkit/blame/develop/tsk/fs/apfs_fs.cpp#L146 seems to date back to version 4.8.0 (https://github.com/sleuthkit/sleuthkit/commit/edb2e46a2cbcfdc97ec88c47225e9c024096cea1)

joachimmetz:
```
==482633==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x621000007544 at pc 0x7f87dc049e0b bp 0x7fffdb60a400 sp 0x7fffdb609bb0
READ of size 3604 at 0x621000007544 thread T0
    #0 0x7f87dc049e0a in __interceptor_memcpy (/lib64/libasan.so.8+0x49e0a)
    #1 0x417bfd in std::char_traits<char>::copy(char*, char const*, unsigned long) /usr/include/c++/12/bits/char_traits.h:431
    #2 0x417bfd in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_S_copy(char*, char const*, unsigned long) /usr/include/c++/12/bits/basic_string.h:423
    #3 0x417bfd in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_S_copy(char*, char const*, unsigned long) /usr/include/c++/12/bits/basic_string.h:418
    #4 0x417bfd in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_S_copy_chars(char*, char const*, char const*) /usr/include/c++/12/bits/basic_string.h:477
    #5 0x417bfd in void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*>(char const*, char const*, std::forward_iterator_tag) /usr/include/c++/12/bits/basic_string.tcc:243
[...]

PR #2803, opened 2023-02-11
Fix for OOB read in APFSJObject::add_entry #2802

joachimmetz:


joachimmetz:
@bcarrier @rcordovano PTAL

lfcnassif:
Hi @joachimmetz. I'm testing this on some APFS images. For 2 I've tested until now, it seems to freeze FS transversal at the very beginning, not sure if it is forever or if it became very slow at some point, I canceled the decoding after some dozens of minutes.

joachimmetz:
any test image you can share? or test case to reproduce as part of https://github.com/dfirlabs/apfs-specimens? The thing is there are no unit tests for TSK and I use its APFS functionality VERY SPORADIC.

joachimmetz:
one option is to use `fls -v` (with and without patches) and see where it errors and then use fsapfsinfo with [debug info](https://github.com/libyal/libfsapfs/wiki/Troubleshooting#verbose-and-debug-output) to see what data structures TSK might no longer be fully parsing.

> I'm not experienced with C code debugging,

this highly depends on the coding style as well, TSK APFS is C++ heavy with templating, so hard to debug (and maintain)

simsong:
Can you rebase this with the current develop and trigger the self-tests?

joachimmetz:
rebased with develop branch

joachimmetz:
@simsong PTAL if/how these changes can provide better error condition feedback to the user

Issue #2700, opened 2022-04-12
When APFS volume/file system fails to open, the next volumes should be tried instead of aborting


lfcnassif:
This is another regression we detected when upgrading from our old fork to TSK-4.11.0 (https://github.com/sepinf-inc/IPED/pull/771)

We have some decrypted APFS disk images to process. Although they are decrypted, opening a first volume fails and TSK aborts with an "Encrypted APFS file system" message. But later volumes are decrypted and should be processed. Removing this block makes later decrypted volumes to be processed fine:

https://github.com/sleuthkit/sleuthkit/blob/sleuthkit-4.11.1/tsk/auto/auto.cpp#L489-L491

I'll submit a PR shortly targeting develop branch.

lfcnassif:
Could someone reopen this (I don't have permission), since the solution was rolled back by #2820 (see the discussion there).

PR #2820, opened 2023-03-25
Numerous LVM support improvements

joachimmetz:
Includes numerous changes contributed by arisjr@

### DIFF
--- a/bindings/java/jni/auto_db_java.cpp
+++ b/bindings/java/jni/auto_db_java.cpp
@@ -307,7 +307,7 @@ TskAutoDbJava::addPoolInfoAndVS(const TSK_POOL_INFO *pool_info, int64_t parObjId
         m_savedVsInfo.push_back(vs_db);
         saveObjectInfo(objId, poolObjId, TSK_DB_OBJECT_TYPE_VS);
     } 
-    if (pool_info->ctype == TSK_POOL_TYPE_LVM){
+    else if (pool_info->ctype == TSK_POOL_TYPE_LVM){
         // Add the APFS pool volume
         jlong objIdj = m_jniEnv->CallLongMethod(m_javaDbObj, m_addVolumeSystemMethodID,
             poolObjIdj, TSK_VS_TYPE_LVM, pool_info->img_offset, (uint64_t)pool_info->block_size);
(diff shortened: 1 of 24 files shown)


simsong:
Hi @joachimmetz  - do you have test media that validates these patches?

joachimmetz:
There are several mentioned in https://github.com/sleuthkit/autopsy/issues/7888 and you can create some with https://github.com/dfirlabs/lvm-specimens

simsong:
> https://github.com/dfirlabs/lvm-specimens

I'm collecting the test media at digitalcorpora.org; if you think that there is a better place to put them, please let me know. I'm hesitant to use github because they charge for I/O downloads from their large object support.

lfcnassif:
Hi @arisjr and @joachimmetz. Any specific reason why commit [2fa3241](https://github.com/sleuthkit/sleuthkit/pull/2820/commits/2fa3241ec99af7b7f41daeab545468e8b1cf08bb) reverted the change of #2702 or was it by accident???

joachimmetz:
@lfcnassif what is https://github.com/sleuthkit/sleuthkit/pull/2702 ? First time I hear of it. How is a change in the common code supposed to address an APFS specific issue https://github.com/sleuthkit/sleuthkit/issues/2700 ?

My guess it was not intentional "reverted" but overlooked in merge/review.

joachimmetz:
> I'm collecting the test media at digitalcorpora.org; if you think that there is a better place to put them, please let me know. I'm hesitant to use github because they charge for I/O downloads from their large object support.

@simsong the goal of lvm-specimens is make reproducible test images with different versions of LVM implementations (broad set of format cases and features), not to provide a collection of (narrowly scoped) pre-generated test images.

simsong:
That's great. I'm fine if we can make reproducible test images without having to archive the bits.

lfcnassif:
> @lfcnassif what is #2702 ? First time I hear of it. How is a change in the common code supposed to address an APFS specific issue #2700 ?

Hi @joachimmetz. Maybe it is not specific to APFS, but it showed up in a real case APFS image. If TSK fails to decode a specific volume, shouldn't it try to decode the next volumes in the same image instead of aborting?

lfcnassif:
> I'm talking to @arisjr to know if it was intentional or not.

@arisjr confirmed it wasn't intentional.

PR #2821, opened 2023-04-04
6807 store content collected status

APriestman:


bindings/java/src/org/sleuthkit/datamodel/AbstractFile.java
```diff
@@ -97,6 +97,9 @@ public abstract class AbstractFile extends AbstractContent {
 	protected String sha1Hash;
 	private boolean sha1HashDirty = false;
 	
+	private TskData.CollectedStatus collected; // Collected status of file data
```
jayaramcs, line 100: should we set this to volatile? we could have one thread setting up the collected status and another invoking save?

eugene7646 (approved):

PR #2832, opened 2023-06-02
7074 content provider

gdicristofaro:


bindings/java/src/org/sleuthkit/datamodel/SleuthkitCase.java
```diff
@@ -406,6 +413,17 @@ private void init() throws Exception {
...
 	}
+	
+	/**
+	 * Returns the custom content provider for this case if one exists.
+	 * Otherwise, returns null.
+	 *
+	 * @return The custom content provider for this case if one exists.
+	 *         Otherwise, returns null.
+	 */
+	ContentProvider getContentProvider() {
```
bcarrier, line 424: Should this return an Optional instead of NULL?

gdicristofaro: @bcarrier , sure.  I will change it to return `Optional<ContentProvider>`

PR #2834, opened 2023-06-21
fix #2800: return unsigned byte by InputStream API (-1 just if EOF)

### LINKED ISSUE #2800: ReadContentInputStream.read() method returning wrong values
The `InputStream.read()` contract says it should return a byte as an unsigned int in [0, 255] range when EOF is not reached. The current implementation is:
https://github.com/sleuthkit/sleuthkit/blob/sleuthkit-4.12.0/bindings/java/src/org/sleuthkit/datamodel/ReadContentInputStream.java#L41-L42

Last line should be fixed to:
`return (read(buff) != -1) ? buff[0] & 0xFF : -1;`

lfcnassif:
Fix #2800. Returns unsigned byte according to method contract, avoiding to return -1 before EOF.

### DIFF
--- a/bindings/java/src/org/sleuthkit/datamodel/ReadContentInputStream.java
+++ b/bindings/java/src/org/sleuthkit/datamodel/ReadContentInputStream.java
@@ -39,7 +39,7 @@ public final class ReadContentInputStream extends InputStream {
 	@Override
 	public int read() throws ReadContentInputStreamException {
 		byte[] buff = new byte[1];
-		return (read(buff) != -1) ? buff[0] : -1;
+		return (read(buff) != -1) ? buff[0] & 0xFF : -1;
 	}
 
 	@Override


lfcnassif:
Could someone review this? All modules calling SleuthkitInputStream.read() method can be affected and stop processing files before EOF.

markmckinnon:
Do you have a sample dataset that causes the error that you can provide to expedite testing?

lfcnassif:
Fix is just an one line change.

lfcnassif:
> Do you have a sample dataset that causes the error that you can provide to expedite testing?

We were affected by this on our project because some modules use the affected method. I don't know if TSK or Autopsy use this public method, but if not it may be used in the future.

To reproduce the error just create a file with a 0xFF as starting byte and more random bytes. Try to read that whole file using just the mentioned method, checking if the returned value is EOF (-1) after each read. The EOF (-1) will be returned in the first read, although the file is not 0 sized.

markmckinnon (approved):

simsong (approved):

simsong:
let's see if closing and opening the PR triggers the action...

simsong:
It did!

simsong:
@lfcnassif  — Hi. I'm eager to take this code, but the Java code is not compiled as part of our test cases. And we don't have a test that demonstrates this error. Can you help out with either?

simsong (approved):
Okay, I'm approving this.

lfcnassif:
Sorry @simsong for not creating an unit test, AFAIK it was not required by this project rules when I submited this. I would like to help and create one, unfortunately these months I'm not having time to work even on my own projects... The step by step to create a test file and test code are outlined above. Anyway, thank you for merging this PR.

simsong:
I looked at the code and figured it was fine. Thank you for the fix.

PR #2835, opened 2023-06-27
configure.ac: use '=' for comparison instead of '=='

gktrk:
The operator '==' isn't POSIX compliant[0]. Use the standard '=', as it's done everywhere else in configure.ac.

[0] https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html

Bug: https://bugs.gentoo.org/870250


markmckinnon (approved):

simsong (approved):
That's right. Thanks!

PR #2844, opened 2023-07-11
CT-7152

ryandeisma:
Changes to DB Diff to read changes in CT tables

### DIFF
--- a/db_diff/tskdbdiff.py
+++ b/db_diff/tskdbdiff.py
@@ -1027,20 +1027,9 @@ def normalize_tsk_objects_path(guid_util: TskGuidUtils, objid: int,
                             path_parts[idx + 2].lower().endswith("report.html"):
                         path_parts = ["Reports", "html-report.html"]
                         break
-            for idx in range(0, len(path_parts) - 1):
-                if path_parts[idx].lower() == "reports" and \
-                        path_parts[idx + 1].lower().startswith("autopsytestcase html report"):
-                    path_parts = ["Reports", "AutopsyTestCase HTML Report"]
-                    break
-                if path_parts[idx].lower() == "reports" and \
-                        "html report" in path_parts[idx + 1].lower() and \
-                        len(path_parts) > idx + 2 and \
-                        path_parts[idx + 2].lower().endswith("report.html"):
-                    path_parts = ["Reports", "html-report.html"]
-                    break
 
 
-        path = os.path.join(*path_parts) if (path_parts != None and len(path_parts)) > 0 else '/'
+        path = os.path.join(*path_parts) if (path_parts is not None and len(path_parts) > 0) else '/'
 
         return path
 
--- a/db_diff/tskdbdiff.py
+++ b/db_diff/tskdbdiff.py
@@ -272,6 +272,7 @@ class TskDbDiff(object):
                             msg = "There were inconsistent sources for artifact with id #" + str(row["artifact_id"]) + ".\n"
 
                         try:
+                            attr_value_as_string = None
                             if attr["value_type"] == 0:
                                 attr_value_as_string = str(attr["value_text"])                        
                             elif attr["value_type"] == 1:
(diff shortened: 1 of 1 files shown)


db_diff/tskdbdiff.py
```diff
@@ -285,10 +286,16 @@ def _dump_output_db_bb(db_file, bb_dump_file, isMultiUser, pgSettings, id_obj_pa
...
-                                attr_value_as_string = str(attr["value_int64"])                        
+                                attr_value_as_string = str(attr["value_int64"])
+                            elif attr["value_type"] == 6:
+                                attr_value_as_string = str(attr["value_text"])
                             if attr["display_name"] == "Associated Artifact":
                                 attr_value_as_string = TskDbDiff._get_associated_artifact_type(attribute_cursor, attr_value_as_string, isMultiUser)
                             patrn = re.compile("[\n\0\a\b\r\f]")
+                            if attr_value_as_string == None:
+                                print(f'{attr["value_type"]}')
+                                print(f'{attr["display_name"]}')
```
gdicristofaro, line 297: I think these are good checks to have, but could it be a little more descriptive of the problem in case we need to debug this later (i.e. `print(f'Could not determine attribute value for value type: {attr["value_type"]}, display name: {attr["display_name"]}')` )?

db_diff/tskdbdiff.py
```diff
@@ -285,10 +286,16 @@ def _dump_output_db_bb(db_file, bb_dump_file, isMultiUser, pgSettings, id_obj_pa
...
                                 attr_value_as_string = "bytes"                        
                             elif attr["value_type"] == 5:
-                                attr_value_as_string = str(attr["value_int64"])                        
+                                attr_value_as_string = str(attr["value_int64"])
+                            elif attr["value_type"] == 6:
+                                attr_value_as_string = str(attr["value_text"])
                             if attr["display_name"] == "Associated Artifact":
                                 attr_value_as_string = TskDbDiff._get_associated_artifact_type(attribute_cursor, attr_value_as_string, isMultiUser)
                             patrn = re.compile("[\n\0\a\b\r\f]")
+                            if attr_value_as_string == None:
```
gdicristofaro, line 295: I doubt it matters too much, but I think `is None` and `is not None` is the preferred way to check for `None` based on [this](https://stackoverflow.com/a/3965129) and [this](https://stackoverflow.com/a/2020606).

db_diff/tskdbdiff.py
```diff
@@ -652,6 +659,9 @@ def index_of(lst, search_item) -> int:
     Returns: The index in the list of the item or -1.
 
     """
+    if lst == None:
```
gdicristofaro, line 662: I think `if lst is None:` would be a better check.

db_diff/tskdbdiff.py
```diff
@@ -999,28 +1009,20 @@ def normalize_tsk_objects_path(guid_util: TskGuidUtils, objid: int,
                 path_parts = path_parts[:-1]
 
         if path_parts and len(path_parts) >= 2:
-            is_leapp = False
```
gdicristofaro, line 1002: I think we need the leapp path sanitization for some of our current testing.  It removes some timestamp path segments which change daily.

db_diff/tskdbdiff.py
```diff
@@ -999,28 +1009,20 @@ def normalize_tsk_objects_path(guid_util: TskGuidUtils, objid: int,
...
-                        break
-                    if path_parts[idx].lower() == "reports" and \
-                            "html report" in path_parts[idx + 1].lower() and \
-                            len(path_parts) > idx + 2 and \
-                            path_parts[idx + 2].lower().endswith("report.html"):
-                        path_parts = ["Reports", "html-report.html"]
-                        break
 
-        path = os.path.join(*path_parts) if len(path_parts) > 0 else '/'
+        path = os.path.join(*path_parts) if (path_parts != None and len(path_parts)) > 0 else '/'
```
gdicristofaro, line 1025: I think the parenthesis might be off here.  Maybe something like: `path = os.path.join(*path_parts) if (path_parts is not None and len(path_parts) > 0) else '/'`?

gdicristofaro (changes requested):

db_diff/tskdbdiff.py
```diff
@@ -1019,8 +1027,20 @@ def normalize_tsk_objects_path(guid_util: TskGuidUtils, objid: int,
...
+                if path_parts[idx].lower() == "reports" and \
+                        path_parts[idx + 1].lower().startswith("autopsytestcase html report"):
+                    path_parts = ["Reports", "AutopsyTestCase HTML Report"]
+                    break
+                if path_parts[idx].lower() == "reports" and \
+                        "html report" in path_parts[idx + 1].lower() and \
+                        len(path_parts) > idx + 2 and \
+                        path_parts[idx + 2].lower().endswith("report.html"):
+                    path_parts = ["Reports", "html-report.html"]
+                    break
```
gdicristofaro, line 1040: I think you can take this out since it already happens if it isn't a leapp report.

db_diff/tskdbdiff.py
gdicristofaro: I think you may have missed this.

gdicristofaro (changes requested):

gdicristofaro (approved):

Issue #2735, opened 2022-08-04
NTFS: non-existing file (moved or deleted) is shown as regular allocated file


0x4d4c:
In the raw image attached (`fs.img` in the zip file), the file `xazjlgis/ldniruft/yqntnnnb/lkrabebi/lykvvzey` has been renamed to `xazjlgis/ldniruft/xyalaefy/cfipubey/blpytzyw` (sorry for the file names).  However, `lykvvzey` is still reported as allocated by `fls`.
```
$ fls -r fs.img | grep -A5 -B5 lykvvzey
++++++ d/d 97-144-2:	tfkzejer
+++++++ d/d 94-144-2:	idilfoyx
+++++ r/r 78-128-2:	kyiujdrs
+++++ r/r 112-128-2:	lbcexece
+++ d/d 71-144-2:	lkrabebi
++++ r/r 83-128-2:	lykvvzey        <---- file is presented as regular allocated file
+++ r/r 111-128-2:	thkzlxxq
+ d/d 86-144-2:	ovigoofn
++ r/r 118-128-2:	bffpmbjg
++ d/d 87-144-2:	mjmlvebp
+++ r/r 95-128-2:	goezwkxo
```
Calling `istat` with the metadata address `83-128` shows the correct new file name:
```
$ istat fs.img 83-128
MFT Entry Header Values:
Entry: 83        Sequence: 1
$LogFile Sequence Number: 0
Allocated File
Links: 1

$STANDARD_INFORMATION Attribute Values:
Flags: Archive
Owner ID: 0
Security ID: 0  ()
Created:	2022-08-02 17:13:19.902670700 (CEST)
File Modified:	2022-08-02 17:13:20.465722800 (CEST)
MFT Modified:	2022-08-02 17:15:20.386325300 (CEST)
Accessed:	2022-08-02 17:13:38.785556900 (CEST)

$FILE_NAME Attribute Values:
Flags: Archive
Name: blpytzyw                              <---- this is the correct new file name
Parent MFT Entry: 77 	Sequence: 1
Allocated Size: 208896   	Actual Size: 206188
Created:	2022-08-02 17:13:19.902670700 (CEST)
File Modified:	2022-08-02 17:13:20.465722800 (CEST)
MFT Modified:	2022-08-02 17:13:20.465722800 (CEST)
Accessed:	2022-08-02 17:13:38.785556900 (CEST)

Attributes: 
[...]

PR #2851, opened 2023-07-26
fix NTFS index record processing

NBojic:
Proposed fix for NTFS  index entry processing in function "ntfs_dir_open_meta" that solves issue #2735.

- Function "ntfs_dir_open_meta" now checks for each index record whether it is active or not by reading the bitmap attribute. "ntfs_dir_open_meta" scans a big buffer from start to end in a for-loop and looks for index records. The fix uses the current parsing offset from the big for-loop to find out which bit of the bitmap is associated with the index record. 
- "ntfs_dir_open_meta" calls function "ntfs_proc_idxentry" to process an index entry. "ntfs_proc_idxentry" takes argument "a_is_del". The fix sets "a_is_del" to 1 if the current index entry corresponds to an inactive index record.

### DIFF
--- a/tsk/fs/ntfs_dent.cpp
+++ b/tsk/fs/ntfs_dent.cpp
@@ -367,7 +367,7 @@ is_time(uint64_t t)
 /**
  * Process a lsit of index entries and add to FS_DIR
  *
- * @param a_is_del Set to 1 if these entries are for a deleted directory
+ * @param a_is_del Set to 1 if these entries are for a deleted directory or if they correspond to index record that is unallocated according to bitmap attribute
  * @param idxe Buffer with index entries to process
  * @param idxe_len Length of idxe buffer (in bytes)
  * @param used_len Length of data as reported by idexlist header (everything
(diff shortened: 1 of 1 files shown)


simsong:
@NBojic — do you have a test media that demonstrates this bug? How did you find it?
I'd like to accept it, but I am trying to implement a more sophisticated testing regime for TSK that will help us validate patches such as this one.

0x4d4c:
@simsong, there is an image where this bug occurs attached to the issue #2735. We found this behavior by randomly creating and deleting files and using TSK to give us the list of active files. Feel free to ping me, if you need more information.

simsong:
Thank you. And did this fix resolve the problem?

simsong:
> @simsong, there is an image where this bug occurs attached to the issue #2735. We found this behavior by randomly creating and deleting files and using TSK to give us the list of active files. Feel free to ping me, if you need more information.

also, can you get me a copy of the image?

simsong:
> > @simsong, there is an image where this bug occurs attached to the issue #2735. We found this behavior by randomly creating and deleting files and using TSK to give us the list of active files. Feel free to ping me, if you need more information.
> 
> also, can you get me a copy of the image?

Also, @0x4d4c  can you resolve the conflict with `develop` ? Thanks.

PR #2868, opened 2023-09-20
7279 Add Linux OS accounts

APriestman:


bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -331,6 +332,84 @@ public OsAccount newWindowsOsAccount(String sid, String loginName, OsAccountReal
...
+	 * @param referringHost     The associated host.
+	 *
+	 * @return OsAccount.
+	 *
+	 * @throws TskCoreException                     If there is an error in
+	 *                                              creating the OSAccount.
+	 *
+	 */
+	@Beta
+	public OsAccount newLocalLinuxOsAccount(String uid, String loginName, Host referringHost) throws TskCoreException {
```
jayaramcs, line 352: should we add in OsAccountRealm.RealmScope realmScope since this a public api? It should be set to local by default

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -1730,6 +1855,69 @@ private OsAccountUpdateResult updateCoreWindowsOsAccountAttributes(OsAccount osA
...
+	 * @param osAccount     OsAccount that needs to be updated in the database.
+	 * @param uid           Account ID, may be null.
+	 * @param loginName     Login name, may be null.
+	 *
+	 * @return OsAccountUpdateResult Account update status, and the updated
+	 *         account.
+	 *
+	 * @throws TskCoreException If there is a database error or if the updated
+	 *                          information conflicts with an existing account.
+	 */
```
raman-bt, line 1874: Is this "beta" as well?

APriestman: Yeah probably should be. updated

raman-bt (approved):

PR #2871, opened 2023-09-22
Add resident $DATA of NTFS to tsk_file_layout

paul1278:
When using `loaddb`, all files of an NTFS file-system will be available on the `tsk_file_layout` table. Except for resident-files, they are missing there. I noticed this while I was working with this table and needed to extract the location for all files.

This PR adds those files to the `tsk_file-layout`-table, the `byte_start` & `byte_len` will point directly to the resident `$DATA`-buffer, the `sequence` is always 0. This allows complete transparent processing of those file-layouts, including extractions of files from the image.

It does this by calculating the offset in bytes relative to the start of the file-system (TSK_FS_ATTR.rd.offset) on NTFS file-systems. When I introduced new parameters to heavily used functions, I made an override for them to maintain compatibility.

This implementation works for my cases, I don't know if it works on all edge cases though.


simsong:
Thanks for the contribution. Can you rebase this to the current develop and add a test case to our unit-test system?

paul1278:
@simsong yes, I'll do that.

simsong:
> @simsong yes, I'll do that.

Thanks. There is now a catch2-based set of unit tests in the unit_test directory. I'm still trying to get the code coverage tools working properly, though. I have an idea of how to solve it.

simsong:
@paul1278  — Can you merge this against the current `develop` branch so that we can pull in the code-coverage stuff and all the current code base? It would also be great if you could give us a test case.

simsong:
@bcarrier - do we have a disk image that has a known resident file? I want to accept this, but don't want to do so without a test case.

tsk/fs/fs_attr.c
```diff
@@ -209,12 +210,13 @@ fs_attr_put_name(TSK_FS_ATTR * fs_attr, const char *name)
  * @param res_data Pointer to where resident data is located (data will
  * be copied from here into FS_DATA)
  * @param len Length of resident data
+ * @param offset Offset on disk
```
bcarrier, line 213: It looks like -1 is being used to mean special things. We should document that in this API and in the rd struct.

simsong:
Moving this to here:
https://github.com/sleuthkit/sleuthkit/pull/3263

tsk/fs/fs_attr.c
simsong: Okay. Let me know how you want to do that?

bcarrier:
Any existing NTFS test image should probably have a resident file. They are small files that are a few hundred bytes and don't get their own block allocated to them. They use space in the fixed-size MFT entry (1024-bytes).

Some context on this proposed change though. I considered this when we first built it out, but decided it against it because technically those blocks are allocated to $MFT (The master file table, where the MFT entries reside). 

I assume that with this change, TSK will report that two allocated files are associated with blocks for resident files:
* The $MFT
* The resident file that is using space in $MFT

According to the file system, only $MFT is using that block.

This may break code that assumes only one allocated file will be using a block.

simsong:
> This may break code that assumes only one allocated file will be using a block.

If this may break code, then should we not take it?

In any event, I want to close this PR and move to https://github.com/sleuthkit/sleuthkit/pull/3263, since we can control that one and can't make changes to this one.

PR #2875, opened 2023-10-05
Support for NTFS large clusters

nmandrikin:
Added support for NTFS partition with allocation unit size greater than 64K


simsong:
Hi. What happened to this one?

PR #2876, opened 2023-10-05
HFS additional null checks

nmandrikin:


simsong:
Hi. Thanks for your contribution. When you submitted this the appveyor checks failed. Can you rebase with the current develop and resubmit this when checks do not fail?

simsong:
Do you have a small HFS image that we can use?

simsong:
OK. I’ll find a small HFS test image so I can at least add something to the
test suite. Thanks.

PR #2879, opened 2023-10-22
HFS too strict check

nmandrikin:
Size of hfs_thread structure is equal to 520 bytes. This check is too strict and results in the loss of valid files in HFS file systems.


uckelman-sf:
This should be closed. The bug is fixed in #2940.

PR #2889, opened 2023-12-05
Add aarch64 support to Java bindings for mac and linux

cho-m:
Was working on updating `autopsy` in Homebrew (https://github.com/Homebrew/homebrew-core/pull/156402) and noticed that it doesn't start up on ARM macOS due to missing NATIVELIBS in `sleuthkit` JAR even though native libraries are built.

`System.getProperty("os.arch")` is "aarch64" when tested on ARM macOS and on Docker Linux.


simsong:
Hi. When you submitted this the appveyor checks failed. Can you rebase this to develop and resubmit so that the checks do not fail?

PR #2892, opened 2024-01-16
CT-7939 TSK content tag bug fix 

eugene7646:
Previously List<ContentTag> getContentTagsByTagName(TagName tagName, long dsObjId) method had a bug where it would only find and return content tags for Content objects that have entries in tsk_files. OsAccounts do not have entries in tsk_files and therefore existing content tags for OsAccounts were not returned. This PR fixes that bug.


raman-bt (changes requested):
For posterity's sake, it would be useful to have a one or two line comment explaining what the queries are doing and why the union?

eugene7646:
> For posterity's sake, it would be useful to have a one or two line comment explaining what the queries are doing and why the union?

@raman-bt Done.

PR #2906, opened 2024-03-27
Performance improvements for domain controller processing with 10,000 + users

jayaramcs:


bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
```diff
@@ -414,7 +429,7 @@ public OsAccount newLocalLinuxOsAccount(String uid, String loginName, Host refer
 	/**
 	 * Creates a OS account with the given uid, name, and realm.
 	 *
-	 * @param uniqueId      Account sid/uid. May be null.
+	 * @param uniqueId      Account sid/uid. May be null. Saved in uppercase.
 	 * @param loginName     Login name. May be null only if SID is not null.
```
raman-bt, line 433: Maybe update the comments for name/realm as well that they are saved in lower case?

raman-bt (changes requested):

bindings/java/src/org/sleuthkit/datamodel/OsAccountManager.java
APriestman: I was also thinking that, but it doesn't look like this method does the conversion, so maybe the comments should all say what case is expected?

PR #2923, opened 2024-05-21
Add BitLocker support

APriestman:


tsk/auto/is_image_supported.cpp
```diff
@@ -44,6 +46,47 @@ bool TskIsImageSupported::isImageEncrypted()
...
+* reason the image is not supported
+*/
+std::string TskIsImageSupported::getSingleLineErrorMessage() {
+    // If we have this, we are very confident we have a BitLocker-protected partition
+    // and that we either need a password or that the one given was incorrect.
+    if (m_bitlockerError) {
+        if (strnlen(m_bitlockerDesc, 1024) > 0) {
+            return std::string(m_bitlockerDesc);
+        }
+        return "BitLocker error";
```
crayy8, line 60: Do we ever expect this generic "Bitlocker error" to be returned or is this just a safety message?

tsk/fs/encryptionHelper.cpp
```diff
@@ -0,0 +1,181 @@
...
+* - We found encryption but had an unspecified error in initialization
+* Returns -1 if:
+* - We got far enough to be confident that it's Bitlocker and have a specific error message to get back to the user
+*/
+int handleBitlocker(TSK_FS_INFO* a_fs_info, const char* a_pass) {
+#ifdef HAVE_LIBMBEDTLS
+	BitlockerParser* bitlockerParser = new BitlockerParser();
+	BITLOCKER_STATUS status = bitlockerParser->initialize(a_fs_info->img_info, a_fs_info->offset, a_pass);
+	if (status == BITLOCKER_STATUS::NOT_BITLOCKER) {
+		return 0;
```
crayy8, line 43: Need to delete bitlockerParser  here

tsk/fs/encryptionHelper.cpp
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	if (status != BITLOCKER_STATUS::SUCCESS) {
+
+		// If we have some specific error cases we want to get that information back to the user
+		if (status == BITLOCKER_STATUS::WRONG_PASSWORD) {
```
crayy8, line 49: Need to delete bitlockerParser  in all of these conditions

tsk/fs/encryptionHelper.cpp
```diff
@@ -0,0 +1,181 @@
...
+		return -1;
+	}
+
+	// Store the BitLocker data to use when reading the volume
+	a_fs_info->encryption_type = TSK_FS_ENCRYPTION_TYPE_BITLOCKER;
+	a_fs_info->encryption_data = (void*)bitlockerParser;
+	a_fs_info->flags |= TSK_FS_INFO_FLAG_ENCRYPTED;
+	a_fs_info->block_size = bitlockerParser->getSectorSize();
+	// We don't set a_fs_info->decrypt_block here because Bitlocker needs to handle both reading in the block
+	// and doing the decryption since some sectors may have been relocated
```
crayy8, line 85: need to delete bitlockerParser

tsk/fs/encryptionHelper.cpp
```diff
@@ -0,0 +1,181 @@
...
+* - There was no encryption found
+* - We found encryption and did all the initialization successfully
+* - We found encryption but had an unspecified error in initialization  
+* Returns -1 if:
+* - We found encryption and got far enough that we're confident we should not continue trying to parse the file system and
+*     have potentially useful feedback to give the user (like that the password was incorrect)
+*/
+int handleVolumeEncryption(TSK_FS_INFO* a_fs_info, const char* a_pass) {
+	int ret = 0;
+#ifdef HAVE_LIBMBEDTLS
```
crayy8, line 103: Are we ever expecting to not have MBEDTLS? Does it make sense to return 0 if we don't have the library, we can't check so we don't know if there is encryption. Is that the idea that we want to fail safe if we cant check and assume there is no encryption?

tsk/util/Bitlocker/MetadataValueVolumeMasterKey.cpp
```diff
@@ -0,0 +1,47 @@
...
+    // Format should be:
+    // - 16 byte GUID
+    // - 8 byte last modification timestamp
+    // - 2 byte unknown
+    // - 2 byte key protection type
+    // - list of metadata entries
+    memcpy(guid, buf, 16);
+    lastModificationTime = tsk_getu64(TSK_LIT_ENDIAN, &(buf[16]));
+    unknown = tsk_getu16(TSK_LIT_ENDIAN, &(buf[24]));
+    keyProtectionType = getKeyProtectionTypeEnum(tsk_getu16(TSK_LIT_ENDIAN, &(buf[26])));
```
crayy8, line 32: do we need to be checking buf size before accessing these offsets

tsk/util/Bitlocker/MetadataValueVolumeMasterKey.cpp
```diff
@@ -0,0 +1,47 @@
...
+        return;
+    }
+
+    // Format should be:
+    // - 16 byte GUID
+    // - 8 byte last modification timestamp
+    // - 2 byte unknown
+    // - 2 byte key protection type
+    // - list of metadata entries
+    memcpy(guid, buf, 16);
```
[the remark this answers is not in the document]
crayy8: nevermind i see headerLen is 28

tsk/util/Bitlocker/MetadataValueVolumeMasterKey.cpp
crayy8: nevermind i see headerLen is 28

tsk/util/Bitlocker/MetadataEntry.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (!metadataValue->wasLoadedSuccessfully()) {
+        delete metadataValue;
+        return nullptr;
+    }
+
+    // Create the entry
+    MetadataEntry* entry = new MetadataEntry();
+    if (entry == nullptr) {
+        writeError("MetadataEntry::createMetadataEntry: Error allocating memory");
+        return nullptr;
```
crayy8, line 63: don't we need to delete metadatValue here?

tsk/util/Bitlocker/MetadataValueAesCcmEncryptedKey.cpp
```diff
@@ -0,0 +1,258 @@
...
+    if (ret != BITLOCKER_STATUS::SUCCESS) {
+        memset(decryptedData, 0, encryptedDataLen);
+        free(decryptedData);
+        return ret; // Propagate the return value in case it indicates an incorrect password
+    }
+
+    // Try to create the key entry
+    *keyEntry = MetadataEntry::createMetadataEntry(&(decryptedData[BITLOCKER_KEY_MAC_LEN]), encryptedDataLen);
+    memset(decryptedData, 0, encryptedDataLen);
+    free(decryptedData);
```
crayy8, line 75: should we set this to a null pointer since we freed it

tsk/util/Bitlocker/MetadataValueAesCcmEncryptedKey.cpp
```diff
@@ -0,0 +1,258 @@
...
+* 
+* @param key      Key bytes
+* @param keyLen   Length of the key
+* @param keyEntry Will hold decrypted keyEntry object on success
+* 
+* @return SUCCESS if key is successfully decrypted
+*         GENERAL_ERROR if an unspecified error occurs
+*         WRONG_PASSWORD if the supplied key appears to be incorrect
+*/
+BITLOCKER_STATUS MetadataValueAesCcmEncryptedKey::decrypt(uint8_t* key, size_t keyLen, MetadataEntry** keyEntry) {
```
crayy8, line 50: Should we add a comment that the caller must free keyEntry?

tsk/fs/encryptionHelper.cpp
APriestman: I'm confused? This is the success case. I need it to do the decryption later - it's saved in a_fs_info->encryption_data = (void*)bitlockerParser; for later use/deletion.

tsk/fs/encryptionHelper.cpp
APriestman: Resolved I think - mbedtls will not be present in a nolibs build and is not currently included in the linux build (and it would be optional there anyway). And we definitely want to continue trying to open the file system in that case.

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+        writeError("BitlockerParser::parseVMKEntry: Error casting MetadataValueStretchKey");
+        return BITLOCKER_STATUS::GENERAL_ERROR;
+    }
+
+    // Decrypt it using the stretched key, which should produce a MetadataEntry of type KEY.
+    // This includes testing a 16-byte message authentication code to verify that
+    // the decrypted key is correct.
+    MetadataEntry* keyEntry = nullptr;
+    ret = aesCcmKey->decrypt(stretchedKey, BITLOCKER_STRETCH_KEY_SHA256_LEN, &keyEntry);
+    if (ret != BITLOCKER_STATUS::SUCCESS) {
```
crayy8, line 654: should we add a comment that keyEntry will always be a valid allocated obj if bitlocker_Status is success since we aren't verifying that its not a nullptr. I thought we had an issue until I went back to see what decrypt() was doing.

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+        return ret;
+    }
+
+    // Make sure the value is of type Key
+    if (keyEntry->getValueType() != BITLOCKER_METADATA_VALUE_TYPE::KEY) {
+        writeError("BitlockerParser::parseVMKEntry: keyEntry does not have value of type KEY ("
+            + convertMetadataValueTypeToString(keyEntry->getValueType()) + ")");
+        return BITLOCKER_STATUS::GENERAL_ERROR;
```
crayy8, line 662: shouldn't we be freeing keyEntry here?

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+        // password they probably didn't even enter is incorrect.
+        writeError("BitlockerParser::parseVMKEntry: Failed to decrypt VMK using the supplied clear key");
+        return BITLOCKER_STATUS::GENERAL_ERROR;
+    }
+
+    // Make sure the value is of type Key
+    if (keyEntry->getValueType() != BITLOCKER_METADATA_VALUE_TYPE::KEY) {
+        writeError("BitlockerParser::parseVMKEntry: keyEntry does not have value of type KEY ("
+            + convertMetadataValueTypeToString(keyEntry->getValueType()) + ")");
+        return BITLOCKER_STATUS::GENERAL_ERROR;
```
crayy8, line 731: free keyEntry here?

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+    MetadataEntry* keyEntry = nullptr;
+    BITLOCKER_STATUS ret = aesCcmKey->decrypt(keyBytes, keyLen, &keyEntry);
+    if (ret != BITLOCKER_STATUS::SUCCESS || keyEntry == nullptr) {
+        return ret;
+    }
+
+    // Make sure the value is of type Key
+    if (keyEntry->getValueType() != BITLOCKER_METADATA_VALUE_TYPE::KEY) {
+        writeError("BitlockerParser::parseVMKEntry: keyEntry does not have value of type KEY ("
+            + convertMetadataValueTypeToString(keyEntry->getValueType()) + ")");
```
crayy8, line 791: free keyEntry

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+        writeError("BitlockerParser::getFullVolumeEncryptionKey: Error loading keys");
+        return BITLOCKER_STATUS::GENERAL_ERROR;
+    }
+
+    // Then use that key to decrypt the FVEK entry, which should produce a MetadataEntry of type KEY.
+    // This includes testing a 16-byte message authentication code to verify that
+    // the decrypted key is correct.
+    MetadataEntry* keyEntry = nullptr;
+    BITLOCKER_STATUS ret = aesCcmKey->decrypt(keyBytes, keyLen, &keyEntry);
+    if (ret != BITLOCKER_STATUS::SUCCESS || keyEntry == nullptr) {
```
crayy8, line 784: other calls to decrypt() only check the status we should be consistent if we should be checking for keyEntry being nullptr

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+/**
+* Use the decrypted full volume encryption key (FVEK) to initialize the appropriate AES contexts
+* 
+* @param fvekEntry The entry containing the decrypted FVEK
+* 
+* @return SUCCESS on success, GENERAL_ERROR otherwise
+*/
+BITLOCKER_STATUS BitlockerParser::setKeys(MetadataEntry* fvekEntry) {
+
+    MetadataValueKey* fvek = dynamic_cast<MetadataValueKey*>(fvekEntry->getValue());
```
crayy8, line 843: should we be checking is fvekEntry is null

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+
+    // Make sure the value is of type Key
+    if (keyEntry->getValueType() != BITLOCKER_METADATA_VALUE_TYPE::KEY) {
+        writeError("BitlockerParser::parseVMKEntry: keyEntry does not have value of type KEY ("
+            + convertMetadataValueTypeToString(keyEntry->getValueType()) + ")");
+        return BITLOCKER_STATUS::GENERAL_ERROR;
+    }
+
+    // Use the decrypted FVEK to intialize the AES contexts we'll use to decrypt the volume
+    return (setKeys(keyEntry));
```
crayy8, line 796: I'm a little confused here but I think we should be freeing keyEntry here or within setKeys since it doesn't look like we continuing to use it?

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+
+    // Attempt to parse each of the VMK entries, keeping track of some specific errors if they occur.
+    BITLOCKER_STATUS ret = BITLOCKER_STATUS::GENERAL_ERROR;
+    MetadataEntry* vmk = nullptr;
+    bool possibleMissingPassword = false;
+    bool possibleWrongPassword = false;
+    bool possibleUnsupportedProtectionType = false;
+    for (auto it = vmkEntries.begin(); it != vmkEntries.end(); ++it) {
+        ret = parseVMKEntry(*it, &vmk);
+        if (ret == BITLOCKER_STATUS::SUCCESS && vmk != nullptr) {
```
crayy8, line 460: I think we should be freeing vmk at the end of this loop?

APriestman: So I think we don't need to because vmk will never be allocated if we failed and on success we save it to be freed later. I think it just needs better comments.

tsk/util/Bitlocker/MetadataValueStretchKey.h
```diff
@@ -0,0 +1,36 @@
...
+	MetadataValueStretchKey(BITLOCKER_METADATA_VALUE_TYPE a_valueType, uint8_t* buf, size_t bufLen);
+
+	BITLOCKER_STATUS parseStretchKeyUsingPassword(uint8_t* passwordHash, size_t passwordHashLen, uint8_t* stretchKey, size_t stretchKeyLen);
+
+	~MetadataValueStretchKey();
+private:
+	BITLOCKER_STATUS generateStretchedKey(uint8_t* passwordHash, size_t passwordHashLen, uint8_t* salt, size_t saltLen, uint8_t* result, size_t resultLen);
+
+	const size_t headerLen = 20;
+	BITLOCKER_ENCRYPTION_TYPE encryptionType;
```
crayy8, line 31: Should we be consistent here and renamed all of these to start with m_. I was confused while reviewing thinking that some places should be freeing data but then realized after doing a search that they were member variables and not local variables. I don't have a strong opinion just seems to deviate from our standard naming convention?

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+int BitlockerParser::decryptSectorAESCBC_noDiffuser(uint64_t offset, uint8_t* data) {
+
+    // Make temporary buffer to copy encrypted data to
+    uint8_t* encryptedData = (uint8_t*)tsk_malloc(m_sectorSize);
+    if (encryptedData == nullptr) {
+        writeError("BitlockerParser::decryptSectorAESCBC_noDiffuser: Error allocating encryptedData");
+        return -1;
+    }
+    memcpy(encryptedData, data, m_sectorSize);
+    memset(data, 0, m_sectorSize);
```
crayy8, line 1273: We don't know the size of data here at least from this method, would it be safer to pass in the size of data? Should data always be m_sectorSize and if not throw an error?

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+
+    // Make temporary buffer to copy encrypted data to
+    uint8_t* encryptedData = (uint8_t*)tsk_malloc(m_sectorSize);
+    uint8_t* tempBuffer = (uint8_t*)tsk_malloc(m_sectorSize);
+    if (encryptedData == nullptr || tempBuffer == nullptr) {
+        writeError("BitlockerParser::decryptSectorAESCBC_noDiffuser: Error allocating encryptedData or tempBuffer");
+        return -1;
+    }
+    memcpy(encryptedData, data, m_sectorSize);
+    memset(data, 0, m_sectorSize);
```
crayy8, line 1376: same comment as before, should we be passing in size of data?

crayy8 (changes requested):
First pass done

tsk/fs/encryptionHelper.cpp
APriestman: I changed the logic a little here so please give it a quick look

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1553 @@
...
+            possibleUnsupportedProtectionType = true;
+        }
+    }
+
+    // If we failed to decrypt any of the VMK entries return a specific error if we have one.
+    // Note that the order is important here - if we have a normal password that failed to decrypt the
+    // password protected VMK entry, we don't want to report that we didn't have a recovery password to
+    // try in the recovery password protected VMK.
+    if (ret != BITLOCKER_STATUS::SUCCESS || vmk == nullptr) {
+        writeError("BitlockerParser::setVolumeMasterKey: Failed to extract Volume Master Key");
```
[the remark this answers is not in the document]
APriestman: Added comment above that vmk will never be allocated on error conditions

tsk/fs/encryptionHelper.cpp
crayy8: I missed that. This can be resolved.

tsk/fs/encryptionHelper.cpp
crayy8: Looks good.

tsk/fs/encryptionHelper.cpp
crayy8: This can be resolved. This was a lack of understanding on my part.

tsk/util/Bitlocker/MetadataValueStretchKey.h
APriestman: I think I got them all

tsk/util/Bitlocker/BitlockerParser.cpp
crayy8: Can be resolved based on previous comment https://github.com/sleuthkit/sleuthkit/pull/2923#discussion_r1610131510

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1565 @@
...
+        writeError("BitlockerParser::handlePassword: Error converting password to UTF16");
+        // We'll return this error if we fail to parse it as a recovery password
+        ret = BITLOCKER_STATUS::GENERAL_ERROR;
+    }
+
+    // Try to parse the password as a recovery password. We don't really want to add another password field to TSK
+    // so we'll just use any supplied password as a normal password and potentially as a recovery password (if it has the right format).
+    // Example: 162294-601403-607013-155265-438779-479028-357148-102091
+    // Each part should be divisible by 11 to produce a 16-bit value. Those eight values are combined to form a 16 byte key.
+    std::regex recoveryPasswordPattern("^(\\d{1,6})-(\\d{1,6})-(\\d{1,6})-(\\d{1,6})-(\\d{1,6})-(\\d{1,6})-(\\d{1,6})-(\\d{1,6})$");
```
crayy8, line 1100: Is each group supposed to be 6 digits or can it really be 1-6 digits like the regex suggests?

APriestman: I thought I had a shorter one but it probably just had a bunch of 1's since it does appear that the groups are zero-padded.

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -0,0 +1,1565 @@
...
+            m_encryptionType = fvek->getEncryptionType();
+            return BITLOCKER_STATUS::SUCCESS;
+        }
+        return BITLOCKER_STATUS::GENERAL_ERROR;
+    }
+
+    m_encryptedDataBuffer = (uint8_t*)tsk_malloc(m_sectorSize);
+    m_diffuserTempBuffer = (uint8_t*)tsk_malloc(m_sectorSize);
+    if (m_encryptedDataBuffer == nullptr || m_diffuserTempBuffer == nullptr) {
+        writeError("BitlockerParser::setKeys: Error allocating temp buffers");
```
crayy8, line 886: Very unlikely, but the first allocation could succeed and the second could fail. We would then fail to release the memory for the first allocation.

APriestman: I think the destructor would get it later

crayy8: good to resolve.

tsk/util/Bitlocker/BitlockerParser.cpp
crayy8: good to resolve.

PR #2940, opened 2024-08-12
Fix incorrect bounds check for hfs_thread structs

### LINKED ISSUE #2899: HFS: Wrong calculation of the real hfs_thread size.
According to HFS documentation  the hfs_thread structure is of variable size.  But in  tsk_hfs.h it is defined as fixed (member of hfs_thread structure which is hfs_uni_str substructure  is defined as fixed size of 512 bytes).

So in my opinion test in hfs_dent.c line 261 is wong. 

`
else if (rec_type == HFS_FOLDER_THREAD) {
 if ((nodesize < sizeof(hfs_thread)) || (rec_off2 > nodesize - sizeof(hfs_thread))) {

`

There are some records which fails on it, because they are smaller than sizeof(hfs_thread), but they are built correctly according to file system spec, and they are in my opinion valid. 

So I think hfs_thread size calculation should be as follows :

hfs_thread* thread = (hfs_thread*)&rec_buf[rec_off2];
const int32_t thread_size = 0x50 + tsk_getu16(hfs->fs_info.endian, thread->name.length);

where:
     0x50 is size of constant part of the structure  which is size of hfs_thread header  +
            length of the name field "thread->name.length".

So summarizing we should calculate thread size as  :  size of constant part  + size of variable part.

Unfortunately I cannot provide samples. They do not belong to me and are confidential

Kind regards 
Bogdan

uckelman-sf:
Fixes #2899. The problem was caused by an incorrect bounds check introduced in #2647.

 The `hfs_thread` struct is sized for the maximum possible name length, but the actual on-disk storage of the name is variable length. It's incorrect to require the buffer to be the maximum size, and results in entries being skipped as corrupt. Instead, the right thing to do is check that enough of the data fits into the buffer so that the name length can be read, and if so, read the name length and check that a name of that length also fits.

### DIFF
--- a/tsk/fs/hfs_dent.c
+++ b/tsk/fs/hfs_dent.c
@@ -258,12 +258,20 @@ hfs_dir_open_meta_cb(HFS_INFO * hfs, int8_t level_type,
 
         /* This will link the folder to its parent, which is the ".." entry */
         else if (rec_type == HFS_FOLDER_THREAD) {
-            if ((nodesize < sizeof(hfs_thread)) || (rec_off2 > nodesize - sizeof(hfs_thread))) {
+            hfs_thread *thread = (hfs_thread *) & rec_buf[rec_off2];
+
+            // hfs_thread is of variable size on disk. The minimum size is
+            // 10 bytes (8 for the fields, 2 for the length of an empty name).
+            // First, check that we can read as far as the name length; then
+            // get the name length and check that the whole record fits into
+            // the buffer.
+            if (rec_off2 > nodesize - (sizeof(hfs_thread) - sizeof(hfs_uni_str) + 2) ||
+                rec_off2 > nodesize - (sizeof(hfs_thread) - sizeof(hfs_uni_str) + 2 + tsk_getu16(hfs->fs_info.endian, thread->name.length))) {
                 tsk_error_set_errno(TSK_ERR_FS_GENFS);
                 tsk_error_set_errstr("hfs_dir_open_meta: nodesize value out of bounds");
                 return HFS_BTREE_CB_ERR;
             }
-            hfs_thread *thread = (hfs_thread *) & rec_buf[rec_off2];
+
             strcpy(info->fs_name->name, "..");
             info->fs_name->meta_addr =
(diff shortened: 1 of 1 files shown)


APriestman:
Do you have any test data you could share?

jonstewart:
We noticed it with a SANS training image. Digital Corpora has https://corp.digitalcorpora.org/corpora/drives/nps-2009-hfsjtest1/, not sure whether that exhibits missing entries in `fls`/`fiwalk`. I imagine that most HFS+ images would demonstrate the bug.

jonstewart:
#2879 notes the same issue.

simsong:
> We noticed it with a SANS training image. Digital Corpora has https://corp.digitalcorpora.org/corpora/drives/nps-2009-hfsjtest1/, not sure whether that exhibits missing entries in `fls`/`fiwalk`. I imagine that most HFS+ images would demonstrate the bug.

Yea digital corpora! I need to do some testing to see how the various code coverage tools happen if multiple executables get run as part of a single test. I suspect that we can't just run `fiwalk` repeatedly and get accurate code coverage results.

simsong:
Can you figure out why this pull request triggered no runs?

uckelman-sf:
> Can you figure out why this pull request triggered no runs?

I think it did: 
image

simsong:
I did.

simsong:
@uckelman-sf  - Where are all of your commits coming from ? Your internal fork?

simsong:
> We noticed it with a SANS training image. Digital Corpora has https://corp.digitalcorpora.org/corpora/drives/nps-2009-hfsjtest1/, not sure whether that exhibits missing entries in `fls`/`fiwalk`. I imagine that most HFS+ images would demonstrate the bug.

Is the SANS training image publicly available?

jonstewart:
> s the SANS training image publicly available?

No, but maybe they'd grant permission for them to be shared with you privately (they gave me that permission in the past, but that was a long time ago).

My rough understanding of this HFS+ bug, though, is that it'll probably be exhibited with any normal nontrivial HFS+ image.

simsong:
Well, let’s make another image that demonstrates the bug then

PR #2941, opened 2024-08-23
Fix source file permissions

uckelman-sf:
chmod a-x all of the C/C++ source files and headers.


simsong:
Hi. Thanks for contributing this. However, the appveyor build fails:
https://ci.appveyor.com/project/bcarrier/sleuthkit/builds/50461252/job/haljfwcfteabm972
image
Can you fix this?

uckelman-sf:
Maybe, but it's not related to this PR. Let me see...

simsong:
> It looks like all Linux CI builds are failing due to this.

Is this something you can fix? 
My goal is to start adding GitHub actions as I find them easier to work with than AppVeyor, but there is no reason that we can't use both.

simsong:
> Maybe, but it's not related to this PR. Let me see...

True, but I like having clean builds before accepting PRs!

uckelman-sf:
I was going to suggest moving to GitHub Actions, as I've had good experience with them in the past. They're also _much_ faster to start than appveyor. I'm still waiting for the job to run to see if my change worked...

simsong:
> I was going to suggest moving to GitHub Actions, as I've had good experience with them in the past. They're also _much_ faster to start than appveyor. I'm still waiting for the job to run to see if my change worked...

My plan is to run them in parallel.

uckelman-sf:
There we go. The test runner didn't have autoconf installed. It works now.

simsong:
Yay! Let's see if I can make this happen now...

uckelman-sf:
Woo! Thanks!

uckelman-sf:
@simsong Next one is here: PR 2942.

PR #2957, opened 2024-09-07
Catch2 switch (from Catch2-switch)

simsong:


simsong:
Giving up on this one. Will let @uckelman-sf rebase from `develop` to get new GitHub action YAML files and then make the catch2 migration.

PR #2963, opened 2024-09-11
Fixed incorrectly cased header file name

uckelman-sf:
The case of header filenames doesn't matter on Windows, but needs to be correct on systems where paths are case-sensitive.


uckelman-sf:
@simsong Ready for review.

simsong (approved):
Well, it works on windows.

PR #2964, opened 2024-09-11
Don't define fseeko if it's already defined

uckelman-sf:
This fixes a compilation error when cross-compiling with MinGW.

### DIFF
--- a/tsk/base/tsk_os.h
+++ b/tsk/base/tsk_os.h
@@ -51,7 +51,10 @@
 #define roundup(x, y)   \
     ( ( ((x)+((y) - 1)) / (y)) * (y) )
 
+#ifndef fseeko
 #define fseeko fseek
+#endif
+
 #define daddr_t int
 #endif
 


tsk/base/tsk_os.h
```diff
@@ -51,7 +51,10 @@
 #define roundup(x, y)   \
     ( ( ((x)+((y) - 1)) / (y)) * (y) )
 
+#ifndef fseeko
 #define fseeko fseek
+#endif
```
simsong, line 56: what if fseeko is a function?
https://www.codecogs.com/library/computing/c/stdio.h/fseek.php?alias=fseeko

Shouldn't we also check to see if it is a function in sys/types.h?

It's a built-in on my mac. Of course, fseek instead of fseeko is probably safe...
```
simsong@Simsons-MacBook-Pro sleuthkit % grep -r fseeko /Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr/include/                                                 (develop)sleuthkit
/Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr/include//c++/v1/fstream:    if (::fseeko(__file_, __width > 0 ? __width * __off : 0, __whence))
/Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr/include//c++/v1/fstream:    if (::fseeko(__file_, __sp, SEEK_SET))
/Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr/include//c++/v1/fstream:        if (::fseeko(__file_, -__c, SEEK_CUR))
/Library/Developer/CommandLineTools/SDKs/MacOSX14.4.sdk/usr/include//stdio.h:int	 fseeko(FILE * __stream, off_t __offset, int __whence);
simsong@Simsons-MacBook-Pro sleuthkit %                                                                                                                                    
```

simsong: what do you think?

simsong (commented):
what do you think?

tsk/base/tsk_os.h
uckelman-sf: Defining `fseeko` or not doesn't apply globally here. Line 55, the line where `fseeko` is `#define`d, is inside an `#ifdef __MINGW32__`, so nothing but MinGW will ever see this. What your system compiler on MacOS has for `fseeko` isn't relevant for that reason.

simsong: Thanks!

PR #2968, opened 2024-09-11
Update Github actions to compile on mingw; check that tools are statically linked in static builds; clearly surface all test images used in tests

uckelman-sf:


simsong:
@uckelman-sf - do you want me to take this?

uckelman:
No, I'll get back to this on Monday.

simsong (approved):
Really nice. Thanks for doing this.

simsong:
> Typos. Augh, yaml.
It happens to all of us.  Me more than you.

simsong (approved):
Neat. How can we pick up these errors automatically?

simsong:
Ick. Look at all of those build errors...

simsong (commented):
Why not?

uckelman-sf:
> Neat. How can we pick up these errors automatically?

This PR is picking up these errors automatically. The manual part is fixing them.

uckelman-sf:
@simsong Ready for review.

There is cleanup I'd like to do, but if I do that before this is merged I'm afraid I'll get clobbered again by some other merged PR and have to spend several days resolving conflicts and fixing problems AddressSanitizer reveals in the changes which got in before mine.

simsong:
> @simsong Ready for review.
> 
> There is cleanup I'd like to do, but if I do that before this is merged I'm afraid I'll get clobbered again by some other merged PR and have to spend several days resolving conflicts and fixing problems AddressSanitizer reveals in the changes which got in before mine.

I will avoid making any changes myself until you give me the 'all clear.'

simsong:
@uckelman-sf  - did you see my note about disk images?

uckelman-sf:
> did you see my note about disk images?

No, I did not. There was a notification about a comment of yours that won't load when I go to the page where the comment is supposed to be.

simsong:
> > did you see my note about disk images?
> 
> No, I did not. There was a notification about a comment of yours that won't load when I go to the page where the comment is supposed to be.

Ah, GitHub...
Here is the page:
https://github.com/sleuthkit/sleuthkit/commit/cd734440c72ed13dd4f1667a67bc8a8e5f3de412

Here is the comment:
@uckelman-sf - there are some problems here.
1. I'm using a YAML file to define the test images because they need to be got from digitalcorpora.org and other places, and some may need processing (unzip to extract a particular file). I don't want to have to keep multiple files in sync. 
2. I want to know if LIBEWF isn't available because I want to test that code as part of the regression and acceptance tests. So I don't want the test `#ifdef`'ed out if the library isn't available, I want the test to fail.
3. The Appveyor system that SKL is using puts the `from_brian` directory in the `$HOME/from_brian`, and not as a subdir of `tests/`

uckelman-sf:
That's the page I went to. I still can't see the comment there:
image

simsong:
Well, I pasted my comment about disk images in the comment above, so you
have it now!

uckelman-sf:
>     2. I want to know if LIBEWF isn't available because I want to test that code as part of the regression and acceptance tests. So I don't want the test `#ifdef`'ed out if the library isn't available, I want the test to fail.

Having tests fail because an optional dependency is not present isn't right. It should be possible for anyone who has specified a legal set of optional dependencies to successfully run `make check`---but that won't happen if tests which require libewf are compiled unconditionally.

A better approach is to have `configure` error out when a requested optional dependency isn't found. This way, you'll immediately have a failure if you specify `--with-libewf` but libewf isn't available, instead of having that failure occur far into the build. 

This is a problem I've wanted to solve for years. My team has wasted literally hundreds of CPU hours because libtsk wasn't built with a dependency we asked for and the first thing that caught it was farther on in our build process.

PR #2999 fixes this problem.

simsong:
I like that approach and I see that you have implemented it.

uckelman-sf:
I'm still mulling the other two points.

I'm reorganizing the tests now, but you're clear to go ahead with whatever you were going to do next. #2986 was the PR I wanted in before other changes.

PR #2975, opened 2024-09-12
Remove unnecessary m4_includes

uckelman-sf:
`m4` is the set as the include directory. Use of `m4_include` for specific files is unnecessary.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -5,8 +5,6 @@ AC_PREREQ([2.69])
 
 AC_INIT([sleuthkit],[4.12.1])
 AC_CONFIG_AUX_DIR([config])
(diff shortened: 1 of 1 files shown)


simsong (approved):
You mean everything in the directory just gets magically included? I don't have to explicitly include each one?

uckelman-sf:
> You mean everything in the directory just gets magically included? I don't have to explicitly include each one?

Exactly.

simsong:
I love the magic.

simsong:
Right, but with `-I` you still need to `#include` each file...

uckelman-sf:
Yes, the analogy is between include filenames for `-I` and and macro names for `AC_CONFIG_MACRO_DIR`.

PR #2981, opened 2024-09-13
Don't use recursive Make

uckelman-sf:
This PR eliminates the bulk of our recursive usage of Make, which has the following advantages:

* More of the DAG is now transparent, for better parallelism
* Almost all paths are now relative to the project root, which makes them easier to get right
* Compiler and linker flags are centralized in one place

A follow-on PR will complete the process, and eliminate a lot of repetition. This PR is intended to leave things in an improved, working state.


uckelman-sf:
@simsong Ready for review.

PR #2989, opened 2024-09-18
Fix compilation errors exposed by compiling for MinGW

uckelman-sf:


simsong:
If we do this, we probably want to have a new unit test that also does a mingw compilation.

simsong:
> If we do this, we probably want to have a new unit test that also does a mingw compilation.

I did this for `bulk_extractor`, BTW, with a compile and then a test using `wine`.

uckelman-sf:
Closing, as all commits here are part of #2968 and can't be demonstrated to compile without it.

PR #2990, opened 2024-09-27
Switch fiwalk from libregex to std::regex

uckelman-sf:
fiwalk depends on libregex. There isn't a mingw package for libregex in Ubuntu, which means we can't compile for mingw in the test runners as-is. C++11 has `std::regex`, which has the same functionality as libregex with no additional dependencies.

This PR adjusts fiwalk to use `std::regex`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -259,7 +259,6 @@ AC_SUBST([CXXLD])
 
 dnl Dependencies for fiwalk
 AC_CHECK_FUNCS([getline])
-AC_SEARCH_LIBS(regexec, [regex], , AC_MSG_ERROR([missing regex]))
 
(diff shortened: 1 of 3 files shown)


uckelman-sf:
@simsong Ready for review.

simsong:
Neat, but we really need to add a test for mingw compilation. Do you want to create that, or should I?

simsong (approved):

uckelman-sf:
> Neat, but we really need to add a test for mingw compilation. Do you want to create that, or should I?

I've already done it: #2968

This PR is pulling some parts out of that one to make it easier to review when it's finished.

PR #2992, opened 2024-09-29
Add new disk images to unit tests; reduce Actions matrix to just mac and ubuntu once each.

simsong:


simsong (commented):
Completed as suggested by @jonstewart

PR #2993, opened 2024-10-01
Use libewf_handle_read_buffer_at_offset with newer versions of libewf

uckelman-sf:
libewf_handle_read_random() was renamed libewf_handle_read_buffer_at_offset() as of libewf 20141129. Check for libewf_handle_read_buffer_at_offseton so we can select the correct function name.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -143,8 +143,19 @@ dnl Check if we should link with zlib
 TSK_OPT_DEP_CHECK([zlib], [ZLIB], [zlib], [zlib.h], [z], [inflate])
 dnl Check if we should link with libbfio
 TSK_OPT_DEP_CHECK([libbfio], [], [libbfio], [libbfio.h], [bfio], [libbfio_get_version])
+
 dnl Check if we should link with libewf
 TSK_OPT_DEP_CHECK([libewf], [EWF], [libewf], [libewf.h], [ewf], [libewf_get_version])
+dnl Check for libewf_handle_read_buffer_at_offset, in newer versions of libewf
+AS_IF(
+  [test "x$ax_libewf" != "xno"],
+  [AC_CHECK_LIB(
+    [ewf],
+    [libewf_handle_read_buffer_at_offset],
+    [AC_DEFINE([HAVE_LIBEWF_HANDLE_READ_BUFFER_AT_OFFSET], [1], [Define to 1 if you have the `libewf_handle_read_buffer_at_offset' function.])]
+  )]
+)
+
 dnl Check if we should link with libqcow
 TSK_OPT_DEP_CHECK([libqcow], [QCOW], [libqcow], [libqcow.h], [qcow], [libqcow_get_version])
 dnl Check if we should link with libvhdi
--- a/tsk/img/ewf.cpp
+++ b/tsk/img/ewf.cpp
@@ -68,7 +68,7 @@ ewf_image_read(TSK_IMG_INFO * img_info, TSK_OFF_T offset, char *buf,
 
     tsk_take_lock(&(ewf_info->read_lock));
 #if defined( HAVE_LIBEWF_V2_API )
-#if defined(LIBEWF_VERSION) && LIBEWF_VERSION >= 20141129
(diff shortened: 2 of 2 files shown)


uckelman-sf:
@simsong Ready for review.

uckelman-sf:
I agree that the display of results is not great. I'll work on that in #2968 so they're more visible.

simsong:
@uckelman-sf  - Can you tell me if a .E01 volume was scanned as part of the acceptance tests for this PR?

uckelman-sf:
The intent is that an E01 be tested. I'm fairly sure from working on #2968 that no E01 is being tested right now, either on this branch or on the develop branch.

uckelman-sf:
The branch for #2968 has the commits here and is testing an E01. E.g., see this run: https://github.com/sleuthkit/sleuthkit/actions/runs/11129960992/job/30928260322?pr=2968

simsong:
Well, the fiwalk code that I pushed and committed to develop is testing an E01, and that's how I found out that the libewf-dev wasn't included as part of the ubuntu runner. And then I added it.

Successful run: https://github.com/sleuthkit/sleuthkit/actions/runs/11095275800

Failed run: https://github.com/sleuthkit/sleuthkit/actions/runs/11094219708

What do you think? Am I misdiagnosing what happened?

simsong:
Actually, this is the failed run: https://github.com/sleuthkit/sleuthkit/actions/runs/11094170824/job/30821343267#step:14:1095

simsong:
I find these outputs very hard to interpret.
image
image

Compared to the eeasy-to-understand dfxml output.

I also prefer using DFXML for the test cases because I can then test off DFXML code on digitalcorpora.org.

uckelman-sf:
The link I provided is a successful run for 64-bit shared Linux builds:
image
The intent was to demonstrate that this PR works correctly. I can't confirm that fiwalk is correctly testing E01s myself on develop at present. If you think it is, then builds for this PR also show that this PR works correctly.

simsong:
Right. It builds. But does it work? I want to see that it's reading an E01 file and that TSK still works.

uckelman-sf:
I believe the code in this PR works because we've been using the code in this PR for years.

I don't believe that E01s are being tested on develop right now. However, if you do believe that E01s are being tested, then you should also believe the code works, because the tests pass.

simsong:
When get_test_images.py was being run, and then runner was being run, one of the images that was fetched was an E01 image. I know that the image was being scanned because in https://github.com/sleuthkit/sleuthkit/actions/runs/11094170824/job/30821343267#step:14:1095 the `unit_tests/runner` program was being linked without `libewf-dev` and the program was failing to process `imageformat_mmls_1.E01`:

```
  - step1:
      image: https://digitalcorpora.s3.amazonaws.com/corpora/drives/dftt-2004/imageformat_mmls_1.E01
      xml:   https://digitalcorpora.s3.amazonaws.com/corpora/drives/dftt-2004/imageformat_mmls_1.E01.xml
```
The reason why I want to use DFXML to validate the tests is that the DFXML files are stored on digitalcorpora. These  are useful to other people for other reasons — for example, @ajnelson-nist makes use of the DFXML files. He has no interest in the `img_dump` output.

Since the code has been working for you, I'll accept it. I'm happy to call the runner with fiwalk an acceptance test rather than a unit test, but I really do want to run fiwalk multiple times from a single execution against multiple disk images.  I think that this is a better approach moving forward, and it's more likely to be resilient to internal changes in TSK over the years. A JSON dump won't be resilient against adding more information to the JSON format.

uckelman-sf:
```
0158:err:msvcrt:_invalid_parameter (null):0 (null): (null) 0
```
isn't coming from `img_dump` itself, but rather from `wine`. I aim to run down what that is in time.

simsong:
> ```
> 0158:err:msvcrt:_invalid_parameter (null):0 (null): (null) 0
> ```
> 
> isn't coming from `img_dump` itself, but rather from `wine`. I aim to run down what that is in time.
> 
> I will reply to the remainder of what you wrote above, but am not that far along yet.

Oh, if that's the case, just put a `grep -v` in to get rid of this.

PR #2997, opened 2024-10-01
Fix file permissions

uckelman-sf:
Non-executable files shouldn't have the executable bit set.


uckelman-sf:
@simsong Ready for review.

PR #3002, opened 2024-10-03
Build with libqcow, libvhdi, libvmdk in GitHub Actions

uckelman-sf:


uckelman-sf:
@simsong Ready for review.

PR #3003, opened 2024-10-03
Break out of infinite loop

### LINKED ISSUE #2268: Infinite loop when processing a Linux disk image (tsk_fs_file_attr_getsize)
Hello everyone,

We use Sleuthkit 4.10.2 to process disk images. One of our customers reported that the app freezes when processing a disk image with Linux filesystem (ext2). We don't have direct access to the disk image, but we asked the customer to capture a process dump.

So, here is where it is getting stuck:
```
libtsk-x64.dll!tsk_fs_attr_append_run(TSK_FS_INFO * a_fs=0x0000000000000000, TSK_FS_ATTR * a_fs_attr=0x000000b432e84570, TSK_FS_ATTR_RUN * a_data_run=0x000000b43b97d520) Line 728
libtsk-x64.dll!unix_make_data_run_direct(TSK_FS_INFO * fs=0x000000b4225562e0, TSK_FS_ATTR * fs_attr=0x000000b432e84570, unsigned __int64 * addrs, unsigned __int64 addr_len=1024, __int64 length=-3177641115648) Line 91
libtsk-x64.dll!unix_make_data_run_indirect(TSK_FS_INFO * fs=0x000000b4225562e0, TSK_FS_ATTR * fs_attr=0x000000b432e84570, TSK_FS_ATTR * fs_attr_indir=0x000000b432e83df0, char * * buf=0x000000b420ceddc8, int level=1, unsigned __int64 addr=0, __int64 length=-3177641115648) Line 215
libtsk-x64.dll!unix_make_data_run_indirect(TSK_FS_INFO * fs=0x000000b4225562e0, TSK_FS_ATTR * fs_attr=0x000000b432e84570, TSK_FS_ATTR * fs_attr_indir=0x000000b432e83df0, char * * buf=0x000000b420ceddc8, int level=2, unsigned __int64 addr, __int64 length=-3173505531904) Line 226
libtsk-x64.dll!unix_make_data_run_indirect(TSK_FS_INFO * fs=0x000000b4225562e0, TSK_FS_ATTR * fs_attr=0x000000b432e84570, TSK_FS_ATTR * fs_attr_indir=0x000000b432e83df0, char * * buf=0x000000b420ceddc8, int level=3, unsigned __int64 addr, __int64 length=475295744) Line 226
libtsk-x64.dll!tsk_fs_unix_make_data_run(TSK_FS_FILE * fs_file) Line 413
libtsk-x64.dll!tsk_fs_file_attr_check(TSK_FS_FILE * a_fs_file, char * a_func=0x000007f8a7309be0) Line 235
libtsk-x64.dll!tsk_fs_file_attr_getsize(TSK_FS_FILE * a_fs_file=0x000000b41f66a0c0) Line 251
[Frames may be missing, no binary loaded for bridj.dll]
bridj.dll!000007f8d4fc994e()
```

So, our app is trying to read attributes for a specific file by calling:
`tsk_fs_file_attr_getsize`

And this method never returns. According to the stack trace, it might be stuck in this loop, fs_attr.c line 728:
```
data_run_cur = a_fs_attr->nrd.run;
while (data_run_cur->next) {
   data_run_cur = data_run_cur->next;   // <-- stuck here
}
```

I understand that this might be some kind of file system corruption. But the customer claims that he was able to process this disk image with other tools.

Is there any fix possible in this situation?

simsong:
Closes #2268

(Also removes some cruft at the end of lines. I guess we should do this globally.)


simsong:
Waiting for comment from @igorrogov if this fixes noted problem.

uckelman-sf:
Where's the test case for this?

simsong:
We never got it, but it was reported by the OP that the patch worked.

uckelman-sf:
We have know way of knowing now if this actually fixed a bug. What if an adversary tried to introduce a flaw this way?

simsong:
I looked at the code in detailed. If it loops more than a million times, something is wrong. Take a look at the change. It isn’t big.

APriestman:
I realize this is old, but why is this defined:
#define LOGICAL_MAX_ATTR_RUN 0x7fffffff // see fs_attr.c
in the logical file code? It doesn't even run on linux or on disk images (which was the test case described)

simsong:
someone reported an infinite loop. I looked at the code and yes, a corrupt disk image would result in an infinite loop. We should never have infinite loops. So I added this code.

uckelman-sf:
`LOGICAL_MAX_ATTR_RUN` can't be the correct loop limit. It's not related to the contents of fs_attr.cpp

simsong:
> `LOGICAL_MAX_ATTR_RUN` can't be the correct loop limit. It's not related to the contents of fs_attr.cpp

Do you think hat we are likely to see more than 0x7fffffff data runs?
What would you suggest instead?

uckelman-sf:
I'm not commenting on the _value_ of the define, but rather that the define chosen is semantically unrelated to the thing you're testing for. If you want `0x7fffffff`, make it that instead of coopting an unrelated define which could get changed to something else.

uckelman-sf:
Would you explain what the connection is between this bug and the logical image type, which is what `tsk_logical_fs.h` deals with? I don't see it at present.

simsong:
We had a report of an infinite loop in that section of code.
We could not get a copy of the disk image.
I believed that the data_run_cur->next pointer came from the disk and pointed to another data_run_cur.
My fear was that there was a circular data structure on the disk in question, like this:
￼

simsong:
> But what does that have to do with the logical image type? Nothing in `tsk_logical_fs.h` should be used when processing a Linux image.

Oh, okay. I screwed up and put the `#define` in the wrong file. I can move it....

APriestman:
Yes I was only commenting on the location. It 100% does not belong in the logical file system code.

simsong:
> Yes I was only commenting on the location. It 100% does not belong in the logical file system code.

Okay. I'll move it.

Issue #2911, opened 2024-04-14
cc1plus: warning: unrecognized command line option ‘-Wno-unused-command-line-argument’


joachimmetz:
While working on some CI tests https://github.com/sleuthkit/sleuthkit/pull/2428 I noticed https://ci.appveyor.com/project/bcarrier/sleuthkit/builds/49610863/job/tag39mo5c48bxwof#L1091 looks like ‘-Wno-unused-command-line-argument’ is not supported by `g++`

```
cc1plus: warning: unrecognized command line option ‘-Wno-unused-command-line-argument’
```

A later version of `g++`

```
At global scope:
cc1plus: note: unrecognized command-line option ‘-Wno-unused-command-line-argument’ may have been intended to silence earlier diagnostics
```

joachimmetz:
I also recommend setting `APPVEYOR_BUILD_WORKER_IMAGE: ubuntu2204` also see https://www.appveyor.com/docs/linux-images-software/ given that the Ubuntu image is using 18.04

joachimmetz:
I see I had reported this previously as well https://github.com/sleuthkit/sleuthkit/issues/2747

uckelman:
I aim to do something about this in the coming week.

joachimmetz:
This still persist with clang

```
At global scope:
cc1plus: note: unrecognized command-line option ‘-Wno-unused-command-line-argument’ may have been intended to silence earlier diagnostics
```

Would recommend to add a clang CI test

uckelman:
I'd be happy to review a PR for adding clang to CI.

simsong:
And I’m looking forward to PRs that eliminate more of our compiler warnings

joachimmetz:
Then maybe reopen this issue first and merge/review some of the remaining pending PRs as well https://github.com/sleuthkit/sleuthkit/pulls?q=is%3Apr+is%3Aopen+joachimmetz

My proposal would be to remove setting `Wno-unused-command-line-argument`, happy to rip it out and send you a PR if this is the direction you want to take it?

joachimmetz:
PR for a Linux clang CI test https://github.com/sleuthkit/sleuthkit/pull/3013

uckelman:
I thought this part of `configure.ac` would be sufficient to suppress `-Wno-unused-command-line-argument` when the compiler doesn't understand it:
```
dnl Use -Wno-unused-command-line-argument only if the compiler understands it
[...]

PR #3013, opened 2024-10-05
Added Linux clang build target for premature closed issue #2911

joachimmetz:


uckelman:
This appears to switch Linux builds to clang, so that there are no longer gcc builds.

.github/workflows/build-unix.yml
```diff
@@ -35,6 +35,14 @@ jobs:
             configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java"
             codecov: "no"
 
+          - os: "linux"
+            arch: "x86_64"
+            compiler: "clang"
+            linkage: "shared"
+            runner: "ubuntu-22.04"
+            configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java CC=clang"
```
simsong, line 43: Don't we also want `CXX=clang++` ?

simsong (changes requested):
shouldn't we also specify CXX?

simsong:
> This appears to switch Linux builds to clang, so that there are no longer gcc builds.

I think that we want two builds on linux - one with clang and one with gcc. But we already have two builds - one with shared libraries, one without. So the question is do we need two or three?

joachimmetz:
Gents, given the size of this PR, it might be better to just edit it

joachimmetz:
Also note that this PR does not fix #2911, it just adds a clang test as requested on the issue.

simsong:
> Gents, given the size of this PR, it might be better to just edit it, than going back and forth over comments.
> 
> I can only guess about your intended test strategy given this is not documented. My recommendation would be do the same tests with both clang as with gcc or at minimum have 1 clang build test

I like to document things in PRs...

PR #3021, opened 2024-10-06
Adds fiwalk back to codecov

simsong:
Also:
* moves disk image directory from `tests/from_brian` back to `$HOME/from_brian`
* replaces manual run of runner with `make check` for codecov, as ChatGPT promises me that this will work (see https://chatgpt.com/share/670270fa-3c60-8010-b706-60a620607982).


simsong:
Getting closer. Now I just need to fix the mingw problem:
image

PR #3025, opened 2024-10-07
added two fat test images

simsong:


simsong:
The first check has hung in that place they are all hanging:
image

simsong:
And this shows what happens when I added that disk image:
https://app.codecov.io/github/sleuthkit/sleuthkit/commit/29af3d3db69782d60c62293eb4e663cd95c1ab57/indirect-changes
image

PR #3030, opened 2024-10-08
Set -Wno-unused-command-line-argument in CI configuration not in configure #2911

joachimmetz:


configure.ac
```diff
@@ -285,10 +285,6 @@ AC_SUBST([CXXLD])
 dnl Dependencies for fiwalk
 AC_CHECK_FUNCS([getline])
 
-dnl Use -Wno-unused-command-line-argument only if the compiler understands it
-AX_APPEND_COMPILE_FLAGS([-Wno-unused-command-line-argument], [CFLAGS], [-Werror-unknown-warning])
-AX_APPEND_COMPILE_FLAGS([-Wno-unused-command-line-argument], [CXXFLAGS], [-Werror-unknown-warning])
-
```
simsong, line 291: This seems to be the right approach here, unless you want to test to see if you can use the command line argument.  I have a macro that does this:  https://github.com/simsong/be20_api/blob/main/m4/slg_gcc_all_warnings.m4

joachimmetz: this currently macros do not do the right thing, see previous comments in the corresponding issue. I'll have a look at the alternative approach when time permits.

PR #3031, opened 2024-10-08
added tests for HFS journaled image

simsong:


simsong:
Hm. Looks like HFS was already being tested. I'm going to leave this in, however, as this disk image also checks the HFS journal. We don't have a test case for that now, however.

PR #3034, opened 2024-10-08
cleaned up github action.

simsong:
Not sure why we want to run make check wit and without address sanitizer.
Perhaps @uckelman-sf  knows?

### DIFF
--- a/.github/workflows/build-unix.yml
+++ b/.github/workflows/build-unix.yml
@@ -115,7 +115,7 @@ jobs:
           echo "Using $CORES cores"
 
       - name: Install MacOS packages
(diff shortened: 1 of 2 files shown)


simsong:
yay! it works.

uckelman-sf:
We want to run both ways because they could have different behavior.

simsong:
> We want to run both ways because they could have different behavior.

Really? I've never seen that.

uckelman-sf:
You won't ever see it if you don't run both.

Each has a reason to be run: The sanitizer build finds bugs. The non-sanitizer build is a production build, which is exactly the thing we want to pass the tests.

PR #3035, opened 2024-10-09
Fix compiler warnings

uckelman-sf:


tools/autotools/tsk_comparedir.cpp
```diff
@@ -165,18 +165,18 @@ uint8_t
     char fullPath[TSK_CD_BUFSIZE];
     struct stat status;
 
-    strncpy(fullPath, m_lclDir, TSK_CD_BUFSIZE);
+    strncpy(fullPath, m_lclDir, TSK_CD_BUFSIZE - 1);
```
simsong, line 168: It's 2024 and strncpy *still* doesn't guarantee null termination. 
Do you need to add this:
```
fullPath[TSK_CD_BUFSIZE-1]='\0';
```
?

simsong: (And there are a lot more like this.)

uckelman-sf: I'm going to hold off on the bounded string fixes until later. I've removed those from this PR.

uckelman-sf:
@simsong AppVeyor seems not to be working.

simsong:
@uckelman-sf Here is the error in appveyor:
```
tsk/util/crypto.cpp:196:43: error: ‘EVP_MD_CTX_get_size’ was not declared in this scope; did you mean ‘EVP_MD_CTX_size’?
  196 |   auto hash = std::make_unique<uint8_t[]>(EVP_MD_CTX_get_size(ctx.get()));
      |                                           ^~~~~~~~~~~~~~~~~~~
      |                                           EVP_MD_CTX_size
```

simsong:
Now, the big question — why did this trigger in appveyor but not in our tests?

uckelman-sf:
My guess is that AppVeyor is using an older version of OpenSSL. I'll run this down on Monday.

simsong:
> My guess is that AppVeyor is using an older version of OpenSSL. I'll run this down on Monday.

That's quite possible. We should probably work with both. Thanks again for all of your help on this.
Personally, I need to figure out how to make codecov show coverage changes on PRs like this one, where the HEAD is not in this repo. That's what's causing the problem. Thoughts?

uckelman-sf:
I fixed the problem; apparently `EVP_MD_CTX_get_size` has an older, but not yet deprecated name of `EVP_MD_CTX_size`, which works.

As for codecov with forks, this is what I turned up in a minute of looking:

https://github.com/codecov/codecov-action/issues/29
https://github.com/codecov/codecov-action/issues/22

I haven't read through it all yet.

simsong:
Thanks for the codecov pointers.  I'll check them out.

Issue #3029, opened 2024-10-08
configure: disable javac if not available


joachimmetz:
Currently configure fails with `configure: error: no acceptable Java compiler found in $PATH` why is it a requirement to have to pass `--disable-java` ? Can the default behavior of configure be changed to just continue without Java given this is possible with `--disable-java`

joachimmetz:
Also https://github.com/sleuthkit/sleuthkit/blob/develop/configure.ac#L85 says `Add enable/disable option` but configure only provides `--disable-java          Do not build the java bindings or jar file`

uckelman-sf:
Closed by PR #3038.

simsong:
@brianc - I think that the default should be not to build java unless requested, since most TSK users probably do not require the Java bindings. 

As of #3038, you need to do `--enable-java` to enable it. I think that's fine?

PR #3038, opened 2024-10-12
Changed disable-java to enable-java #3029

joachimmetz:


uckelman-sf:
When building with `--enable-java`, it's correct to abort when `javac` isn't available.

simsong:
@joachimmetz  - are you okay with this?

joachimmetz:
> When building with --enable-java, it's correct to abort when javac isn't available. What is this change intended to do?

there is no `--enable-java` option at the moment, it is implicitly assumed that JNI/Java support should be built. This PR changes configure to not abort if there is no javac

having a `--enable-java` (maybe `--enable-jni` would be better to disambiguate with SleuthKit Java framework)

uckelman-sf:
> there is no --enable-java option at the moment, it is implicitly assumed that JNI/Java support should be built.

I don't understand on what basis you're saying this. From what I can see, the option is defined in `configure.ac`:
```
AC_ARG_ENABLE([java],
    [AS_HELP_STRING([--disable-java], [Do not build the java bindings or jar file])])
```

joachimmetz:
```
./configure --help | grep java
  --disable-java          Do not build the java bindings or jar file
```

This is currently an implicit option

uckelman-sf:
What do you mean by "implicit option"?

joachimmetz:
that is not specifically called out in the help

uckelman-sf:
Whether the option appears in the help doesn't affect the behavior of the option.

joachimmetz:
> Whether the option appears in the help doesn't affect the behavior of the option.

it is currently implicitly enabled, why? this makes things unnecessary convoluted for people not wanting to build the JNI bindings

joachimmetz:
> Would you state clearly what this PR is intended to do?

The PR does exactly say what it does, it prevents configure to abort if javac is not installed

uckelman-sf:
I'm not in favor of the change as stated. I'm not convinced the PR does what it says it does yet.

joachimmetz:
> I'm not in favor of the change as stated. I'm not convinced the PR does what it says it does yet.

why are you thinking this?

uckelman-sf:
I have time at the moment only to address one part of that: I don't think the user should be forced to specify `--disable-java`. The Java build should be disabled by default.

I'll return to the rest later.

joachimmetz:
>  I don't think the user should be forced to specify --disable-java. The Java build should be disabled by default.

then I'm happy to change the PR to match that

uckelman-sf:
Wouldn't these changes be sufficient to make no Java the default?
```diff --git a/configure.ac b/configure.ac
index fe0870c83..b87cc32e3 100644
--- a/configure.ac
+++ b/configure.ac
@@ -84,7 +84,7 @@ AS_IF([test "x$ax_pthread_ok" = "xyes" && test "x$enable_multithreading" != "xno
 
 dnl Add enable/disable option
 AC_ARG_ENABLE([java],
-    [AS_HELP_STRING([--disable-java], [Do not build the java bindings or jar file])])
+    [AS_HELP_STRING([--enable-java], [Build the java bindings or jar file])])
 
 dnl Checks for libraries.
 
@@ -196,7 +196,7 @@ AM_COND_IF([CUSTOM_DEFAULT_JAR_LOCATION],
 )
 
 dnl Test for the various java things that we need for bindings
-AS_IF([test "x$enable_java" != "xno"], [
+AS_IF([test "x$enable_java" = "xyes"], [
     dnl javac is needed to compile the JAR file
     AX_PROG_JAVAC
     if test "x$JAVAC" != x; then
```

joachimmetz:
`if test "x$JAVAC" != x; then` would likely have to be changed to use autoconf macros as well

uckelman-sf:
I've had a chance to read through the changes now. I wouldn't bother with defining `AX_COMMON_ARG_DISABLE`, and would change the test as I mentioned above; otherwise, this seems fine.

joachimmetz:
@uckelman-sf changes update, PTAL

joachimmetz:
> I've had a chance to read through the changes now. I wouldn't bother with defining AX_COMMON_ARG_DISABLE, and would change the test as I mentioned above; otherwise, this seems fine.

the problem is that AC_ARG_ENABLE does not handle caching of options, but this is something for a different PR

simsong (approved):

bcarrier:
So to confirm, the intended new behavior is that you need to opt-in for Java support? Even if all of the libraries exist, you need to supply 'enable-java'? 

If so, this seems inconsistent. Is the plan to also have people opt-in for libewf, afflib, etc. Isn't the whole point of autoconf that it detects what's there and includes it? 

If this is the intended behavior, it also requires documentation changes since it is a backward incompatible change.

bcarrier:
100% agreed. Was configure failing though because javac wasn't detected? That's a bug in the macros and configure.ac if it was. Was it failing before all of the recent changes? I saw a bunch of that stuff moved around. 

At this point, I don't really care which way it goes.  We just need to update our build systems to start using this branch.

joachimmetz:
@bcarrier this is the original issue https://github.com/sleuthkit/sleuthkit/issues/3029#issuecomment-2408418288  from the sound of it you would favor the original proposal of disabling java when no javac and enabling when all requirements are met. 

> What I objected to originally was configure failing when javac wasn't detected and --enable-java wasn't specified. In that situation, you should get a build with Java disabled because you don't have what's required for building with it---as you say.

no you did not you did the opposite, the original proposal was to no fail if javac was not available. The TSK behavior at that time was to fail per https://github.com/sleuthkit/sleuthkit/issues/3029

uckelman:
What should happen is:

* `configure` should give an error if `--enable-java` is given but the requirements aren't met
* `configure` should not build anything with Java if `--disable-java` is given.
* `configure` should not build anything with Java if neither option is specified and the requirements for building with Java aren't met

Regarding configure defaults: It's not unusual for projects to have an assortment of defaults, some enabled, some detected, and some disabled, depending on how likely users are to want them. If Joachim wants that changed, he could make another PR for it, I suppose.

joachimmetz:
> That's not controversial, I think?

sigh you still don't get it do you? in https://github.com/sleuthkit/sleuthkit/issues/3029#issuecomment-2408418288 i flagged that the behavior you claim "should happen" did not happen. My guess it that https://github.com/sleuthkit/sleuthkit/commit/bbad55df332bce3a37ff6cf785361792a631df2e introduced the issue that the build fails when there is no javac

My initial changes were to have "configure should not build anything with Java if neither option is specified and the requirements for building with Java aren't met"

Issue #3039, opened 2024-10-12
configure: Catch2 support shows no status


joachimmetz:
Catch2 support in the configure output is missing a status indicator

```
Building:
   afflib support:                        no
   libaff4 support:                       no
   libbfio support:                       yes
   libewf support:                        no
   libqcow support:                       yes
   libvhdi support:                       no
   libvmdk support:                       no
   libvslvm support:                      no

   openssl support:                       no
   zlib support:                          yes

Features:
   Catch2 support:                        
   Java/JNI support:                      no
   Multithreading:                        yes
```

joachimmetz:
There appears to be no configure test that sets with_catch2 ?

uckelman-sf:
@simsong Closed by #3040.

PR #3040, opened 2024-10-12
Git removed unused Catch2 support output #3039

joachimmetz:


uckelman-sf:
@simsong This looks fine.

Issue #3020, opened 2024-10-06
Static build on Linux fails due to use of dlopen for sqlite


joachimmetz:
Added CI test for static build on Linux https://github.com/sleuthkit/sleuthkit/pull/3019 result:

```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libsqlite3.a(os_unix.o): in function `unixDlOpen':
(.text+0x8ad): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:3624: unit_tests/runner] Error 1
make[1]: Leaving directory '/home/runner/work/sleuthkit/sleuthkit'
make: *** [Makefile:7105: check-am] Error 2
```

Which appears to be called in tsk/auto/sqlite3.c

Recommend bundled sqlite3 not to be build when `--enable-static` is set or add something like `--without-sqlite3`

joachimmetz:
Looks like the current configuration convolutes "static library" and "static binary" builds. 

https://github.com/sleuthkit/sleuthkit/issues/3041#issuecomment-2408867456

joachimmetz:
Some changes to remove the current convolution https://github.com/sleuthkit/sleuthkit/pull/3043

uckelman-sf:
PR #3046 addresses the problem with bundled sqlite3 using dlopen in static builds---but as this is only a warning, it's not what's causing your build to fail.

joachimmetz:
per https://github.com/google/oss-fuzz/actions/runs/11201208381/job/31135856628#step:7:1664 I conclude differently

PR #3043, opened 2024-10-13
Changes for static library build #3020

joachimmetz:


configure.ac
```diff
@@ -128,12 +135,16 @@ AS_IF([test "x$ax_sqlite3" != "xyes"],
   ]
 )
 
-dnl Compile the bundled sqlite if there is no system one installed
-AC_MSG_CHECKING(which sqlite3 to use)
-AS_IF([test "x$ax_sqlite3" = "xyes"],
-      [AC_MSG_RESULT([system])
-       PACKAGE_LIBS_PRIVATE="$PACKAGE_LIBS_PRIVATE -lsqlite3"],
-      [AC_MSG_RESULT([bundled])])
+AS_IF([test "x$static_executables" != "xyes"],
```
uckelman-sf, line 138: Making the fallback sqlite3 check contingent on building static executables isn't correct regardless of anything else in this PR. It's possible that someone is doing a static build but has no static sqlite3 library installed.

joachimmetz: There is no other option to disable sqlite3 as a dependency at this moment.

If that is the preferred route happy to go that way. If not, it does not appear to be directly needed for many of the tools. 

Please come with solutions, not with judgement.

uckelman-sf: Have you looked at PR #3046? That contains a change which turns off the use of dlopen in the bundled sqlite3.

I think you're taking all this too personally.

joachimmetz: No I have not looked at https://github.com/sleuthkit/sleuthkit/pull/3046 given that this is a newer PR. Remove the linux static test, get the PR working and merged and I can update this PR.

> I think you're taking all this too personally.

You are derailing the conversation. This also is a judgement.

uckelman-sf:
I suspect this will cause undesired results when building static tools. I aim to look into substantiating that on Friday.

simsong:
@joachimmetz and @uckelman-sf  — I cleaned up the comments a bit.

simsong:
@uckelman-sf - are you okay with this?

simsong:
> Absolutely opposed at the moment. I need some time to try a few scenarios that I suspect it will fail with.

Okay. Why don't you try your scenarios and let me know how they work out. I'll make this a draft until we know. If you do have some scenarios, let's add them to test cases so that we can explicitly test for them. Thanks!

uckelman-sf:
The first problem I see is that `configure --disable-static --enable-static-executables` is an invalid combination of options, so ought to produce an error, but does not.

uckelman-sf:
This should be updated after #3046 is merged.

simsong:
@uckelman-sf - is this still relevant? I tried to resolve the conflicts but I can't quite figure out what is going on.

simsong:
This doesn't seem relevant anymore, so I'm closing it. @uckelman-sf  and @joachimmetz  - feel free to resolve the conflicts and re-open if this is still relevant.

PR #3045, opened 2024-10-16
Fix Win TZ bug

uckelman-sf:


sf-jonstewart:
@uckelman-sf Can you please describe the issue here? I see the test but not a fix.

test/img_dump/test_imgs.sh
```diff
@@ -3,5 +3,6 @@
 ID=test/img_dump/img_differ.sh
 
 $ID ../data/image.dd dump/image.dd.json
+$ID ../data/fat_dst_test.dd dump/fat_dst_test.dd.json
 # FIXME: check iso on MinGW after fixing timestamp bug
```
atcuno, line 7:
```suggestion
# FIXME: check iso on MinGW after fixing timestamp bug
# Testing new GitHub feature.
```

simsong:
I've done some research on the handling of TZ under mingw. Apparently the variable sometimes ignored, depending on which C runtime is being used. Setting TZ before the program runs sometimes works, but changing it when the program is running mostly does not unless `_tzset()` is called.

I've created some tests for this here:
https://github.com/sleuthkit/sleuthkit/pull/3351/files#diff-8c1bfabcdffc6a7f7fc6510c6fe2b9457aa53a921b159481cb4eb6bd53a843d4

simsong:
I'm moving this from draft to ready status, cleaned up the comment history, and will be accepting this after review by ChatGPT.

PR #3046, opened 2024-10-16
Disable dynamic extension loading in bundled sqlite3

uckelman-sf:
The bundled sqlite3 is currently built with dynamic extension loading turned on. Nothing in Sleuthkit uses dynamic extension loading. Dynamic extension loading uses libdl (which contains dlopen), and that generates compiler warnings during static builds when using the bundled sqlite3.

This PR turns off dynamic extension loading when using the bundled sqlite3.


simsong:
Waiting for AppVeyor...

PR #3051, opened 2024-10-21
Fix MinGW warnings

uckelman-sf:


simsong:
It happened again — https://github.com/sleuthkit/sleuthkit/actions/runs/11445237431/job/31841885981 hung. For some reason, it looks like there is some kind of error in fiwalk that is triggered by address sanitizer that is causing a process or something not to finish, and `make check` is not terminating. Honestly, the easiest way around this may be to run the test program directly, rather than from `make check.` @uckelman-sf  - what do you think?

uckelman-sf:
I expect you'd have the same problem running directly. It's unlikely that `make` itself is what's hanging.

uckelman-sf:
I'll take a look after I've cleared away all the MinGW warnings.

uckelman-sf:
@simsong Ready for review.

uckelman-sf:
@simsong I've moved all of the completed warning fixes from #3052 here and will stop updating this PR, so it can be reviewed.

PR #3053, opened 2024-10-22
Update bundled sqlite3 to 3.47.0

uckelman-sf:
We're bundling a very old version of sqlite3. This makes it current, which has the nice side effect of eliminating many compiler warnings.


uckelman-sf:
@simsong Ready for review.

Issue #3001, opened 2024-10-03
sometimes one of the tests is hanging


simsong:
Thoughts @uckelman-sf ?

https://github.com/sleuthkit/sleuthkit/actions/runs/11150480178/job/30991707724

uckelman-sf:
I don't have any ideas atm. Nothing unusual was logged. Possibly this is a transient problem with the job runner?

simsong:
No idea. Let's see what happens next time.

simsong:
happened again:
https://github.com/sleuthkit/sleuthkit/actions/runs/11172561508/job/31059210985

simsong:
I'm re-running https://github.com/sleuthkit/sleuthkit/actions/runs/11172561508 to see if that solves the problem, and I'm setting the job timeout to 30 min, which should be enough.

simsong:
Here is the output of are recent run in which address-sanitizer failed:
https://github.com/sleuthkit/sleuthkit/actions/runs/11196499434/job/31125415741
![image](https://github.com/user-attachments/assets/bc79238c-8ac5-44fc-ab87-2f18ddb1019e)

And here is the output where it worked:
https://github.com/sleuthkit/sleuthkit/actions/runs/11195090186/job/31122341504
![image](https://github.com/user-attachments/assets/3250ebfe-dc77-4edf-91e9-88ddb522e0c7)

@uckelman-sf  - This doesn't make sense to me. What's happening after the `PASS: tests/fiwalk_test` runs that could sometimes hang?

joachimmetz:
Maybe related https://github.com/joachimmetz/sleuthkit/actions/runs/11201164749/job/31135761436#step:12:2865 could be a network connectivity issue

simsong:
https://github.com/sleuthkit/sleuthkit/issues/3022

simsong:
Still, I wonder what's failing. What happens after it prints the last `PASS` before it prints `Testsuite summary` ?

joachimmetz:
asan can slow things significantly as well, not sure how long a local run takes, also realize that the CI test might run with different conditions https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners

simsong:
> asan can slow things significantly as well, not sure how long a local run takes, also realize that the CI test might run with different conditions https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners

[...]

PR #3062, opened 2024-10-25
Fix deadlock in error test

uckelman-sf:
The main thread incremented the semaphore, then decremented the semaphore. Most of the time the child thread got to the semaphore between those, and everything ran as intended. Once in a while, the child thread lost the race, resulting in a deadlock.


uckelman-sf:
Closes #3001.

uckelman-sf:
@simsong Ready for review.

simsong (approved):

PR #413, opened 2015-03-27
Btrfs support

basicmaster:
As part of a practical course at my university, I developed Btrfs support for TSK and would like to contribute it. There are some restrictions of the support which I mentioned in the commit message.

I already did a rough indent to the code and hope it is enough.
Regarding my design decisions I already wrote some kind of implementation notes which I then will add to the wiki. The notes are ATM in German but I will translate them into English.
Soon I will also post a link to a Btrfs test image here.

Issue #3044, opened 2024-10-14
Put test images in submodule


uckelman-sf:
Presently, the test images which are used by the fiwalk tests are fetched using `get_test_images.py`. It would require no code at all if those test images were in a submodule instead, as they'd be checked out at the same time as the main repo (if desired).

@simsong Should we create a subrepo in which to store test images?

simsong:
I've thought about that. What's your thought on the disk space requirements?

uckelman-sf:
The entire repo is about 370MB. The test images which are downloaded to `tests/from_brian` are 35MB. I think there's no problem presently with disk space.

simsong:
Okay. I’ll do it!

uckelman-sf:
I thought that creating a repo for test images would be a 5-minute job. Has something gone wrong?

simsong:
I have had a lot of other things to do. The repo is created, but not populated. I can do that today or tomorrow.

simsong:
i'm not going to get to this tonight. Perhaps tomorrow.

simsong:
https://github.com/sleuthkit/sleuthkit_test_data

simsong:
I have added the images on my local machine, but don't have write access to the repo yet. Either @bcarrier  or @jayaramcs needs to fix this.

bcarrier:
Sorry. Fixed.

simsong:
```
simsong@Simsons-MacBook-Pro sleuthkit_test_data % git push origin
Uploading LFS objects: 100% (10/10), 15 MB | 675 KB/s, done.
Enumerating objects: 16, done.
Counting objects: 100% (16/16), done.
Delta compression using up to 10 threads
Compressing objects: 100% (14/14), done.
Writing objects: 100% (14/14), 2.21 KiB | 2.21 MiB/s, done.
Total 14 (delta 0), reused 0 (delta 0), pack-reused 0
To ssh://github.com/sleuthkit/sleuthkit_test_data
   0bf9304..03d276b  main -> main
simsong@Simsons-MacBook-Pro sleuthkit_test_data %
```
@uckelman-sf will be happy.

PR #3065, opened 2024-10-26
BTRFS from basic master

simsong:
This is https://github.com/sleuthkit/sleuthkit/pull/413 but brought up to develop. 
- [x] Add test case for btrfs
- [x] Make sure any other code changed is also tested.


simsong:
Still waiting for AppVeyor, but this now works on windows, linux and MacOS. It would be great for @bcarrier and @jayaramcs to take a look, as this is a new file system for TSK. That's pretty significant! @joachimmetz and @uckelman-sf are also invited to comment.

simsong:
The AppVeyor fails were because:
* PyYAML didn't build
* AppVeyor doesn't have libewf installed.

joachimmetz:
Nice, would recommend adding a fuzz target as well https://github.com/sleuthkit/sleuthkit/tree/develop/ossfuzz

simsong:
I don't know how to add the fuzz targets, but if you tell me how, I'll do it!

joachimmetz:
> I don't know how to add the fuzz targets, but if you tell me how, I'll do it!

sure, should not be too involved, let me ask a couple of questions first to get some more context

* is there a small test file publicly available that can be used as seed? for https://github.com/sleuthkit/sleuthkit/blob/develop/ossfuzz/buildcorpus.sh
* does the btrfs back end require the pool layer like apfs or is it more comparable to the ext backend?

simsong:
> > I don't know how to add the fuzz targets, but if you tell me how, I'll do it!
> 
> sure, should not be too involved, let me ask a couple of questions first to get some more context
> 
> * is there a small test file publicly available that can be used as seed? for https://github.com/sleuthkit/sleuthkit/blob/develop/ossfuzz/buildcorpus.sh
> * does the btrfs back end require the pool layer like apfs or is it more comparable to the ext backend?

How about https://corp.digitalcorpora.org/corpora/drives/2015-basicmaster/btrfs_testimage_50MB.E01?

btrfs sits on top of raw storage, like ext.

joachimmetz:
> How about https://corp.digitalcorpora.org/corpora/drives/2015-basicmaster/btrfs_testimage_50MB.E01?

to isolate just fuzzing the btrfs backend a smaller only btrfs raw file system is needed, https://github.com/dfirlabs/btrfs-specimens can likely. also note that the ossfuzz build does not build with E01 support

> btrfs sits on top of raw storage, like ext.

the fls fuzz target likely can be reused then https://github.com/sleuthkit/sleuthkit/blob/develop/ossfuzz/fls_fuzzer.cc
and only an update to the Makefile is needed https://github.com/sleuthkit/sleuthkit/blob/develop/Makefile.am#L825 comparable to

```
fls_btrfs_fuzzer_SOURCES = \
	ossfuzz/fls_fuzzer.cc \
	ossfuzz/mem_img.h

fls_btrfs_fuzzer_CPPFLAGS = \
	-DFSTYPE=TSK_FS_TYPE_BTRFS_DETECT \
	$(AM_CPPFLAGS)
```

joachimmetz:
> let me see if I can create a script and add it to https://github.com/dfirlabs/btrfs-specimens

Looks like the minimum required size is 114294784, that might be too large. Note that the seed file is not required but helps with more targeted fuzzing.

simsong:
I’m unclear what the seed file does if it is not needed. Instead of making
the seed files once and downloading it, why not make it as part of the
fuzzing operation?

joachimmetz:
> I’m unclear what the seed file does if it is not needed. 

tldr it helps with improving test coverage, maybe useful context https://google.github.io/oss-fuzz/getting-started/new-project-guide/#seed-corpus

> why not make it as part of the fuzzing operation?

not sure what you specifically are asking with this, but what the fuzzer does is invoke the fuzz targets with mutated input based on the seed corpus

simsong:
Since the fuzzer is running on a Linux system, instead of downloading the seed, perhaps it could just run the script to create the seed.

joachimmetz:
The seed file is not required, adding the fuzz target should be sufficient for now. Seed file can always be added later as well.

simsong:
I clearly don’t understand how the fuzzer works. Doesn’t it need an
executable that it runs? So doesn’t it need a data file to run with the
executable? Can you point me at academic paper that describes how it works?
Thanks..

simsong:
> I'm unaware of academic paper specifically but the project documentation can be found here https://google.github.io/oss-fuzz/

Thanks. I wonder how hard it would be to get it to work with bulk_extractor.
> 
> The proposed changes to Makefile.am (in [#3065 (comment)](https://github.com/sleuthkit/sleuthkit/pull/3065#issuecomment-2443224669)) and adding `fls_btrfs_fuzzer` to the corresponding `noinst_PROGRAMS` in should be sufficient to get a btrfs fuzz target binary to build and picked up by OSSFuzz.

I'll look into it and see if I can add it to this PR. Meanwhile, I'm during this BT back to a draft until I hear from @jayaramcs.

simsong:
> 
> ```
> fls_btrfs_fuzzer_SOURCES = \
> 	ossfuzz/fls_fuzzer.cc \
> 	ossfuzz/mem_img.h
> 
> fls_btrfs_fuzzer_CPPFLAGS = \
> 	-DFSTYPE=TSK_FS_TYPE_BTRFS_DETECT \
> 	$(AM_CPPFLAGS)
> 
> fls_btrfs_fuzzer_LDADD = \
> 	@LIB_FUZZING_ENGINE@ \
> 	$(TSK_LIB)
> ```

Oh, I get it.  This compiles a fls_btrfs_fuzzer executable with the fuzzing engine linked in and `LLVMFuzzerTestOneInput` as the test harness.  But the fuzzing doesn't happen in the GitHub action, it happens over at Google in GCP. That's why the size is an issue. 

I'm still unclear on where the code analysis is happening. Is the fuzz happening with source code inspection or object code inspection? This is presumably data mutation to cause more execution graph exploration, right?

joachimmetz:
> I'm still unclear on where the code analysis is happening. Is the fuzz happening with source code inspection or object code inspection?

It is configured to run on clusterfuzz. Also see https://google.github.io/oss-fuzz/further-reading/clusterfuzz/ but you can also run it locally.

Issues are available through https://issues.oss-fuzz.com/ Brian currently has access, given he is configure as project owner in https://github.com/google/oss-fuzz/tree/master/projects/sleuthkit

joachimmetz:
Regarding https://github.com/sleuthkit/sleuthkit/commit/1f84b56ff1eb44d330bef08bb8d80ddfed33b5c8#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4afR807

you also need to add `fls_btrfs_fuzzer` to `noinst_PROGRAMS` on line 805

uckelman:
There should be a test image for validating how Sleuthkit reads BTRFS, to go with the other test images in #3044. Has there been any progress on that?

simsong:
> There should be a test image for validating how Sleuthkit reads BTRFS, to go with the other test images in #3044. Has there been any progress on that?

Yes. We have one here:
https://corp.digitalcorpora.org/corpora/drives/2015-basicmaster/btrfs_testimage_50MB.E01

My understanding is that @joachimmetz is concerned about the size.

simsong:
How can I get access to the issues?

tsk/fs/Makefile.am
```diff
@@ -0,0 +1,27 @@
+AM_CPPFLAGS = -I../.. -I$(srcdir)/../.. -Wall
```
uckelman, line 1: New makefiles shouldn't be created in subdirectories. This should go into the main (and only) `Makefile.am`.

simsong: Legacy cruft from the original PR. It's not being used. I'll delete it. Thanks for catching it.

simsong:
It's unclear to me why the AppVeryor failed, so I'm doing an update branch and we will see if it fails again.

Makefile.am
```diff
@@ -701,9 +704,6 @@ endif
 tsk_jar = $(top_builddir)/bindings/java/dist/sleuthkit-$(PACKAGE_VERSION).jar
 jardir = $(prefix)/share/java
 
-$(tsk_jar):
-	cd bindings/java ; ant dist $(ant_args)
-
```
uckelman, line 706: Why is this being removed?

tsk/fs/btrfs_csum.c
```diff
@@ -0,0 +1,43 @@
...
+** Copyright (c) 2015 Stefan Pöschel.  All rights reserved
+**
+** This software is distributed under the Common Public License 1.0
+*/
+
+/*
+ * Contains the checksum part for Btrfs file system support.
+ *
+ * This is outsourced to plain C due to the CRC API defining its own bool type
+ *  - which collides with the bool of C++
```
uckelman, line 17: This is outdated as of PR #3071, which removes that `bool` typedef.

uckelman:
I've reviewed everything except btrfs.c and tsk_btrfs.h now.

I don't understand the objection to the test image being too large. Too large for what?

tsk/fs/btrfs_csum.c
simsong: Okay. So what should I do?

simsong:
> I've reviewed everything except btrfs.c and tsk_btrfs.h now.
> 
> I don't understand the objection to the test image being too large. Too large for what?

My understanding is that @joachimmetz indicated that the test image is too large for the Google fuzzer.

Makefile.am
simsong: user error. I'll add it back.

uckelman:
The test image isn't too large for us to use in CI, is it?

simsong:
Hm...
```
Makefile.am:828: warning: variable 'fls_btrfs_fuzzer_SOURCES' is defined but no program or
Makefile.am:828: library has 'fls_btrfs_fuzzer' as canonical name (possible typo)
Makefile.am:836: warning: variable 'fls_btrfs_fuzzer_LDADD' is defined but no program or
Makefile.am:836: library has 'fls_btrfs_fuzzer' as canonical name (possible typo)
```

tsk/fs/btrfs_csum.c
uckelman: Delete the comment, I think.

simsong:
We don’t have any problems in CI.

tsk/fs/btrfs.cpp
```diff
@@ -1262,7 +1262,8 @@ static bool
 btrfs_treenode_push(BTRFS_INFO * a_btrfs, BTRFS_TREENODE ** a_node,
     const TSK_DADDR_T a_address, const BTRFS_DIRECTION a_initial_index)
 {
-    uint8_t raw[a_btrfs->sb->nodesize];
+    // was: uint8_t raw[a_btrfs->sb->nodesize];
+    uint8_t *raw = new uint8_t[a_btrfs->sb->nodesize];
```
uckelman: Put it in a `std::unique_ptr`. Then you don't need the `delete`s and you can't leak it.

uckelman:
> Hm...
> 
> ```
> Makefile.am:828: warning: variable 'fls_btrfs_fuzzer_SOURCES' is defined but no program or
> Makefile.am:828: library has 'fls_btrfs_fuzzer' as canonical name (possible typo)
> Makefile.am:836: warning: variable 'fls_btrfs_fuzzer_LDADD' is defined but no program or
> Makefile.am:836: library has 'fls_btrfs_fuzzer' as canonical name (possible typo)
> ```

`fls_btrfs_fuzzer` is missing from `noinst_PROGRAMS`.

simsong:
> > Hm...
> > ```
> > Makefile.am:828: warning: variable 'fls_btrfs_fuzzer_SOURCES' is defined but no program or
> > Makefile.am:828: library has 'fls_btrfs_fuzzer' as canonical name (possible typo)
> > Makefile.am:836: warning: variable 'fls_btrfs_fuzzer_LDADD' is defined but no program or
> > Makefile.am:836: library has 'fls_btrfs_fuzzer' as canonical name (possible typo)
> > ```
> 
> `fls_btrfs_fuzzer` is missing from `noinst_PROGRAMS`.

Yep. Found that one. Thanks.

tsk/fs/btrfs.cpp
uckelman: Do this:

```
std::unique_ptr<uint8_t[]> raw{new uint8_t[a_btrfs->sb->nodesize]};
```

After that, you need to access the pointer using `raw.get()`, but otherwise you should need no other changes. The allocation will be cleaned up when `std::unique_ptr` goes out of scope.

We should eventually replace every `malloc` and bare `new` for with `std::unique_ptr` (or `std::string`, if it's a string) for dynamic memory that's scope constrained.

simsong:
Clearly working on two branches at the same time in a single repo is a mistake.

simsong:
@uckelman-sf - AppVeyor is failing with:
image
Any clue why? 
@jayaramcs ?

uckelman-sf:
I don't know what that Python problem is, but it's not why AppVeyor is failing.

This is why AppVeyor is failing:
```
-------------------------------------------------------------------------------
test_disk_images btrfs_test_image.E01 btrfs
-------------------------------------------------------------------------------
test/fiwalk/fiwalk_test.cpp:130
...............................................................................
test/fiwalk/fiwalk_test.cpp:48: FAILED:
  CHECK( o.file_count > 0 )
with expansion:
  0 > 0
with messages:
  img_path := "test/from_brian/btrfs_testimage_50MB.E01"
  test: fiwalk test/from_brian/btrfs_testimage_50MB.E01
0.363 s: test_disk_images btrfs_test_image.E01 btrfs
===============================================================================
test cases: 10 |  9 passed | 1 failed
assertions: 20 | 19 passed | 1 failed
FAIL test/fiwalk/fiwalk_test (exit status: 1)
```

simsong:
OK. Why does that work on GitHub but not in AppVeyor?

uckelman-sf:
It looks like the file count isn't being set? I'm not that familiar with fiwalk to know where that comes from.

simsong:
> It looks like the file count isn't being set? I'm not that familiar with fiwalk to know where that comes from.

If fiwalk doesn't find any files, then this is because the file system isn't finding any. Which means that the file system is working when compiled on GitHub but not when compiled on AppVeyor. I didn't write this code. My suspicion, though, is that there is some structure that is declared that is reading directly from the disk image and the AppVeyor MSC may be different than the GitHub MSC.

.github/workflows/build-unix.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             linkage: "shared"
             compiler: "gcc"
             runner: "ubuntu-22.04"
-            configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java CFLAGS=\"-Wno-unused-command-line-argument -Werror\" CXXFLAGS=-Werror"
+            configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java CFLAGS=\"-Wno-unused-command-line-argument -Werror -Wno-error=unused-parameter\" CXXFLAGS=\"-Werror -Wno-error=unused-parameter\""
```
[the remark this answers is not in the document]
uckelman-sf: This is still not fixed.

.github/workflows/build-unix.yml
```diff
@@ -46,7 +46,7 @@ jobs:
             linkage: "shared"
             compiler: "clang"
             runner: "ubuntu-22.04"
-            configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java CC=clang CXX=clang++ CFLAGS=-Werror CXXFLAGS=-Werror"
+            configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java CC=clang CXX=clang++ CFLAGS=\"-Werror -Wno-error=unused-parameter\" CXXFLAGS=\"-Werror -Wno-error=unused-parameter\""
```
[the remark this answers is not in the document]
uckelman-sf: This is still not fixed.

Makefile.am
```diff
@@ -139,7 +140,7 @@ noinst_LTLIBRARIES = \
 	tsk/vs/libtskvs.la
 
 tsk_auto_libtskauto_la_CFLAGS = $(AM_CFLAGS) -Wmultichar
-tsk_auto_libtskauto_la_CXXFLAGS = $(AM_CXXFLAGS) -Wmultichar -Wsign-promo
+tsk_auto_libtskauto_la_CXXFLAGS = $(AM_CXXFLAGS) -Wmultichar -Wsign-promo -Wno-error=overloaded-virtual
```
uckelman-sf, line 143: `-Wno-error=overloaded-virtual` is no longer needed, since 2cbd0d5f.

Makefile.am
```diff
@@ -412,6 +415,7 @@ tools_autotools_tsk_gettimes_SOURCES = tools/autotools/tsk_gettimes.cpp
 tools_autotools_tsk_imageinfo_LDADD = $(TSK_LIB)
 tools_autotools_tsk_imageinfo_SOURCES = tools/autotools/tsk_imageinfo.cpp
 
+tools_autotools_tsk_loaddb_CXXFLAGS = $(AM_CXXFLAGS) -Wno-error=overloaded-virtual
```
uckelman-sf, line 418: `-Wno-error=overloaded-virtual` is no longer needed. This line should be removed.

tsk/fs/btrfs_csum.cpp
```diff
@@ -0,0 +1,40 @@
...
+**
+** This software is distributed under the Common Public License 1.0
+*/
+
+/*
+ * Contains the checksum part for Btrfs file system support.
+ */
+
+#include "tsk/base/crc.h"
+#include <stdio.h>
```
uckelman-sf, line 18: What here needs `<stdio.h>`?

tsk/fs/btrfs_csum.cpp
```diff
@@ -0,0 +1,40 @@
...
+ */
+extern "C" unsigned long
+btrfs_csum_crc32c(const unsigned char *a_data, const int a_len)
+{
+    cm_t cm;
+    cm.cm_width = 32;
+    cm.cm_poly = 0x1EDC6F41;
+    cm.cm_init = 0xFFFFFFFF;
+    cm.cm_refin = TRUE;
+    cm.cm_refot = TRUE;
```
uckelman-sf, line 34: These ought to be using a real `bool`, so `true`, not `TRUE`.

tsk/fs/btrfs_csum.cpp
simsong: I may have added it for debugging. I'll remove it.

tsk/fs/btrfs_csum.cpp
simsong: thanks.

tsk/fs/btrfs_csum.cpp
simsong: fixed in 29968391c

tsk/fs/btrfs_csum.cpp
simsong: fixed in 29968391c

Makefile.am
simsong: fixed in 7ae73033c

Makefile.am
simsong: removed in 68185891e.

Makefile.am
```diff
@@ -636,7 +640,7 @@ check_PROGRAMS = \
 # This makes things easier with codecov.
 test_runner_CPPFLAGS = $(AM_CPPFLAGS) -Ivendors $(CATCH2_CPPFLAGS)
 test_runner_CFLAGS = $(AM_CFLAGS) $(PTHREAD_CFLAGS)
-test_runner_CXXFLAGS = $(AM_CXXFLAGS) $(PTHREAD_CFLAGS)
+test_runner_CXXFLAGS = $(AM_CXXFLAGS) $(PTHREAD_CFLAGS) -Wno-error=overloaded-virtual
```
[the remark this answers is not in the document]
simsong: removed in 12978cab7

Makefile.am
```diff
@@ -2,7 +2,7 @@ ACLOCAL_AMFLAGS = -I m4
 
 AM_CPPFLAGS = -I$(srcdir)/tsk
 AM_CFLAGS = -Wall -Wextra
-AM_CXXFLAGS = -Wall -Wextra -Woverloaded-virtual
+AM_CXXFLAGS = -Wall -Wextra
```
uckelman-sf, line 5: `-Woverloaded-virtual` should not be removed here. We want warnings for this now that we've fixed all the old problems.

Makefile.am
```diff
@@ -412,6 +415,7 @@ tools_autotools_tsk_gettimes_SOURCES = tools/autotools/tsk_gettimes.cpp
 tools_autotools_tsk_imageinfo_LDADD = $(TSK_LIB)
 tools_autotools_tsk_imageinfo_SOURCES = tools/autotools/tsk_imageinfo.cpp
 
+tools_autotools_tsk_loaddb_CXXFLAGS = $(AM_CXXFLAGS)
```
uckelman-sf, line 418: This line is unnecessary, since `$AM_CXXFLAGS` is the default value.

Makefile.am
simsong: not present now?

Makefile.am
simsong: fixed in 2584e0e67.

simsong:
fixed in d98b278c9

Makefile.am
```diff
@@ -2,7 +2,7 @@ ACLOCAL_AMFLAGS = -I m4
 
 AM_CPPFLAGS = -I$(srcdir)/tsk
 AM_CFLAGS = -Wall -Wextra
-AM_CXXFLAGS = -Wall -Wextra -Woverloaded-virtual
+AM_CXXFLAGS = -Wall -Wextra  -Woverloaded-virtual
```
uckelman-sf, line 5: Extra space.

simsong:
got it.

.github/workflows/build-unix.yml
```diff
@@ -49,7 +49,7 @@ jobs:
             linkage: "shared"
             compiler: "clang"
             runner: "ubuntu-22.04"
-            configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java CC=clang CXX=clang++ CFLAGS=-Werror CXXFLAGS=-Werror"
+            configure_opts: "--with-libewf --with-libqcow --with-libvhdi --with-libvmdk --disable-java CC=clang CXX=clang++ CFLAGS=-Werror CXXFLAGS=-Werror
```
uckelman-sf, line 52: This is missing the trailing double quote.

simsong:
Thanks to @uckelman-sf for all of your help!

simsong:
> > Wait, this should not have been merged yet.
> > 
> > * I didn't review the substantive parts of the code---I was waiting for all the build problems to be fixed before doing that.
> > * How are we validating that any of this works? How was the test data produced?
> 
> Whoops. I validated it on the disk image btrfs_testimage_50MB.E01

How do you want to proceed?

simsong:
> > Whoops. I validated it on the disk image btrfs_testimage_50MB.E01
> 
> How? By hand? Against some expected results?

By hand. I created the test image and ran fiwalk on it and I got results that I expected.
Separately, I've been working with NIST to see if we can create a new forensic image generator that generates a disk image and creates a DFXML file at the same time, so that we can automate the process of creating novel complex images and then testing them.

uckelman-sf:
Ok. I'd like to have a look through `tsk/fs/btrfs.cpp` and `tsk/fs/tsk_btrfs.h` when I'm back on Tuesday.

uckelman-sf:
I had a very cursory look over btrfs.cpp. There are many things which could be brought up to a modern standard---e.g., most of the uses of bare `new`,  the formatting function using `va_list`---which I don't have time at present to work at.

There are also no unit tests. I recognize that this code was submitted a long time ago and the project wasn't doing unit tests at that point; consequentially, there are still a number of old PRs without them. That said, we should not accept _new_ code without unit tests.

simsong:
I'm hoping to create some tests for this before we bring it in. 
I'm concerned about the way we are doing some of the end-to-end tests: the XFS test that we just submitted has a 2GB disk image, currently being distributed as a ZIP file and unzipped. I'll want to change it into an E01. Still haven't moved to LFS yet.

uckelman-sf:
What do you mean by "bring it in"? We've already brought it in by merging it. Is that referring to a release?

I haven't been able to return to LFS yet, due to being clobbered by the TZ bug.

simsong:
I guess we did bring it in, but perhaps we should comment it out, or add a
conditional flag, or runtime flag to enable experimental filesystems?

I’d like that idea a lot. Having it in keeps it up-to-date with other
changes, but having a runtime flag prevent it from being used accidentally.
what do you think?

PR #3067, opened 2024-10-29
test sha1 as well for fiwalk for test coverage

simsong:


simsong:
Well, didn't seem to matter much.

PR #3069, opened 2024-10-30
Stop suppressing unused-parameter warnings

uckelman-sf:
This stops suppressing unused parameter warnings so we can see them again, but excludes them from being errors in CI.


uckelman-sf:
@simsong Ready for review.

simsong (approved):

PR #3070, opened 2024-10-30
Ensure functions in sha2.h aren't declared if we have libcrypto

uckelman-sf:
This fixes a name collision when compiling with libcrypto.


uckelman:
@simsong Ready for review.

simsong (approved):

simsong:
I saw this earlier and was wondering about it.

PR #3072, opened 2024-10-31
Jgru add csv output to mmls

simsong:


simsong:
Hm. I'm calling mmls() twice, which means two calls to getopt(). I am getting around that by resetting optind. That's not working with gcc on Linux...

image

simsong:
argh.
https://github.com/lattera/glibc/blob/master/posix/getopt.c

simsong:
@uckelman  - When you have a chance, can you look at this?
https://github.com/sleuthkit/sleuthkit/actions/runs/11655070750/job/32449331870#step:11:185
I'm trying to have the mmls command linked into the runner. The problem I'm having is that the second and third calls to `getopt()` are not working properly. I've had this problem before when trying to get `bulk_extractor` work with Catch, and I ended up moving from `getopt()` to an argument parser that is explicitly reentrant. 

I don't want to change all of the argument parsers that all of the TSK tools are using, so if I can't get this to work, my plan is to just run `mmls` as a subprocess. That will work, but it seems gross.

Google's Gemini recommends using Catch2's mock system to mock the call. It recommends code like this:
```c++
#include <catch2/catch_test_macros.hpp>
#include <catch2/matchers/string_matchers.hpp>

#include "your_program.hpp" // Assuming your getopt() logic is in this file

TEST_CASE("Test getopt() with mock", "[getopt]") {
    // Mock getopt() to return specific values
    auto mock_getopt = [](int argc, char** argv) -> int {
        // Implement your mock logic here
        // For example, return specific values based on the current iteration
        static int index = 0;
        switch (index++) {
            case 0:
                return 'a';
            case 1:
                return 'b';
            default:
                return -1;
        }
    };

    // Replace the real getopt() with the mock
    auto original_getopt = ::getopt;
    ::getopt = mock_getopt;

    // Call your program's main function with test arguments
    int argc = 3;
    char* argv[] = {"your_program", "-a", "-b"};
    int result = main(argc, argv);

    // Assert the expected behavior based on your mock implementation
    // ...

    // Restore the original getopt()
    ::getopt = original_getopt;
}
```

Thoughts?

test/legacy/runtests.sh
```diff
@@ -32,7 +32,7 @@ fi
 FS_THREAD_TEST="tests/fs_thread_test"
 
 if ! test -x ${FS_THREAD_TEST} ; then
-    FS_THREAD_TEST="tests/fs_thread_test.exe"
+    FS_THREAD_TEST="tests/fs_thread_test"
```
uckelman-sf, line 35: This change defeats the purpose of the case. When `tests/fs_thread_test` isn't executable, `tests/fs_thread_test.exe` is supposed to be tried, because when that happens we're probably running tests on Windows or via Wine.

test/runner.cpp
```diff
@@ -11,5 +11,100 @@
...
+        fseeko(file,0L,0);
+        return file_contents_is(filename, contents);
+    }
+
+    std::string tempfile::first_line() {
+        char buf[1024];
+        fseek(file,0L,0);
+        return fgets(buf,sizeof(buf),file);
+    }
+}
```
uckelman-sf, line 110: This should be in its own file.

uckelman-sf:
I do have a better suggestion, but I'm not that far along yet.

test/tsk/vstools/test_mmls.cpp
```diff
@@ -0,0 +1,114 @@
+#include <fstream>
```
uckelman-sf, line 1: Tests for `tools/vstools/mmls.cpp` should be in `test/tools/vstools/test_mmls.cpp`.

simsong: Noted. Thanks.

simsong:
> I do have a better suggestion, but I'm not that far along yet.

I don't want to start patching tools until I have a way of validating that I haven't changed their output. So even though this PR is clearly not a problem, I'm using it as an opportunity to figure out how to test the command-line tools.

The obvious way is with a subprocess, but that causes visibility problems with the catch testing framework.

test/legacy/runtests.sh
simsong: Oh, okay.

tools/vstools/mmls_main.cpp
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * mmls - list media management structure contents.
+ *      - This is main() that gets linked for the stand-alone program
+ *
+ */
+#include "tsk/tsk_tools_i.h"
+#include "mmls.h"
+
+int
+main(int argc, char **argv1)
```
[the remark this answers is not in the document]
simsong: I know, but this is the way that they are named in the TSK code base. Do you think I should just go ahead and change it?

uckelman-sf:
What specifically are you trying to test?

* If you want to test the output of the `mmls` command, that's best done with a shell script and diffing the output against expected output.
* If you want to test what `mmls` does after argument parsing, that would be best done by refactoring to have the argument parser fill a struct which is then passed on to a function that does all the work (roughly from the original line 266 to the end of mmls.cpp), and then write tests which pass that struct filled in various ways to the work-doing function.
* If you want to test the argument parsing, that would be best done by refactoring to have the argument parser in its own function and fill a struct with the results, which you can then write tests against.

The thing I would not do is try to mock out the argument parser. That will be much messier than decoupling the code which does the work from the code which parses the arguments.

simsong:
> What specifically are you trying to test?
> 
> * If you want to test the output of the `mmls` command, that's best done with a shell script and diffing the output against expected output.
> * If you want to test what `mmls` does after argument parsing, that would be best done by refactoring to have the argument parser fill a struct which is then passed on to a function that does all the work (roughly from the original line 266 to the end of mmls.cpp), and then write tests which pass that struct filled in various ways to the work-doing function.
> * If you want to test the argument parsing, that would be best done by refactoring to have the argument parser in its own function and fill a struct with the results, which you can then write tests against.
> 
> The thing I would not do is try to mock out the argument parser. That will be much messier than decoupling the code which does the work from the code which parses the arguments.

I'm trying to test the functionality of the mmls command, and I'm also simultaneously trying to show that the functionality is covered by the code coverage tool.

Regarding the functionality, I want to show the stability of the output for a specific set of flags and inputs.

I can certainly run mmls as a subprocess. My understanding is that these runs with the codecov binary will produce output in the file system that may be included in the code coverage reports. 

The problem with the argument parser is that it's legacy `getopt()` and it doesn't work properly when it is called more than once. I've looked at the source code. It should work if I change `optind` but it doesn't, and I don't know why, but the code is a tangled mess. That's why I stopped using it in bulk_extractor when I was trying to get code coverage of the options processing.

I guess your recommendation is to have tests that simply test each output and not run those tests under catch?

test/runner.cpp
simsong: as support for the runner?

test/runner.cpp
```diff
@@ -11,5 +11,100 @@
...
+        for (std::size_t i = 0; i < length; ++i) { random_string += CHARACTERS[distribution(generator)]; }
+
+        return random_string;
+    }
+
+    bool contains(std::string line, std::string substr) {
+        return line.find(substr) != std::string::npos;
+    }
+
+    std::string get_tempdir(std::string testname) {
```
uckelman-sf, line 44: This should be done with `std::filesystem::temp_directory_path()`.

test/runner.cpp
uckelman-sf: Yes, in a utils file.

simsong:
It was, but that was not available in C++11. I think it needs to go to 17
for this.

uckelman-sf:
> The problem with the argument parser is that it's legacy getopt() and it doesn't work properly when it is called more than once. I've looked at the source code. It should work if I change optind but it doesn't, and I don't know why, but the code is a tangled mess. That's why I stopped using it in bulk_extractor when I was trying to get code coverage of the options processing.

Give me a chance to show you what I'm talking about w/r/t separating out the code tomorrow.

simsong:
Okay. I’ll push what I have and you can use that if you wish.
Simson

uckelman-sf:
Didn't get to the end of this today. It's on my list for Thursday now.

simsong:
No rush.

simsong:
@uckelman-sf  — 
The mocking approach that I am using it working, so I'm inclined to go ahead with it.
Let me try to do a few more tests and see how it works.
https://app.codecov.io/github/sleuthkit/sleuthkit/pull/3072/blob/tools/vstools/mmls.cpp

uckelman-sf:
What I suggest be done with `mmls.cpp` prior to the changes in this MR is here: https://github.com/sleuthkit/sleuthkit/commit/3d939337e36a049a804416b0f1b1789d23756125

This separates the arg parsing and processing into their own functions, cleans up memory allocation and error handling, and eliminates all the global variables.

simsong:
I appreciate this, but I do not want to do this until we have the existing code base tested with the existing option parser approaches.

uckelman-sf:
Ok, then I will go back to add tests _before_ the changes.

simsong:
> Ok, then I will go back to add tests _before_ the changes.

I'm trying! The big problem I'm having now is the whole utf-8 / utf-16 issue. But I'm nearly there. I hope to have this completed today. And yes, I will pull things out of runner.h and runner.cpp and into their own support file.

uckelman-sf:
I think maybe you've misunderstood. I really do mean to add the tests before the changes.

uckelman-sf:
PR #3079 adds output tests for mmls, as promised above. The one thing missing is toggling tests for image types depending on whether they're supported, but that doesn't prevent CI from working as-is.

uckelman-sf:
PR #3079 now runs tests only for supported image formats.

simsong:
Wow. I got it to work (finally), and I see that codecov is editing the coverage report.
image

@uckelman-sf  - I'm inclined to leave the support code in runner.h and runner.cpp; can you explain to me why it should be moved to another program?

uckelman-sf:
> I'm inclined to leave the support code in runner.h and runner.cpp; can you explain to me why it should be moved to another program?

I'm not suggesting that the support code be moved to another program, but rather another file. Two reasons:

* Putting test support code into runner.cpp mixes code used by the tests with code that does nothing other than set up the runner. The tests don't need to see the runner setup code; the runner setup code doesn't need to see the test support code. When they're mixed together, there is no longer anything demonstrating that clearly.
* runner.cpp ends up with a `main()` in it supplied by Catch2. If the support code is needed by a program other than `runner`, the support code won't be usable because `runner.o` will contain a `main()`, so the linker will fail with a multiply defined symbol error.

simsong:
Okay, thanks for the clarification. Those are great reasons. I’ll make the changes and will use the tests mostly for testing argument parsing, and not for output correctness, given the test framework that you just checked in.

uckelman-sf:
PR #3090 minimally adds the csv functionality to mmls without conflicts with the develop branch.

simsong:
Right. I just don’t think that is the correct approach at this point. Per
the other messages, I think we want to have a class for doing the
formatting and outputting, and have the class take an option to simply
output, a CSV instead of with formatting.

uckelman-sf:
PR #3090 is a better place to work from to add formatter interface than here.

simsong:
> I think maybe you've misunderstood. I really do mean to add the tests before the changes.

I didn't. There is just a lot happening, and all at the same time.
Any clue why mmls under mingw is not outputing correct data? 
Separately, I'm confused by the UTF-16 output when the tools are compiled with MSC. I guess that Win32 command-line programs can run in UTF-16 mode? I've just never seen this.

simsong:
Closing this as it was completed in #3090.

PR #3075, opened 2024-11-04
Codecov adjustments

uckelman-sf:
Exclude `test/`, `vendors/` from coverage analysis.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,7 +10,7 @@ codecov:
   branch: develop
 
 ignore:
-  - "m4/*"
+  - "m4"
 
 # Code for updating pull request comment on GitHub
 # See https://docs.codecov.com/docs/pull-request-comments
--- a/codecov.yml
+++ b/codecov.yml
@@ -11,6 +11,8 @@ codecov:
 
 ignore:
   - "m4"
+  - "test"
+  - "vendors"
 
 # Code for updating pull request comment on GitHub
 # See https://docs.codecov.com/docs/pull-request-comments


uckelman-sf:
@simsong How do I check if this is working?

simsong:
> @simsong How do I check if this is working?

It's not, but now I know why:
Google Chrome 2024-11-07 07 27 49

Google Chrome 2024-11-07 07 28 12

It appears that you do not have access to the `CODECOV_TOKEN` because it is a repository secret and you don't get access to those secrets when you run.   It's once supposed to need the secrets for an upload from an open source repository, but this seems to have changed...

https://github.com/codecov/codecov-action/issues/1255
"Yes, the release notes contradict GitHubs documentation."

I'll look into this later today.  As a short-term fix, you could try adding codecov to your personal repo.

https://docs.codecov.com/docs/adding-the-codecov-token#:~:text=The%20Codecov%20token%20must%20be,repository%20in%20Codecov%2C%20under%20General.

https://docs.codecov.com/docs/frequently-asked-questions

uckelman-sf:
Maybe someone could cherry-pick the two commits here into a branch were codecov does have access to the secrets.

simsong:
> Maybe someone could cherry-pick the two commits here into a branch were codecov does have access to the secrets.

What do you want me to do?

uckelman-sf:
Nothing, now that this is merged and we can see the result in codecov. `tsk` and `tools` only are showing up in the results, which is the desired result.

PR #3077, opened 2024-11-05
Replaced malloc, free, goto in tsk/fs/hfs.cpp

uckelman-sf:


tsk/fs/hfs.cpp
```diff
@@ -746,16 +730,17 @@ hfs_cat_traverse(HFS_INFO * hfs,
...
     uint8_t is_done = 0;
 
     tsk_error_reset();
 
     nodesize = tsk_getu16(fs->endian, hfs->catalog_header.nodesize);
-    if ((node = (char *) tsk_malloc(nodesize)) == NULL)
+    std::unique_ptr<char[]> node{new char[nodesize]};
+    if (!node) {
         return 1;
+    }
```
simsong, line 743: can this even fail? I thought that a failed `new` would throw an exception.

tsk/fs/hfs.cpp
```diff
@@ -367,7 +366,8 @@ hfs_ext_find_extent_record_attr(HFS_INFO * hfs, uint32_t cnid,
 
     // allocate a node buffer
     nodesize = tsk_getu16(fs->endian, hfs->extents_header.nodesize);
-    if ((node = (char *) tsk_malloc(nodesize)) == NULL) {
+    std::unique_ptr<char[]> node{new char[nodesize]};
+    if (!node) {
```
simsong, line 370: can this even fail? I thought that a failed `new` would throw an exception?

simsong: Copilot
2 minutes ago
In modern C++ implementations, new does indeed throw a std::bad_alloc exception if it fails to allocate memory. Therefore, checking the result of new against NULL or nullptr is unnecessary. The code you are referring to, which checks if the allocation was successful by examining the result of new, is outdated and does not reflect the current behavior of the new operator in C++:

std::unique_ptr<char[]> node{new char[nodesize]};
if (!node) {
    return 1;
}
This check is redundant because new will throw an exception if the allocation fails.

simsong (changes requested):
ChatGPT says we don't need to check to see if `node==NULL`. What do you think?

tsk/fs/hfs.cpp
uckelman-sf: It will. These should all be the nothrow version of `new`.

tsk/fs/hfs.cpp
uckelman-sf: The check isn't redundant. The call to `new` is to the wrong overload of the operator. We're not expecting exceptions; for the error handling code to continue working without modification, we need the nothrow version of `new` here. I've pushed a change to use that. Thanks for spotting the problem. 

It's been a while since I've worked behind a C API that doesn't have exception traps at the boundaries---which we should absolutely have---so I'd forgotten about the need to use nothrow new in cases like this.

uckelman-sf:
See comments above. The check is correct, the version of `new` was not.

simsong (approved):
That makes sense. We are not redoing all of the error handling to be exception-based.

PR #3081, opened 2024-11-07
mmls cleanup

uckelman-sf:


simsong:
> @simsong This is ready for review after #3079.

On this and the other others, I am waiting to see what AppVeyor does..

simsong (approved):
Well, this really is the complete re-implementation of all option processing within mmls.

uckelman-sf:
It's not. The changes to option parsing are limited to storing the parsed values in a struct. That's all. The calls to the parsing functions haven't changed.

simsong:
> It's not. The changes to option parsing are limited to storing the parsed values in a struct. That's all. The calls to the parsing functions haven't changed.
> 
> Look at what has changed on the lines from 160 to 206, approximately. It's almost entirely switching assignments to `x` to `opts.x`.

Yes, that's what you did. "complete re-implementation" is an exaggeration. It would be great to abandon getopt.

uckelman-sf:
Why do you want to abandon getopt?

simsong:
Because it doesn’t work if you call it twice, so it’s hard to test code that uses getopt.

uckelman-sf:
There's no reason it shouldn't work. Can you show me an example where it doesn't?

simsong:
Yes. The multiple errors that I encountered in trying to get mmls to work, which finally resulted in my mocking out `getopt()`, are good examples.
After the initial call, it would not return valid results, even when I reset optind.

Meanwhile, because of the extensive changes in the option handling of mmls, I’m now no longer able to get https://github.com/sleuthkit/sleuthkit/pull/3072 to work, which makes me very sad. It requires substantial modifications now to make it work with the new option handling. This makes me very hesitant to accept more rewrites of the option processing of the other command line tools.

And now we have conflicts with https://github.com/sleuthkit/sleuthkit/pull/3065 as well.

You’ve done a huge amount of work on this project, for which we are grateful, but there are now a lot of conflicts to resolve.

simsong:
Specifically, look at this run:
You'll see that it got an invalid option `'p'` even thought that was never provided.

uckelman-sf:
I'll resolve the conflicts, if you want, but not over the weekend. I was trying to avoid this situation by suggesting that the mmls code be cleaned up first, but I wasn't successful at that.

simsong:
I will do it. It’s a good learning experience for me. You’re a great
teacher!

PR #3083, opened 2024-11-08
Move image downloader to its own directory

uckelman-sf:
Move image downloader to its own directory; it's no longer used for fiwalk only.


uckelman-sf:
@simsong Ready for review.

simsong (approved):
Thanks. I'm still thinking about moving the images to a sub module, but just haven't gotten to it yet.

PR #3085, opened 2024-11-08
Move bundled sqlite3 to vendors

uckelman-sf:
This PR moves the bundled sqlite3 to `vendors/`, where it should be because it's not our code.


uckelman-sf:
@simsong Ready for review.

simsong (approved):
`/vendors/` is fine, but I probably would have gone with `/3rdparty/`...

uckelman:
`vendors/` is used fairly widely. I've never seen `3rdparty/` in a project before.

simsong:
I defer to you!

PR #3091, opened 2024-11-11
Replace gotos in tsk/img/raw.cpp

uckelman-sf:


uckelman-sf:
The most recent run is with the tests only---no modifications to `raw_open()`. They seem to show that `raw_open()` has one or more memory leaks as it exists now.

uckelman-sf:
@simsong Ready for review.

PR #3096, opened 2024-11-13
Add sleuthkit_test_data as a submodule

uckelman-sf:


simsong:
We ran into bandwidth limitations with sleuthkit_test_data as a submodule, so we are giving up on this approach.

simsong:
> We ran into bandwidth limitations with sleuthkit_test_data as module that is cloned due to LFS issues, so we gave up on this approach.

PR #3098, opened 2024-11-13
CT-9000 Loading Files is taking a very long time

eugene7646:


APriestman:
I think this should be targeted at skl_main

simsong:
Perhaps, but we are happy to accept it over here in the world of the
development branch.

simsong:
Great. And we are eager for you to take develop into skl_main once the massive upgrade refactoring progression has stabilized.

PR #3100, opened 2024-11-13
Upstream merge from SKL main

jayaramcs:


simsong:
Let's see if updating the branch fixes the AppVeyor problem.

simsong:
> Let's see if updating the branch fixes the AppVeyor problem.

It did!

bindings/java/src/org/sleuthkit/datamodel/CaseDatabaseFactory.java
```diff
@@ -786,8 +786,12 @@ void performPreInitialization(Connection conn) throws TskCoreException {
...
 			try (Statement stmt = conn.createStatement()) {
 				stmt.execute("ALTER SEQUENCE blackboard_artifacts_artifact_id_seq minvalue -9223372036854775808 restart with -9223372036854775808");
+				
+				// CT-9000: Postgres supports composite and partial indexes which results in smaller indexes and faster inserts. 
+				// So in Postgres we can have an index which indexes only tsk_files with non-null MD5 and non-zero size:
+				stmt.execute("CREATE INDEX tsk_files_datasrc_md5_size_partial_index ON tsk_files(data_source_obj_id, md5, size) WHERE md5 IS NOT NULL AND size > 0");
 			} catch (SQLException ex) {
-				throw new TskCoreException("Error altering artifact ID sequence", ex);
+				throw new TskCoreException("Error performing PostgreSQL post table initialization", ex);
 			}
```
simsong, line 795: I'm completely unclear on why you would add `, size` to the index after `md5`. Isn't every file with the same `md5` going to be the same size? @jayaramcs ?

jayaramcs: Yes, but we get a whole lot of file entries where MD5s are not calculated and is therefore NULL. It is also possible to have entries in this table with MD5s but no size - such as when we ingest flat file with paths and MD5s and not the sizes.

simsong: Ah. Got it. You might want to add a comment to that effect. Thanks.

jayaramcs:
Also, @simsong Please don't delete skl_main branch. We have not tested develop against ongoing releases of Cyber Triage. All our work is from skl_main which we have tested for stability. At some point in the future we will pull in develop to skl_main after we have tested develop for stability and edge cases, but we don't expect to be directly pulling develop as it can make Cyber Triage unstable. 
We will issue a PR to develop should we make any changes to skl_main after it has been tested and reviewed internally first.

simsong:
> Also, @simsong Please don't delete skl_main branch. We have not tested develop against ongoing releases of Cyber Triage. All our work is from skl_main which we have tested for stability. At some point in the future we will pull in develop to skl_main after we have tested develop for stability and edge cases, but we don't expect to be directly pulling develop as it can make Cyber Triage unstable. We will issue a PR to develop should we make any changes to skl_main after it has been tested and reviewed internally first.

Hi. I won't delete the branch (although it's easy to get back deleted branches). Even better, I will protect the branch.

simsong:
@jayaramcs —
image
Feel free to increase the level of protection.

simsong:
(I'm unclear why it is still offering to let me delete the branch.)

simsong:
image

PR #3101, opened 2024-11-14
Added img_open tests, replaced gotos in tsk/img/img_open.cpp

uckelman-sf:


uckelman-sf:
@simsong Ready for review. To be merged after #3097.

uckelman-sf:
AppVeyor's Mac builds are failing to start, it seems.

simsong:
> AppVeyor's Mac builds are failing to start, it seems.

I think that SleuthKitLab's plan with AppVeyor has a limited number of runs and that we are overwhelming it with frequent commits. I think that we should set it up so that it only runs if all of the other checks work. Unfortunately, I have no ability to modify AppVeyor. The integration with GitHub seems to be happening using webhooks. I do not have access to the AppVeyor console, only @bcarrier and @jayaramcs do.

On the other hand, AppVeyor finally ran and it failed, but for no good reason:

Google Chrome 2024-11-16 11 22 33

Google Chrome 2024-11-16 11 23 23

PR #3105, opened 2024-11-19
Add tests for tsk_img_read(), fix tsk_img_read_no_cache bug

uckelman-sf:
* Adds tests for `tsk_img_read()`
* Ensure that `tsk_img_read_no_cache()` returns `-1` when the underlying read function does. The existing code swallows errors, which could result in returning corrupt data.


uckelman-sf:
@simsong Ready for review.

This fixes a serious image reading bug.

simsong:
> @simsong Ready for review.
> 
> This fixes a serious image reading bug.

Thanks. Would it be useful to create a mock test so that we can test the error logic in general? That way we would catch these things.

Flagging this for @jayaramcs and @bcarrier

uckelman-sf:
> Would it be useful to create a mock test so that we can test the error logic in general? That way we would catch these things.

I'm already doing that. `tsk_img_read inner function failed` in `test/tsk/img/test_img_io.cpp` in this PR is a mock test for catching problems like this. I have several more mock tests for `test/tsk/img/test_img_io.cpp` which will be in another PR soon.

I think we should have a release soon, so that we're not sitting on bug fixes like this one. (There is one PR I would like to have merged before that happens, however, which I expect to make today.)

simsong:
Agreed.

PR #3111, opened 2024-11-26
Fix signatures of TskAuto::getErrorList() and TskAuto::errorRecordToString():

uckelman-sf:
This PR fixes two API problems with `TskAuto`:

* Fixed signature of `TskAuto::getErrorList()`: Returning a `const std::vector` makes no sense; the intent is clearly to return a const reference to the error list, not an unmodifiable copy.

* Fixed signature of `TskAuto::errorRecordToString()`: You can't pass a const record to it, and all the records from `getErrorList()` are/were const. `errorRecordToString()` doesn't modify the error record and ought to work with records from `getErrorList()`, so its argument should be a const reference.

### DIFF
--- a/tsk/auto/auto.cpp
+++ b/tsk/auto/auto.cpp
@@ -974,7 +974,7 @@ uint8_t TskAuto::registerError() {
 }
 
 
-const std::vector<TskAuto::error_record> TskAuto::getErrorList() {
+const std::vector<TskAuto::error_record>& TskAuto::getErrorList() {
     return m_errors;
 }
 
@@ -982,7 +982,7 @@ void TskAuto::resetErrorList() {
     m_errors.clear();
 }
 
-std::string TskAuto::errorRecordToString(error_record &rec) {
+std::string TskAuto::errorRecordToString(const error_record& rec) {
     tsk_error_reset();
     tsk_error_set_errno(rec.code);
     tsk_error_set_errstr("%s", rec.msg1.c_str());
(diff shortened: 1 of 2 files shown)


uckelman-sf:
@simsong Ready for review.

simsong:
Thank you for this. However, if we want to get out the next release, we should probably focus on the remaining issues and put a stake in the ground.

uckelman-sf:
What remaining issues do you think need to be resolved for the next release?

simsong:
> What remaining issues do you think need to be resolved for the next release?

I want btrfs in the release, and I don't know why we are having problems with the fiwalk test under mingw.

simsong:
We  need to get this: https://github.com/sleuthkit/sleuthkit/pull/3065

uckelman-sf:
Why? Couldn't it be in the release after next?

uckelman-sf:
Is that a reason why btrfs support needs to be in the next release instead of the release after? I think it's not. 

Putting btrfs support into 4.14.0 instead of 4.13.0 and doing a release now won't delay btrfs support, but holding up 4.13.0 for btrfs support will delay everything else in 4.13.0. In terms of wall clock time, you'll get btrfs support at about the same time either way.

simsong:
I can’t do a release right now. The release is dependent upon coordinating with Sleuth Kit Labs, which is in the process of testing develop with autopsy and cybertriage. Currently there is an integration problem and the problem has not been determined.

PR #3116, opened 2024-11-29
Add -k password field to tsk_recover.

APriestman:
Change vcxproj to use HAVE_LIBCRYPTO.


simsong:
Thanks! I will review in detail in the morning. I will see if I can add a unit test as well...

PR #3120, opened 2024-12-03
Add options struct to image open API

uckelman-sf:
We need to be able to pass options to the image open functions for things like setting a default time zone and image block cache parameters.

### DIFF
--- a/tsk/img/img_open.cpp
+++ b/tsk/img/img_open.cpp
@@ -51,6 +51,9 @@
 #include <vector>
 #include <utility>
 
+const TSK_IMG_OPTIONS DEFAULT_IMG_OPTIONS{
+};
+
 bool sector_size_ok(unsigned int sector_size) {
     if (sector_size > 0 && sector_size < 512) {
         tsk_error_set_errno(TSK_ERR_IMG_ARG);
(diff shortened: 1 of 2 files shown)


uckelman-sf:
> Do you want me to make the initial file?

Sure.

uckelman-sf:
@simsong Ready for review.

tsk/img/img_open.cpp
```diff
@@ -270,10 +273,11 @@ img_open_detect_type(
...
 TSK_IMG_INFO* img_open(
-    int num_img,
-    const TSK_TCHAR* const images[],
-    TSK_IMG_TYPE_ENUM type,
-    unsigned int a_ssize
+  int num_img,
+  const TSK_TCHAR* const images[],
+  TSK_IMG_TYPE_ENUM type,
+  unsigned int a_ssize,
+  [[maybe_unused]] const TSK_IMG_OPTIONS* opts
```
simsong, line 280: I think that we should not be changing indents. It looks like some of the code is indented at 4 and others indented at 2. In the future, can we keep things consistent?

simsong (approved):

tsk/img/img_open.cpp
uckelman: Yes, it would be great if everything used 2 spaces, and no tabs.

simsong: I agree with you on no tabs, but I prefer 4 spaces. However, I do not think that we currently have a coding standards document.

PR #3125, opened 2024-12-04
string format specifier fix

gdicristofaro:
I believe this commit caused issues in Autopsy with importing hash sets: https://github.com/sleuthkit/sleuthkit/commit/a6d4d2917e9dbd8a0fbd6a9a14f4bffad967c1d7.  I believe this code should fix that issue without compiler warnings.

### DIFF
--- a/tsk/base/tsk_os.h
+++ b/tsk/base/tsk_os.h
@@ -212,7 +212,7 @@ typedef char TSK_TCHAR;         ///< Character data type that is UTF-16 (wchar_t
 #define TZSET	tzset
 #define TZNAME	tzname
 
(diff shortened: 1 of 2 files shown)


bcarrier:
For completeness of this story, Windows considers %hs a defined "format":

https://devblogs.microsoft.com/oldnewthing/20190830-00/?p=102823

uckelman-sf:
It would be good to add some test cases for `hdb_binsrch_idx_init_hash_type_info` to establish what the correct outputs are.

simsong:
Thank you @gdicristofaro !

Issue #3112, opened 2024-11-26
FAT file system times were reported in local time previously and are now being reported in GMT.


bcarrier:
I ran a regression test on some non-public images and compared with the previous release.  The time stamps as reported by 'fls' and 'ils' are off by 4 hours. Because 'ils' should be reporting UTC that implies that it's the conversion from the date stored in the FAT file system and its conversion to UTC.  

I haven't been tracking any of the changes that went into this release, so I'm not sure what changed around FAT and time parsing.

simsong:
Thanks. Do you have known good output on a public image that we can work with?

uckelman:
I'll run down the problem if I can get a test image demonstrating it.

bcarrier:
This FAT image does: http://prdownloads.sourceforge.net/dftt/2-kwsrch-fat.zip?download
(I hadn't realized how many ads there are Source Forge these days..)

I ran 'fls -rlp fat-img-kw.dd > ~/fls-VERSION.txt' for each version and then did a diff. That's what my regression test infrastructure does. 

```
$ diff fls-4.12.1.txt fls-develop.txt 
1,8c1,8
< r/r 4:	file1.dat	2003-08-21 01:20:38 (EDT)	2003-08-21 00:00:00 (EDT)	0000-00-00 00:00:00 (UTC)	2003-08-21 01:20:38 (EDT)	512	0	0
< r/r 6:	file2.dat	2003-08-21 01:34:22 (EDT)	2003-08-21 00:00:00 (EDT)	0000-00-00 00:00:00 (UTC)	2003-08-21 01:34:22 (EDT)	400	0	0
< r/r 8:	file3.dat	2003-08-21 01:20:46 (EDT)	2003-08-21 00:00:00 (EDT)	0000-00-00 00:00:00 (UTC)	2003-08-21 01:20:46 (EDT)	900	0	0
< r/r 10:	file4.dat	2003-08-21 01:24:16 (EDT)	2003-08-21 00:00:00 (EDT)	0000-00-00 00:00:00 (UTC)	2003-08-21 01:24:16 (EDT)	631	0	0
< r/r * 12:	_ILE5.DAT	2003-08-21 01:21:36 (EDT)	2003-08-21 00:00:00 (EDT)	0000-00-00 00:00:00 (UTC)	2003-08-21 01:21:36 (EDT)	512	0	0
[...]

PR #3126, opened 2024-12-05
Upgrade CI runners to Ubuntu 24.04

uckelman-sf:
This PR upgrades the CI runners to Ubuntu 24.04. This gives us access to compilers which support C++20.

### DIFF
--- a/.github/workflows/build-unix.yml
+++ b/.github/workflows/build-unix.yml
@@ -62,7 +62,7 @@ jobs:
             runner: "ubuntu-24.04"
             configure_opts: "--host=x86_64-w64-mingw32 --target=x86_64-w64-mingw32 --disable-java --enable-shared --disable-static"
             winearch: 'win64'
-            winepath: 'Z:\usr\lib\gcc\x86_64-w64-mingw32\10-posix'
+            winepath: 'Z:\usr\lib\gcc\x86_64-w64-mingw32\13-posix'
             wineprefix: '/home/runner/.wine64'
             codecov: "no"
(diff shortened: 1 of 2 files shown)


uckelman-sf:
@simsong Ready for review. This is required for fixing #3112.

simsong (approved):

uckelman:
> Why are we not using ubuntu-latest?

What `ubuntu-latest` is can change without our doing anything, and that could make CI begin failing, again, without our doing anything.

The code change is in response to exactly such a failure. One of the newer compilers emits a warning we hadn't had before, which causes the CI run to fail. (The code was not standards compliant to start with---stack allocated arrays need a known size at compile time---but getting a warning for the problem is new.) A failure of this sort could happen spontaneously if we were using `ubuntu-latest` and we wouldn't know why.

simsong:
> > Why are we not using ubuntu-latest?
> 
> What `ubuntu-latest` is can change without our doing anything, and that could make CI begin failing, again, without our doing anything.
> 
> The code change is in response to exactly such a failure. One of the newer compilers emits a warning we hadn't had before, which causes the CI run to fail. (The code was not standards compliant to start with---stack allocated arrays need a known size at compile time---but getting a warning for the problem is new.) A failure of this sort could happen spontaneously if we were using `ubuntu-latest` and we wouldn't know why.

Okay. That's what I suspected you would argue.

On my other projects, I find that the creep of compilers in `-latest` keeps me on my toes, but this project has better staffing.

Issue #3127, opened 2024-12-06
Add some test cases for `hdb_binsrch_idx_init_hash_type_info` to establish what the correct outputs are.


simsong:
Add some test cases for `hdb_binsrch_idx_init_hash_type_info` to establish what the correct outputs are.

c.f. https://github.com/sleuthkit/sleuthkit/pull/3125#issuecomment-2520679491

simsong:
@gdicristofaro - I assigned you to this because this is currently at the top of your mental stack, but if you don't want to do it, just unassign yourself.

Thanks!

bcarrier:
Top if Greg's mental stack is some release work, so it will be a bit until we get to this. I'll unassign in case someone else frees up sooner.

simsong:
> Top if Greg's mental stack is some release work, so it will be a bit until we get to this. I'll unassign in case someone else frees up sooner.

Of course!

PR #3130, opened 2024-12-12
test_binsrch_index

gdicristofaro:
Hi @simsong , here are some unit tests for https://github.com/sleuthkit/sleuthkit/issues/3127

### DIFF
--- a/test/tsk/hashdb/test_binsrch_index.cpp
+++ b/test/tsk/hashdb/test_binsrch_index.cpp
@@ -1,0 +1,94 @@
+#include "tsk/base/tsk_os.h"
+#include "tsk/hashdb/tsk_hashdb_i.h"
+#include "catch.hpp"
+
+void test_hdb_binsrch_idx_init_hash_type_info(
+    const TSK_TCHAR *db_name,
+    TSK_HDB_HTYPE_ENUM htype,
+    int expected_return,
+    int expected_hash_len,
+    const TSK_TCHAR *expected_idx_fname,
+    const TSK_TCHAR *expected_idx_idx_fname)
+{
+    TSK_HDB_BINSRCH_INFO hdb_binsrch_info;
+
+    /**
+     * setup values:
+     * normally, this would be done as a part of opening the hashdb, but not 
+     * for this unit test
+     **/
+    hdb_binsrch_info.hash_type = TSK_HDB_HTYPE_INVALID_ID;
+    // this value is read only in hdb_binsrch_idx_init_hash_type_info
+    hdb_binsrch_info.base.db_fname = (TSK_TCHAR *)db_name;
+    int ret_val = hdb_binsrch_idx_init_hash_type_info(&hdb_binsrch_info, htype);
+
+    REQUIRE(ret_val == expected_return);
+
+    if (expected_return == 0)
+    {
+        // if successful call, do comparisons
+        REQUIRE(hdb_binsrch_info.hash_len == expected_hash_len);
+        REQUIRE(TSTRCMP(hdb_binsrch_info.idx_fname, expected_idx_fname) == 0);
+        REQUIRE(TSTRCMP(hdb_binsrch_info.idx_idx_fname, expected_idx_idx_fname) == 0);
+    }
+
+    /**
+     * cleanup allocated items from hdb_binsrch_idx_init_hash_type_info.
+     * These are values set by hdb_binsrch_idx_init_hash_type_info during run 
+     * that need to be cleaned up. Normally, they would be cleaned up when the
+     * hash database is closed, but not for this unit test.
+     **/
+    if (hdb_binsrch_info.idx_fname != NULL)
+    {
+        free(hdb_binsrch_info.idx_fname);
+    }
+
+    if (hdb_binsrch_info.idx_idx_fname != NULL)
+    {
+        free(hdb_binsrch_info.idx_idx_fname);
+    }
+}
+
+TEST_CASE("test hdb_binsrch_idx_init_hash_type_info with md5 db type")
+{
+    test_hdb_binsrch_idx_init_hash_type_info(
+        _TSK_T("C:\\path\\to\\file.txt"),
+        TSK_HDB_HTYPE_MD5_ID,
+        0,
+        TSK_HDB_HTYPE_MD5_LEN,
+        _TSK_T("C:\\path\\to\\file.txt-md5.idx"),
+        _TSK_T("C:\\path\\to\\file.txt-md5.idx2"));
+}
+
+TEST_CASE("test hdb_binsrch_idx_init_hash_type_info with sha1 db type")
+{
+    test_hdb_binsrch_idx_init_hash_type_info(
+        _TSK_T("C:\\path\\to\\file.txt"),
+        TSK_HDB_HTYPE_SHA1_ID,
+        0,
+        TSK_HDB_HTYPE_SHA1_LEN,
+        _TSK_T("C:\\path\\to\\file.txt-sha1.idx"),
+        _TSK_T("C:\\path\\to\\file.txt-sha1.idx2"));
+}
+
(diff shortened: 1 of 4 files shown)


test/tsk/hashdb/test_binsrch_index.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * cleanup allocated items from hdb_binsrch_idx_init_hash_type_info.
+     * These are values set by hdb_binsrch_idx_init_hash_type_info during run 
+     * that need to be cleaned up. Normally, they would be cleaned up when the
+     * hash database is closed, but not for this unit test.
+     **/
+    if (hdb_binsrch_info.idx_fname != NULL)
+    {
+        free(hdb_binsrch_info.idx_fname);
```
uckelman-sf, line 43: Is there a memory leak if one of the `REQUIRE`s above fails? This code won't be reached.

test/tsk/hashdb/test_binsrch_index.cpp
```diff
@@ -0,0 +1,94 @@
...
+        REQUIRE(TSTRCMP(hdb_binsrch_info.idx_idx_fname, expected_idx_idx_fname) == 0);
+    }
+
+    /**
+     * cleanup allocated items from hdb_binsrch_idx_init_hash_type_info.
+     * These are values set by hdb_binsrch_idx_init_hash_type_info during run 
+     * that need to be cleaned up. Normally, they would be cleaned up when the
+     * hash database is closed, but not for this unit test.
+     **/
+    if (hdb_binsrch_info.idx_fname != NULL)
```
uckelman-sf, line 41: `free` is callable on `nullptr` and checks for nulls itself. There's no reason to check yourself whether the variable you're passing to `free` is nonnull first.

uckelman-sf (changes requested):

test/tsk/hashdb/test_binsrch_index.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // this value is read only in hdb_binsrch_idx_init_hash_type_info
+    hdb_binsrch_info.base.db_fname = (TSK_TCHAR *)db_name;
+    int ret_val = hdb_binsrch_idx_init_hash_type_info(&hdb_binsrch_info, htype);
+
+    REQUIRE(ret_val == expected_return);
+
+    if (expected_return == 0)
+    {
+        // if successful call, do comparisons
+        REQUIRE(hdb_binsrch_info.hash_len == expected_hash_len);
```
uckelman-sf, line 30: Can the `REQUIURE`s after this one be checked sensibly if this one fails? If so, this should be a `CHECK` instead of a `REQUIRE`, so that it doesn't prevent seeing the results of later tests. (Same goes for the two `REQUIRE`s after this. `REQUIRE` should be used only when execution must not proceed on failure.)

test/tsk/hashdb/test_binsrch_index.cpp
simsong: This is so true, but many people think that calling `free(NULL)` will result in a null pointer error.

test/tsk/hashdb/test_binsrch_index.cpp
gdicristofaro: It looks like `idx_idx_fname` wasn't getting deallocated.  I added that to `hdb_binsrch_close`.

simsong:
Any thoughts of using the c++ unique_per class to handle reallocation
automatically?

gdicristofaro:
Hi @simsong ,
I'm using `unique_ptr` [here](https://github.com/sleuthkit/sleuthkit/pull/3130/files#diff-606f6c7a0613159f732308a00349249d677d05ec7cfa8730691b55fa4cd5abd7R18-R20).  Is that what you mean?

simsong:
> Hi @simsong , I'm using `unique_ptr` [here](https://github.com/sleuthkit/sleuthkit/pull/3130/files#diff-606f6c7a0613159f732308a00349249d677d05ec7cfa8730691b55fa4cd5abd7R18-R20). Is that what you mean?

Yes, although it is confusing that you are mixing C++ `unique_ptr` and its memory management technology with `tsk_malloc()`. I believe that @uckelman-sf  suggested that if you are going to use `unique_ptr` you can use a custom deleted so that the `tsk_free` (or whatever it is called) is called automatically when the `unique_ptr` is freed. Then you do not need to have the explicit frees.

Overall, if you lean-in to the C++ memory allocation system, it will generally eliminate the need to explicitly call `free()` anywhere. 

Another bit of confusion for me is the explicit call to `free()` in lines 1587-1589 of binsrc_index.cpp. There this is a `TSK_TCHAR` array, which annoyingly compiles as 8-bit values on POSIX and 16-bit values on WIN32/WIN64. Changing the way that TSK deal with strings internally is another thing that will need to wait for the future. Perhaps a complete rewrite into Rust will make it unnecessary.

gdicristofaro:
Hi @simsong ,

> Yes, although it is confusing that you are mixing C++ `unique_ptr` and its memory management technology with `tsk_malloc()`. I believe that @uckelman-sf suggested that if you are going to use `unique_ptr` you can use a custom deleted so that the `tsk_free` (or whatever it is called) is called automatically when the `unique_ptr` is freed. Then you do not need to have the explicit frees.

In the proposed code,
```
std::unique_ptr<hdb_binsrch_info, decltype(&hdb_binsrch_close)> info{
  new hdb_binsrch_info{},
  htb_binsrch_close
};
```

the deleter, [`hdb_binsrch_info`](https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/hashdb/binsrch_index.cpp#L1574-L1609) uses `free`'s.  I assumed the reference to `new` was an oversight as I believe `new` and `free` [shouldn't be mixed](https://isocpp.org/wiki/faq/freestore-mgmt#mixing-malloc-and-delete) and also causes address sanitizer problems in the tests:

```
   ==48077==ERROR: AddressSanitizer: alloc-dealloc-mismatch (operator new vs free) on 0x518000000880
    #0 0x7ffa598fc4d8 in free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:52
    #1 0x561881a440c3 in hdb_binsrch_close tsk/hashdb/binsrch_index.cpp:1611
    #2 0x5618814d7593 in hdb_binsrch_index_close(TSK_HDB_BINSRCH_INFO*) test/tsk/hashdb/test_binsrch_index.cpp:7
    ...
```

[Link here](https://github.com/sleuthkit/sleuthkit/actions/runs/12324722475/job/34402772042)

Of course, if I misunderstood the ask, please let me know.

> Another bit of confusion for me is the explicit call to `free()` in lines 1587-1589 of binsrc_index.cpp.

I was trying to fix an issue that I identified in expanding the testing scope to also include `hdb_binrch_close` as proposed with `unique_ptr`.  For the same reason regarding the mixing of `free` and `new`, I added `free` here to solve this memory leak.  Within the close function, all other fields use `free` to deallocate, and there are other places in the codebase outside these tests and the functions under test that are calling `tsk_malloc` to allocate these fields.

simsong:
Hi @gdicristofaro ,
You are correct that `new` and `free` should not be mixed! And it's great that you caught the bug with AddressSanitizer. That's why we are using all of this CI/CD stuff, and why we are all taking time with this. 

You didn't misunderstand the ask. We just have a significant code base here and several people trying to improve it. I'm happy to go with your code as-is here and move to more automation on the memory allocation side later.

 I just wanted to bring the whole thing to your attention, but apparently it was already there!

Thanks again for your work on this.

simsong (approved):

gdicristofaro:
Hi @simsong , absolutely. Sounds good. Yes, I would agree.  Having a net of automated tests that can catch problems and regressions is certainly advantageous.

uckelman-sf:
My mistake, I hadn't remembered what was inside `htb_binsrch_close` when I wrote the example. Tthe creation side of `std::unique_ptr` should be using whatever you pair with `htb_binsrch_close`.

PR #3134, opened 2024-12-17
Fix warnings which arise when compiling with C++20

uckelman-sf:
Note that this PR does not switch to compiling with C++20---it only corrects warnings you'd get if you did.


uckelman-sf:
@simsong Ready for review.

#3135 follows this PR.

simsong:
AppVeyor hasn't even started mac, but Windows ran to completion and that's all SKL cares about...

PR #3135, opened 2024-12-17
Don't clobber user variables in configure.ac

uckelman-sf:
The user variables (`CPPFLAGS`, `CFLAGS`, `CXXFLAGS`, `LDFLAGS`) shouldn't be reset in `configure.ac`, as that defeats their purpose as user overrides, e.g., by interfering with setting `-Werror` for builds.


uckelman-sf:
#3136 follows this PR.

simsong (approved):

PR #3136, opened 2024-12-17
Use C++20

uckelman-sf:


simsong:
Failing on AppVeyor because:
image

simsong:
@uckelman-sf  - I suggest moving AppVeyor from `Ubuntu` to `Ubuntu2204` image. See: https://www.appveyor.com/docs/linux-images-software/

simsong (approved):
thanks.

simsong:
appVeyor continues to be annoying. 
image

uckelman-sf:
Apparently apt isn't working.

simsong (approved):
I keep learning new things. Thanks for telling me about this rather odd setting.

uckelman-sf:
> I keep learning new things. Thanks for telling me about this rather odd setting.

I hadn't seen the error before today. Something is out-of-date, but I'm not sure if the fault is with AppVeyor's image or one of the repos they're using.

uckelman-sf:
The MacOS image was also too old. That's fixed now. I think this is working.

uckelman-sf:
@simsong This is finished, and everything else I'm doing is waiting on it.

simsong:
@uckelman-sf - Does Visual Studio 2019 support C++20? We need to be able to run TSK executables on Windows XP. I'm told that Visual Studio 2019 is the last version that supports Windows XP.

We need to add a validation for Windows XP to our test suite.

simsong:
I don't think that we can take this until we incorporate Windows XP testing.
https://chatgpt.com/share/67883d90-14b4-8010-82c0-f3223ec9aeb5

simsong:
> @simsong This is finished, and everything else I'm doing is waiting on it.

Hi @uckelman-sf. Unfortunately, we do not have a way to support C++20 on Windows XP.
We require the ability to run TSK executables on Windows XP machines, as there are many Windows XP machines still in use an TSK is the only way to acquire from a machine that might be infected with a root kit. 

It may be possible to use Cygwin to compile the Windows XP executable that uses C++20. But that is too big of a change given how close the next release is. 

So we need to close this PR. We cannot accept it at this point, and we cannot migrate to any version of C++ that cannot be compiled by VisualStudio 2019, the last version of Visual Studio to support Windows XP.

uckelman-sf:
Can't anyone who needs to run on Windows XP use an older version of TSK? E.g., I would expect the current release to be adequate for that. It's not like you're going to encounter any modern filesystems on Windows XP.

simsong:
> Can't anyone who needs to run on Windows XP use an older version of TSK? E.g., I would expect the current release to be adequate for that. It's not like you're going to encounter any modern filesystems on Windows XP.

My understanding is that Cyber Triage pushes a single executable to a Windows system being analyzed. It doesn't make sense to freeze that executable and to add new features to the current executable.

At least we don't need to *compile* on Windows XP. But we must be able to run on it. So the only way I see forward is moving to compiling the Windows executable with mingw. I personally think that's a better approach, but to know for sure we will need to compare the two executables and do a lot of testing and we're just not able to do that at this point in the release cycle.

This may be resolved later, but we can't resolve it before the next release, which is Friday, with feature freeze this Sunday morning, roughly 36 hours from now.

uckelman:
Where's the evidence that C++20 Windows executables won't run on XP? Do you have a reference for this?

uckelman:
I don't understand the release process, btw. This could use some transparency. I had been asking for a release already in November.

simsong:
> I don't understand the release process, btw. This could use some transparency. I had been asking for a release already in November.

Well, we're getting a release! I now have documented it here:
https://github.com/sleuthkit/sleuthkit/blob/develop/release/Release_Process.md

simsong:
> Where's the evidence that C++20 Windows executables won't run on XP? Do you have a reference for this?

Microsoft: https://learn.microsoft.com/en-us/cpp/build/configuring-programs-for-windows-xp?view=msvc-170

Specifically:

> Visual Studio supports multiple platform toolsets. That means it's possible to target operating systems and runtime libraries that aren't supported by the default toolset. For example, by switching the platform toolset, you can use the Visual Studio 2017 C++ compiler to create apps that target Windows XP and Windows Server 2003. You can also use older platform toolsets to maintain binary-compatible legacy code and still take advantage of the latest features of the Visual Studio IDE.

We also need a Windows XP test environment, something I've asked SKL for. Ideally we could run it in Docker on GitHub.

PR #3138, opened 2024-12-17
Image block cache refactoring

uckelman-sf:


simsong:
Hi. I'm curious about the use of the C++ timers in here. Do you have thoughts on this, as to why they are being used, and what they are for?

uckelman-sf:
> Hi. I'm curious about the use of the C++ timers in here. Do you have thoughts on this, as to why they are being used, and what they are for?

They're for collecting caching stats.

simsong:
But they just give wall time, right? Wouldn't CPU performance registers or os time functions be more accurate?

uckelman-sf:
> But they just give wall time, right? Wouldn't CPU performance registers or os time functions be more accurate?

CPU performance registers or os time functions don't measure wall clock time, which is what I want here.

PR #3147, opened 2025-01-03
AUT-2487 bitlocker in autopsy

gdicristofaro:


bindings/java/jni/dataModel_SleuthkitJNI.cpp
```diff
@@ -824,13 +843,31 @@ JNIEXPORT jlong JNICALL
...
+ * @param timeZone The time zone for the image.
+ * @param addFileSystems Pass true to attempt to add file systems within the image to the case database.
+ * @param addUnallocSpace Pass true to create virtual files for unallocated space. Ignored if addFileSystems is false.
+ * @param skipFatFsOrphans Pass true to skip processing of orphan files for FAT file systems. Ignored if addFileSystems is false.
+ * @param passwordJ Password for the file system or null for no password.
+ *
+ * @return A pointer to the process (TskAutoDbJava object) or NULL on error.
+ */
+JNIEXPORT jlong JNICALL
+Java_org_sleuthkit_datamodel_SleuthkitJNI_initializeAddImgNatPassword(JNIEnv * env, jclass obj,
```
APriestman, line 869: I would put the "Nat" at the end for naming consistency

tsk/auto/tsk_auto.h
```diff
@@ -108,7 +108,12 @@ class TskAuto {
     void setFileSystemPassword(std::string fileSystemPassword) { m_fileSystemPassword = fileSystemPassword; }
 
     /**
-     * TskAuto calls this method before it processes the volume system that is found in an
+     * @returns A password that will be used when trying to open each file system or empty.
+     */
+    std::string getFileSystemPassword() { return m_fileSystemPassword; }
```
uckelman-sf, line 113: Is there a reason you're not returning a `const std::string&` here and making the function `const`?

gdicristofaro: I'm returning a `std::string` for consistency with other TSK string getters.  Adding `const` seems like a good idea that I missed.  Thanks for catching that.  It should be updated.

APriestman (approved):

tsk/auto/tsk_auto.h
uckelman-sf: It would be better to return `const std::string&` for all of those than to return a copy; the caller can always make a copy if one is needed.

gdicristofaro: Sure.  I think it would be worth creating a separate issue/PR to address [those instances](https://github.com/search?q=repo%3Asleuthkit%2Fsleuthkit+%28path%3A*.cpp+OR+path%3A*.c+OR+path%3A*.hpp+OR+path%3A*.h%29+%22string+get%22&type=code&ref=advsearch) simultaneously for consistency in API.

simsong:
What about the arguments here? is returning a const std::string really slower than non-const?stackoverflow.comWhen I did bulk_extractor, I ran into some problems returning const string values.----Sent from my phone.On Jan 15, 2025, at 11:44 AM, Joel Uckelman ***@***.***> wrote:﻿
@uckelman-sf commented on this pull request.



In tsk/auto/tsk_auto.h:
> @@ -108,7 +108,12 @@ class TskAuto {
     void setFileSystemPassword(std::string fileSystemPassword) { m_fileSystemPassword = fileSystemPassword; }
 
     /**
-     * TskAuto calls this method before it processes the volume system that is found in an
+     * @returns A password that will be used when trying to open each file system or empty.
+     */
+    std::string getFileSystemPassword() { return m_fileSystemPassword; }

It would be better to return const std::string& for all of those than to return a copy; the caller can always make a copy if one is needed.

simsong:
> @simsong You should absolutely not return a `const std::string` (or any `const` raw type). I'm suggesting returning a `const std::string&` (note the `&` here), which avoids copying.

Got it. 

In some of my code, I was returning a `const std::string &` but it turned out that I may have been returning it to a value that was dynamic, and I was getting memory errors. I avoided this just by returning a `std::string`, but I was not keeping a copy internally, I think.

tsk/auto/tsk_auto.h
```diff
@@ -108,7 +108,12 @@ class TskAuto {
     void setFileSystemPassword(std::string fileSystemPassword) { m_fileSystemPassword = fileSystemPassword; }
 
     /**
-     * TskAuto calls this method before it processes the volume system that is found in an
+     * @returns A password that will be used when trying to open each file system or empty.
+     */
+    std::string getFileSystemPassword() const { return m_fileSystemPassword; }
+
```
simsong, line 114: Do we want to change this to a `const std::string &` ?

simsong:
I wanted to look at the codecov report for this and I saw that it didn't run because the codecov token wasn't uploaded:
image

@uckelman-sf - any idea why it wasn't run?

uckelman-sf:
> In some of my code, I was returning a `const std::string &` but it turned out that I may have been returning it to a value that was dynamic

No, you shouldn't return a `const std::string&` when the variable being returned is a temporary. That's not the case here; the strings in question here are all members, not temporaries.

uckelman-sf:
> @uckelman-sf - any idea why it wasn't run?

No idea. How long has that been happening?

uckelman-sf:
I believe there's some supported way to handle the codecov secret for public repos. I'm looking into it now.

uckelman-sf:
@simsong Look at #3151.

tsk/auto/tsk_auto.h
uckelman-sf: Yes, it should be.

uckelman-sf:
PR #3152 fixes the other places where `std::string` is copied unnecessarily in `TskAuto`.

tsk/util/Bitlocker/BitlockerParser.cpp
```diff
@@ -1270,21 +1270,25 @@ int BitlockerParser::decryptSector(TSK_DADDR_T volumeOffset, uint8_t* data) {
         return 0;
     }
 
+    tsk_take_lock(&m_decrypt_sector_lock);
```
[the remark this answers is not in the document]
gdicristofaro: Sure.  Created an issue here to resolve that: https://github.com/sleuthkit/sleuthkit/issues/3154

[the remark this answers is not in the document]
simsong: @uckelman-sf  - we are under pressure to get out the next release, so we are going to put a hold on accepting non-critical PRs until I branch for the release (which will happen over the weekend).

simsong:
> It looks like the changes requested are outside the scope of this PR. Please raise an issue to address your concerns regarding refactoring the other string getter functions. We do not have the bandwidth to address this non-critical path issue and assess the impact of the changes. Please merge this so we can get TSK release out.

Done.

PR #3149, opened 2025-01-10
Image block cache refactoring

uckelman-sf:
This PR disentangles the existing image block cache from things which use it intrusively, in preparation for replacing the block cache with something more performant.


uckelman-sf:
@simsong Ready for review.

simsong:
> @simsong Ready for review.

Do you want to change the cache so that it uses a hashmap rather than a linear search, and perhaps a priority queue for eviction?

PR #3151, opened 2025-01-16
Try using codecov action for coverage stats upload

uckelman-sf:


simsong:
Now if only I could make use of the ossfuzz.... Perhaps @joachimmetz can tell me how to do that.

joachimmetz:
> Now if only I could make use of the ossfuzz.... Perhaps @joachimmetz can tell me how to do that.

I'm missing the context, how is OSSFuzz related to codecov? How do you want to make use of OSSFuzz?

uckelman-sf:
I'm not sure how this could fix #3157. That's not a PR for fixing a bug.

PR #3152, opened 2025-01-16
Use const std::string& where possible to avoid unnecessary copies

uckelman-sf:


gdicristofaro:
It looks like these are other candidates for change:
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/Cell.h#L118
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/NKRecord.h#L98
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/NKRecord.h#L118
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/RegistryByteBuffer.h#L53
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/RegistryByteBuffer.h#L56
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/RegistryValue.h#L58
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/ValueData.h#L86
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/VKRecord.h#L105
https://github.com/sleuthkit/sleuthkit/blob/develop/tools/logicalimager/FileExtractor.h#L32
https://github.com/sleuthkit/sleuthkit/blob/develop/rejistry%2B%2B/src/RegistryHiveFile.h#L57
https://github.com/sleuthkit/sleuthkit/blob/develop/tools/logicalimager/LogicalImagerConfiguration.h#L43
https://github.com/sleuthkit/sleuthkit/blob/develop/tools/logicalimager/RegFileInfo.h (4 getters)
https://github.com/sleuthkit/sleuthkit/blob/develop/tools/logicalimager/RegKey.h#L30
https://github.com/sleuthkit/sleuthkit/blob/develop/tools/logicalimager/RegVal.h ( 4 getters)
https://github.com/sleuthkit/sleuthkit/blob/develop/tools/logicalimager/UserAccount.h (11 getters)
https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/util/Bitlocker/BitlockerParser.h (3 other getters)

uckelman-sf:
The functions in `BitlockerParser` which return a `std::string` are doing so correctly, as they're all returning a temporary, not a member.

I'll leave the others for someone else to pursue.

@simsong Ready for review.

gdicristofaro:
My mistake.  I updated the list accordingly.

> The functions in `BitlockerParser` which return a `std::string` are doing so correctly, as they're all returning a temporary, not a member.
> 
> I'll leave the others for someone else to pursue.
> 
> @simsong Ready for review.

simsong:
I'm going to accept this after the branch.

simsong:
This is a great list. I'm going to hold off on accepting it, though, because I've seen this cause crashing problems in the past. For example, you might return the reference and then some time later in code executing the object that owns the reference gets deleted. We don't have a borrow checker the way that Rust does, and we just haven't tested this well enough. It's also not a super huge performance issue - this is an elegance issue. So I'm putting this on hold until after the branch for the release.

simsong:
Now that we did a branch for release 4.13, I can accept this.

simsong:
This was unexpected:
image

uckelman-sf:
I'm looking into it.

uckelman-sf:
See PR #3162.

PR #3153, opened 2025-01-16
CT-9539 Improve logical file caching and fix bug

APriestman:


uckelman-sf:
How did you find the bug? Would you mind adding the test to the test suite?

APriestman:
This only runs on a folder on Windows so there are no tests. I found the bug because when I merged in develop suddenly it took ages to run fls.

crayy8 (approved):

simsong:
We have tests on Windows. How can we create a test to catch this?

simsong (approved):

PR #3161, opened 2025-01-20
Fix leak of TSK_VS_INFO in TskAuto

uckelman-sf:
Fixed memory leak. TSK_VS_INFO was not closed on every return path.

### DIFF
--- a/tsk/auto/auto.cpp
+++ b/tsk/auto/auto.cpp
@@ -17,6 +17,7 @@
 #include "tsk/fs/tsk_fatxxfs.h"
 #include "tsk/img/img_writer.h"
 
+#include <memory>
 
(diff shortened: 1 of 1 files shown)


simsong:
Why wasn't the memory leak revealed by our regression tests? On bulk_extractor memory leaks were.

simsong:
Also, I want to either get value from ossfuzz or take it out. My understanding is that all of the reports are going to some server somewhere that I can't access. Thoughts?
image

uckelman-sf:
> Why wasn't the memory leak revealed by our regression tests? On bulk_extractor memory leaks were.

The tests didn't hit this code path.

simsong (approved):
Nice job. Thanks.

PR #3172, opened 2025-01-25
Changes to get fuzz targets building again after img_info cache member changes

joachimmetz:


joachimmetz:
@simsong PTAL this should get the fuzz targets building again after the changes to img_info

uckelman:
This is why it would be better to have the fuzz target build on PRs if we keep it---we cannot tell if PRs break it this way.

joachimmetz:
> This is why it would be better to have the fuzz target build on PRs if we keep it---we cannot tell if PRs break it this way.

What matters is that you detect and fix breakage before release. Now when and where you detect this is up to you. You should be able to update https://github.com/sleuthkit/sleuthkit/blob/develop/.github/workflows/build_ossfuzz.yml and have it build from the PR branch, it is merely a matter of changing the branch in the Docker file before `python3 infra/helper.py build_image --pull sleuthkit`.

Also you made an API breaking change and did not anticipate things will break? What about code that is not part of this repo?

uckelman:
>  You should be able to update https://github.com/sleuthkit/sleuthkit/blob/develop/.github/workflows/build_ossfuzz.yml and have it build from the PR branch, it is merely a matter of changing the branch in the Docker file before python3 infra/helper.py build_image --pull sleuthkit.

Would you make a PR for that?

joachimmetz:
> Would you make a PR for that?

Maybe document the test strategy first. I'm happy to talk to @simsong about this. Give your previous attitude I have no intention of helping you directly.

uckelman:
> Maybe document the test strategy first. 

Would you explain what you mean by "the test strategy" so that I can give an answer that addresses your question? 

What I would describe as "the test strategy" is: We should have sufficient tests which run on PRs to prevent bugs from being introduced to the main branch. If that's not enough detail, I'd like to know specifically where I can add more to address your concerns.

Happy to help!

simsong:
> > Maybe document the test strategy first.
> 
> Would you explain what you mean by "the test strategy" so that I can give an answer that addresses your question?
> 
> What I would describe as "the test strategy" is: We should have sufficient tests which run on PRs to prevent bugs from being introduced to the main branch. If that's not enough detail, I'd like to know specifically where I can add more to address your concerns.
> 
I think that we need to formalize testing. I've already been confused regarding the difference between unit-tests, end-to-end tests, performance tests, conformance test, regression tests, etc. Right now we have some tests in various directories but it's not clear to me what is testing what and what are the goals of different tests.

I found this document online:

https://en.m.wikipedia.org/wiki/Test_strategy

Perhaps we could create a test strategy document in our docs directory and then begin to formalize this all?

If I start on one, could you help with it?

simsong:
@joachimmetz — thank you for this. Can you re-enable `build_ossfuzz` in this PR so that we can test it? It's currently skipped....  To re-enable it, turn the `false` to `true` here: 

https://github.com/sleuthkit/sleuthkit/blob/36039422094b6b03f8727d32f352e44803d2c833/.github/workflows/build_ossfuzz.yml#L14

simsong:
> > This is why it would be better to have the fuzz target build on PRs if we keep it---we cannot tell if PRs break it this way.
> 
> What matters is that you detect and fix breakage before release. Now when and where you detect this is up to you. You should be able to update https://github.com/sleuthkit/sleuthkit/blob/develop/.github/workflows/build_ossfuzz.yml and have it build from the PR branch, it is merely a matter of changing the branch in the Docker file before `python3 infra/helper.py build_image --pull sleuthkit`.
> 
> Also you made an API breaking change and did not anticipate things will break? What about code that is not part of this repo?

This is really my fault for not realizing what was happening with the ossfuzz system and not catching the breakage before accepting the PR.

uckelman:
> Perhaps we could create a test strategy document in our docs directory and then begin to formalize this all?

I'd be happy to write up something when I'm in on Monday. I think it won't require more than a few sentences, though.

simsong:
That would be really useful. There is no rush. We now have an official decision to move much of the documentation to the docs directory, except for the documentation that is automatically generated. That should go on GitHub pages.

joachimmetz:
> @joachimmetz — thank you for this. Can you re-enable build_ossfuzz in this PR so that we can test it? It's currently skipped.... To re-enable it, turn the false to true here:

@simsong let's do this in 2 steps, otherwise the CI tests will fail to complete again

let's use a separate PR to discuss if this should be an on-PR or daily CI test

simsong:
> > @joachimmetz — thank you for this. Can you re-enable build_ossfuzz in this PR so that we can test it? It's currently skipped.... To re-enable it, turn the false to true here:
> 
> @simsong let's do this in 2 steps, otherwise the CI tests will fail to complete again
> 
> let's use a separate PR to discuss if this should be an on-PR or daily CI test

I'm fine. So do you want me to go ahead and merge this in, even though ossfuzz is disabled?

joachimmetz:
> I'm fine. So do you want me to go ahead and merge this in, even though ossfuzz is disabled?

To be explicit is is only the CI test that is disabled. OSSFuzz runs independently of this project.

Yes fine to proceed with merging the changes,

PR #3175, opened 2025-01-25
Update api version

simsong:
Because of the change to the `img` structure, we need to update the LD version in Makefile.am


simsong:
this is no longer relevant.

PR #3176, opened 2025-01-25
updated version to indicate that API has changed in a backward incompatible way for programs that link with libtsk and reference the members of the TSK_IMG_INFO * structure that is returned by the tsk_img_open_external() function all

simsong:
…

### DIFF
--- a/.github/workflows/build_ossfuzz.yml
+++ b/.github/workflows/build_ossfuzz.yml
@@ -11,6 +11,7 @@ on:
 permissions: read-all
 jobs:
   build_ossfuzz:
+    if: false
     runs-on: ubuntu-22.04
     strategy:
       matrix:
--- a/Makefile.am
+++ b/Makefile.am
@@ -369,7 +369,7 @@ tsk_libtsk_la_LIBADD = \
 	tsk/vs/libtskvs.la
 
 # current:revision:age
-tsk_libtsk_la_LDFLAGS = $(AM_LDFLAGS) -version-info 21:1:2 $(LIBTSK_LDFLAGS)
+tsk_libtsk_la_LDFLAGS = $(AM_LDFLAGS) -version-info 22:0:0 $(LIBTSK_LDFLAGS)
 
 EXTRA_DIST += \
 	tsk/tsk_tools_i.h \


simsong:
@jayaramcs  — Want to make sure you see this. The change in `tsk_img_open_external` added an argument, so we need to bump the LD version number.

uckelman-sf:
Which change to `tsk_img_open_external` is this referring to? Its signature in 4.12.1 is this:
```
    extern TSK_IMG_INFO *tsk_img_open_external(void* ext_img_info,
        TSK_OFF_T size, unsigned int sector_size,
        ssize_t(*read) (TSK_IMG_INFO * img, TSK_OFF_T off, char *buf, size_t len),
        void (*close) (TSK_IMG_INFO *),
        void (*imgstat) (TSK_IMG_INFO *, FILE *));
```
Its signature in `develop` right now is this:
```
    extern TSK_IMG_INFO *tsk_img_open_external(void* ext_img_info,
        TSK_OFF_T size, unsigned int sector_size,
        ssize_t(*read) (TSK_IMG_INFO * img, TSK_OFF_T off, char *buf, size_t len),
        void (*close) (TSK_IMG_INFO *),
        void (*imgstat) (TSK_IMG_INFO *, FILE *));
```
Those are the same.

simsong:
> Which change to `tsk_img_open_external` is this referring to? Its signature in 4.12.1 is this:
> 
> ```
>     extern TSK_IMG_INFO *tsk_img_open_external(void* ext_img_info,
>         TSK_OFF_T size, unsigned int sector_size,
>         ssize_t(*read) (TSK_IMG_INFO * img, TSK_OFF_T off, char *buf, size_t len),
>         void (*close) (TSK_IMG_INFO *),
>         void (*imgstat) (TSK_IMG_INFO *, FILE *));
> ```
> 
> Its signature in `develop` right now is this:
> 
> ```
>     extern TSK_IMG_INFO *tsk_img_open_external(void* ext_img_info,
>         TSK_OFF_T size, unsigned int sector_size,
>         ssize_t(*read) (TSK_IMG_INFO * img, TSK_OFF_T off, char *buf, size_t len),
>         void (*close) (TSK_IMG_INFO *),
>         void (*imgstat) (TSK_IMG_INFO *, FILE *));
> ```
> 
> Those are the same.

They are typographically the same, but they refer to different definitions of `TSK_IMG_INFO` structure.

uckelman-sf:
> They are typographically the same, but I believe that they refer to different definitions of TSK_IMG_INFO structure.

The `TSK_IMG_INFO*` are all pointers. They are actually the same. If this issue is the struct changing (which is not a pointer) maybe the title of the issue should be updated to reflect that? Otherwise it will be confusing for anyone looking back at this.

simsong:
> > They are typographically the same, but I believe that they refer to different definitions of TSK_IMG_INFO structure.
> 
> The `TSK_IMG_INFO*` are all pointers. They are actually the same. If this issue is the struct changing (which is not a pointer) maybe the title of the issue should be updated to reflect that? Otherwise it will be confusing for anyone looking back at this.

They are pointers, but they are referencable. I'll update the title.

simsong:
I think that this may be mistaken and that the possibly API-breaking changes I was concerned about are in Develop, and not in `release-4.13.0`. So I'm closing this.

PR #3178, opened 2025-01-27
Move members of TSK_IMG_INFO intended to be private to their own struct

uckelman-sf:
The non-metadata parts of `TSK_IMG_INFO` shouldn't be public. This PR moves them to an "inner" struct so they don't appear in any public headers.


simsong:
Thanks for doing this. This is what I meant by putting them somewhere private.

PR #3189, opened 2025-02-14
Ensure that libtsk is actually linked against its dependencies

uckelman-sf:
libtsk's dependent libs must be listed in tsk_libtsk_la_LIBADD, or the linker might produce a library which isn't linked to any of them. (This happens on MacOS, e.g.)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -130,6 +130,9 @@ nobase_dist_data_DATA = \
 
 TSK_LIBS = \
 	tsk/libtsk.la \
+	$(TSK_LIB_LIBS)
+
+TSK_LIB_LIBS = \
 	$(PTHREAD_LIBS) \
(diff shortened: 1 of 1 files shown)


sf-jonstewart:
how can we best add a test for this in CI, though? the time spent figuring this out is fine, so long as we never have this regression again.

simsong:
> how can we best add a test for this in CI, though? the time spent figuring this out is fine, so long as we never have this regression again.

I think that we need a linter for autoconf inputs.  I found discussion about this issue here — https://www.mail-archive.com/autoconf@gnu.org/msg23522.html — but I do not think that any such linter exists.

simsong:
Moved into RC2. See https://github.com/sleuthkit/sleuthkit/discussions/3184

PR #3195, opened 2025-02-18
Avoid defining bool datatype

hillu:
This avoids build failures with GCC 15,
see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1097897


tsk/base/crc.h
```diff
@@ -82,6 +82,8 @@ Status  : Copyright (C) Ross Williams, 1993. However, permission is
 #ifndef CM_DONE
 #define CM_DONE
 
+#include <stdbool.h>
+
```
simsong, line 86: I guess we assume that this header exists because it is in C99?

simsong (approved):

Issue #3198, opened 2025-02-23
tsk_UTF16toUTF8 OOB read


joachimmetz:
**Describe the bug or problem**

out-of-bounds read of 1 or 2 flagged by OSSFuzz https://issues.oss-fuzz.com/issues/42520336 in Jan 2023

**To Reproduce**

The version of SleuthKit you are using: https://github.com/sleuthkit/sleuthkit/commit/b6e53ee6d5e992c764eedaccf7b2a3d46091bf3c

Where you got it from:  git checkout

How you compiled it:

```
./boostrap && ./configure -g --enable-shared=no && make
```

valgrind ./tools/fstools/fls clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336

```
==871903== Memcheck, a memory error detector
==871903== Copyright (C) 2002-2024, and GNU GPL'd, by Julian Seward et al.
==871903== Using Valgrind-3.24.0 and LibVEX; rerun with -h for copyright info
==871903== Command: ./tools/fstools/fls /home/lordyesta/Downloads/clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336
==871903== 
==871903== Invalid read of size 1
==871903==    at 0x451523: tsk_UTF16toUTF8 (tsk_unicode.c:163)
==871903==    by 0x43BF6C: ntfs_dent_copy_short_only (ntfs_dent.cpp:316)
==871903==    by 0x43BF6C: ntfs_proc_idxentry(NTFS_INFO*, TSK_FS_DIR*, unsigned char, ntfs_idxentry*, unsigned int, unsigned int) (ntfs_dent.cpp:541)
==871903==    by 0x43CF61: ntfs_dir_open_meta (ntfs_dent.cpp:1250)
==871903==    by 0x41A0D4: tsk_fs_dir_open_meta_internal(TSK_FS_INFO*, unsigned long, int) (fs_dir.cpp:308)
==871903==    by 0x41A22B: tsk_fs_dir_walk_recursive(TSK_FS_INFO*, DENT_DINFO*, unsigned long, TSK_FS_DIR_WALK_FLAG_ENUM, TSK_WALK_RET_ENUM (*)(TSK_FS_FILE*, char const*, void*), void*, int) (fs_dir.cpp:661)
==871903==    by 0x41AF14: tsk_fs_dir_walk_internal (fs_dir.cpp:1001)
[...]

PR #3199, opened 2025-02-23
Create tsk_safeUTF16toUTF8 to Prevent Out of Bounds (OOB) read in ntfs_dent_copy_short_only and ntfs_proc_attrseq

joachimmetz:
Introduces `tsk_safeUTF16toUTF8`

### DIFF
--- a/tsk/base/tsk_unicode.c
+++ b/tsk/base/tsk_unicode.c
@@ -135,7 +135,7 @@ static const UTF8 firstByteMark[7] =
 /**
  * \ingroup baselib
  * Convert a UTF-16 string to UTF-8.
- * @param endian Endian ordering flag of UTF-16 text
+ * @param endian Byte ordering of UTF-16 text
  * @param sourceStart Pointer to pointer to start of UTF-16 string.  Will be updated to last char processed.
  * @param sourceEnd Pointer to one entry past end of UTF-16 string
  * @param targetStart Pointer to pointer to place where UTF-8 string should be written.  Will be updated to next place to write to.
@@ -250,6 +250,42 @@ tsk_UTF16toUTF8(TSK_ENDIAN_ENUM endian, const UTF16 ** sourceStart,
     *sourceStart = source;
     *targetStart = target;
     return result;
+}
+
+/**
+ * \ingroup baselib
+ * Wrapper around tsk_UTF16toUTF8 to ensure read of source remains in bounds.
+ * @param endian Byte ordering of UTF-16 text
+ * @param source Pointer to start of the buffer that contains the UTF-16 string.
+ * @param source_len Number of bytes in buffer that contains the UTF-16 string (source).
+ * @param number_of_characters Number of characters in the UTF-16 string to convert.
+ * @param targetStart Pointer to pointer to place where UTF-8 string should be written.  Will be updated to next place to write to.
+ * @param targetEnd Pointer to end of UTF-8 buffer
+ * @param flags Flags used during conversion
+ * @returns error code
+ */
+TSKConversionResult
+tsk_safeUTF16toUTF8(TSK_ENDIAN_ENUM endian, const uint8_t * source,
+    size_t source_len, size_t number_of_characters, UTF8 ** targetStart,
+    UTF8 * targetEnd, TSKConversionFlags flags)
+{
+	int retVal = TSKconversionOK;
+
+	if (source_len > 0 && source_len % 2 != 0) {
+		// source_len must be even, since UTF-16 uses 2 bytes per character
+		source_len -= 1;
+	}
+	source_len /= 2;
+
+	if (source_len < number_of_characters) {
+		number_of_characters = source_len;
+	}
+	const UTF16 *utf16_source = (const UTF16 *) source;
+
+	retVal = tsk_UTF16toUTF8(endian, &utf16_source,
+		utf16_source + number_of_characters, targetStart, targetEnd, flags);
+
+	return retVal;
 }
 
 
--- a/tsk/fs/ntfs.cpp
+++ b/tsk/fs/ntfs.cpp
@@ -24,6 +24,7 @@
 #include <memory>
 
 #include "encryptionHelper.h"
+#include "tsk/base/tsk_unicode.h"
 
 /**
  * \file ntfs.c
@@ -2437,6 +2438,13 @@ ntfs_proc_attrseq(NTFS_INFO * ntfs,
                 }
                 fs_name->next = NULL;
             }
+	    /* This check does not make sense
+	     * fname->nlen is number of UTF16 characters, 2 bytest per character
+	     * and attr_len is in bytes.
+	     *
+	     * Keeping it for review but recommend removing this in favor of
+	     * the checks doene inside tsk_safeUTF16toUTF8.
+	     *
             if (fname->nlen > attr_len - 66) {
                 tsk_error_reset();
                 tsk_error_set_errno(TSK_ERR_FS_INODE_COR);
@@ -2444,16 +2452,17 @@ ntfs_proc_attrseq(NTFS_INFO * ntfs,
                     ("proc_attrseq: invalid name value size out of bounds!");
                 return TSK_COR;
             }
-            UTF16 *name16 = (UTF16 *) & fname->name;
+	     */
             UTF8 *name8 = (UTF8 *) fs_name->name;
 
-            retVal =
-                tsk_UTF16toUTF8(fs->endian, (const UTF16 **) &name16,
-                (UTF16 *) ((uintptr_t) name16 +
-                    fname->nlen * 2),
+            retVal = tsk_safeUTF16toUTF8(
+                fs->endian,
+                & fname->name,
+                attr_len - attr_off - 66,
+                fname->nlen,
                 &name8,
-                (UTF8 *) ((uintptr_t) name8 +
-                    sizeof(fs_name->name)), TSKlenientConversion);
+                (UTF8 *) ((uintptr_t) name8 + sizeof(fs_name->name)),
+                TSKlenientConversion);
             if (retVal != TSKconversionOK) {
                 if (tsk_verbose)
                     tsk_fprintf(stderr,
--- a/tsk/fs/ntfs_dent.cpp
+++ b/tsk/fs/ntfs_dent.cpp
@@ -303,21 +303,26 @@ ntfs_dent_copy(NTFS_INFO * ntfs, ntfs_idxentry * idxe, uintptr_t endaddr,
  * No other fields are copied.  Just the name into shrt_name. */
 static uint8_t
 ntfs_dent_copy_short_only(NTFS_INFO * ntfs, ntfs_idxentry * idxe,
-    TSK_FS_NAME * fs_name)
+    uint32_t idxe_len, TSK_FS_NAME * fs_name)
 {
     ntfs_attr_fname *fname = (ntfs_attr_fname *) & idxe->stream;
     TSK_FS_INFO *fs = (TSK_FS_INFO *) & ntfs->fs_info;
-    UTF16 *name16;
     UTF8 *name8;
     int retVal;
 
-    name16 = (UTF16 *) & fname->name;
     name8 = (UTF8 *) fs_name->shrt_name;
 
-    retVal = tsk_UTF16toUTF8(fs->endian, (const UTF16 **) &name16,
-        (UTF16 *) ((uintptr_t) name16 +
-            fname->nlen * 2), &name8,
-        (UTF8 *) ((uintptr_t) name8 +
+    if (idxe_len < 16) {
+        *name8 = '\0';
+        if (tsk_verbose)
+            tsk_fprintf(stderr,
+                "Error converting NTFS 8.3 name to UTF8: remaining idxe_len too small.");
+	return 0;
+    }
+    // Note that NTFS uses UCS-2 (not UTF-16) hence lenient conversion is needed.
+    retVal = tsk_safeUTF16toUTF8(fs->endian,
+        &fname->name, (size_t) idxe_len - 16, fname->nlen,
+       	&name8, (UTF8 *) ((uintptr_t) name8 +
             fs_name->shrt_name_size), TSKlenientConversion);
 
     if (retVal != TSKconversionOK) {
@@ -370,8 +375,8 @@ is_time(uint64_t t)
  * Process a list of index entries and add to FS_DIR
  *
  * @param a_is_del Set to 1 if these entries are for a deleted directory
- * @param idxe Buffer with index entries to process
- * @param idxe_len Length of idxe buffer (in bytes)
+ * @param a_idxe Buffer with index entries to process
+ * @param a_idxe_len Length of idxe buffer (in bytes)
  * @param used_len Length of data as reported by idexlist header (everything
  * after which and less then idxe_len is considered deleted)
  *
@@ -475,9 +480,8 @@ ntfs_proc_idxentry(NTFS_INFO * a_ntfs, TSK_FS_DIR * a_fs_dir,
 
         /* do some sanity checks on the deleted entries
          */
-        if ((tsk_getu16(fs->endian, a_idxe->strlen) == 0) ||
-            (((uintptr_t) a_idxe + tsk_getu16(fs->endian,
-                        a_idxe->idxlen)) > endaddr_alloc)) {
+	uint16_t idxlen = tsk_getu16(fs->endian, a_idxe->idxlen);
+        if ((idxlen == 0) || (((uintptr_t) a_idxe + idxlen) > endaddr_alloc)) {
 
             /* name space checks */
             if ((fname->nspace != NTFS_FNAME_POSIX) &&
@@ -539,7 +543,10 @@ ntfs_proc_idxentry(NTFS_INFO * a_ntfs, TSK_FS_DIR * a_fs_dir,
             if (fs_name_preventry) {
                 // check its the same entry and if so, add short name
                 if (fs_name_preventry->meta_addr == tsk_getu48(fs->endian, a_idxe->file_ref)) {
-                    ntfs_dent_copy_short_only(a_ntfs, a_idxe, fs_name_preventry);
+		    if ((uintptr_t) a_idxe < endaddr_alloc) {
+		        size_t remaining_a_idxe_len = (size_t) ( endaddr_alloc - (uintptr_t) a_idxe );
+                        ntfs_dent_copy_short_only(a_ntfs, a_idxe, remaining_a_idxe_len, fs_name_preventry);
+		    }
                 }
 
                 // regardless, add preventry to dir and move on to next entry.
--- a/tsk/base/tsk_unicode.h
+++ b/tsk/base/tsk_unicode.h
@@ -139,6 +139,11 @@ extern "C" {
         const UTF16 ** sourceStart, const UTF16 * sourceEnd,
         UTF8 ** targetStart, UTF8 * targetEnd, TSKConversionFlags flags);
 
+    extern TSKConversionResult tsk_safeUTF16toUTF8(TSK_ENDIAN_ENUM,
+        const uint8_t *source, size_t source_len,
+	size_t number_of_characters,
+        UTF8 ** targetStart, UTF8 * targetEnd, TSKConversionFlags flags);
+
     extern TSKConversionResult
         tsk_UTF16toUTF8_lclorder(const UTF16 ** sourceStart,
         const UTF16 * sourceEnd, UTF8 ** targetStart,


joachimmetz:
@simsong FYI the test output is too cryptic to make sense of it

```
FAIL test/fiwalk/fiwalk_test (exit status: 2)

make[2]: *** [test-suite.log] Error 1
make[1]: *** [check-TESTS] Error 2
make: *** [check-am] Error 2
============================================================================
Testsuite summary for sleuthkit 4.12.1
============================================================================
# TOTAL: 9
# PASS:  8
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
============================================================================
See ./test-suite.log for debugging.
============================================================================
```

joachimmetz:
looks like the root of the issue could be an unchecked cast of 8-bit to 16-bit before the call to tsk_UTF16toUTF8

simsong:
> https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L475

What checks should we add?

joachimmetz:
> What checks should we add?

a check to prevent OOB read, but https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L477 deliberately allows an entry exceeding endaddr_alloc to continue. So I'm wondering what the intended behavior is

simsong:
> > What checks should we add?
> 
> a check to prevent OOB read, but https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L477 deliberately allows an entry exceeding endaddr_alloc to continue. So I'm wondering what the intended behavior is

Well, according to the [git blame](https://github.com/sleuthkit/sleuthkit/blame/develop/tsk/fs/ntfs_dent.cpp#L477), that is 17-year-old code imported from the CVS repo by @bcarrier . So I do not think that there may be intended behavior, and if there was intended behavior, it may no longer be intended.

It may be that this code has _never been run in a production setting_. What do you think?

joachimmetz:
possibly, unless Brian recalls what this is about. Maybe removing the "error tolerance" is the better option here, but have changed the PR, valgrind no flags OOB read on the fuzzer generated test file.

simsong:
@joachimmetz — your patch seems to work. I'm going to ask @jayaramcs to take a look, but if I don't hear from him in a few days, I'll accept.

simsong (approved):
Looks reasonable and the code no longer crashes....

simsong:
Is there any way we can get a test case?

joachimmetz:
> Is there any way we can get a test case?

Yes the POC file is here: https://oss-fuzz.com/download?testcase_id=5124116049166336 I have no idea how the test work, if you can point me to the documentation.

> your patch seems to work. I'm going to ask @jayaramcs or @APriestman take a look, but if I don't hear from either within a few days (it's Sunday here), I'll accept.

Maybe ask them if removing the code from https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L477 to https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L515 would be an option as well

simsong:
I think I will need to write the documentation. What would you like it to cover?----Sent from my phone.On Feb 23, 2025, at 12:15 PM, Joachim Metz ***@***.***> wrote:﻿
  
Is there any way we can get a test case?

Yes the POC file is here: https://oss-fuzz.com/download?testcase_id=5124116049166336 I have no idea how the new tests work, if you can point me to the documentation.

your patch seems to work. I'm going to ask @jayaramcs or @APriestman take a look, but if I don't hear from either within a few days (it's Sunday here), I'll accept.

Maybe ask them if removing the code from https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L477 to https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L515 would be an option as well—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



joachimmetz left a comment (sleuthkit/sleuthkit#3199)

Is there any way we can get a test case?

Yes the POC file is here: https://oss-fuzz.com/download?testcase_id=5124116049166336 I have no idea how the new tests work, if you can point me to the documentation.

your patch seems to work. I'm going to ask @jayaramcs or @APriestman take a look, but if I don't hear from either within a few days (it's Sunday here), I'll accept.

Maybe ask them if removing the code from https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L477 to https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L515 would be an option as well

joachimmetz:
> I think I will need to write the documentation. What would you like it to cover?---

the fuzzer POC will not be "regular" input files, so they don't really seem to "fit" the current "unit test" approach https://github.com/sleuthkit/sleuthkit/blob/develop/test/README.md (although this looks more like a mix of unit and integration tests)

I would need to understand how these fit in the testing framework. So the following information would be useful

* overview of testing framework
* how to write a test, what the different commands like `REQUIRE` are and how they work
* how to integrate a new type of test that fails if a run with ASAN issues warnings

simsong:
It’s using catch2 …. Do you need me to write a new manual for catch2? I thought it was pretty well documented….https://github.com/catchorg/Catch2/blob/devel/docs/assertions.mdThere are also a lot of good tutorials:https://bssw.io/blog_posts/unit-testing-c-with-catchhttps://stackoverflow.com/questions/35757334/catch-unit-testing-framework-require-throws-as----Sent from my phone.On Feb 23, 2025, at 4:56 PM, Joachim Metz ***@***.***> wrote:﻿
  
I think I will need to write the documentation. What would you like it to cover?---

the fuzzer POC will not be "regular" input files, so they don't really seem to "fit" the current "unit test" approach https://github.com/sleuthkit/sleuthkit/blob/develop/test/README.md (although this looks more like a mix of unit and integration tests)
I would need to understand how these fit in the testing framework. So the following information would be useful

overview of testing framework
how to write a test, what the different commands like REQUIRE are and how they work
how to integrate a new type of test that fails if a run with ASAN issues warnings
—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



joachimmetz left a comment (sleuthkit/sleuthkit#3199)

I think I will need to write the documentation. What would you like it to cover?---

the fuzzer POC will not be "regular" input files, so they don't really seem to "fit" the current "unit test" approach https://github.com/sleuthkit/sleuthkit/blob/develop/test/README.md (although this looks more like a mix of unit and integration tests)
I would need to understand how these fit in the testing framework. So the following information would be useful

overview of testing framework
how to write a test, what the different commands like REQUIRE are and how they work
how to integrate a new type of test that fails if a run with ASAN issues warnings

bcarrier:
> possibly, unless Brian recalls what this is about. Maybe removing the "error tolerance" code part and skipping incorrect index values, is the better option here, but I have changed the PR and valgrind no flags OOB read on the fuzzer generated test file.

I _think_ the scenario you are now checking for is when we are at the end of a buffer and trying to scrape up any bits of a remnant entry that got cycled around. 

The 'if' block at [line 477](https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L477) is basically trying to throw out any obviously bogus entries where the data makes absolutely no sense.

My assumption (which could easily be wrong at this point) is that later code would know that we may not have a full data structure and would only copy and parse what is available. That could be failing if you are seeing a crash. 

My guess is that the code in this PR will make it so that we'd see fewer unallocated entries.  At least that's my guess based on the check at [line 579](https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L579) that is the same that you added and that is in 477, but uses that as a trigger to mark the entry as unallocated.

Variable wise, 
* endaddr is the end of the actual buffer
* endaddr_alloc is the end of the used part of the buffer

So, I'd assume any bounds check should use endaddr. 

It looks like the block at [line 536](https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ntfs_dent.cpp#L536) doesn't take endaddr into account. That seems wrong.

Does your fix work if you use 'endaddr' instead of 'endaddr_alloc'

joachimmetz:
> It’s using catch2 

@simsong  that would be good to mention that in the documentation and point to the catch2 documentation. But is uses more than catch2 e.g. what about the scripts? what about the data? What about test build options e.g. having ASAN?

> Does your fix work if you use 'endaddr' instead of 'endaddr_alloc'

Would need to look. IMHO the better approach here is to pass the byte size of the source buffer to tsk_UTF16toUTF8 instead of casting it to a 16-bit pointer instead of accounting for all different (upper level code) boundaries separately. I could add a tsk_UCS2toUTF8 (also given that NTFS uses UCS2 and not UTF16 also see: https://osdfir.blogspot.com/2023/07/whats-in-file-path.html) that takes bytes a input.

joachimmetz:
> @joachimmetz - I downloaded the file and the test case does not crash on my mac...

so a crash (or segfault) only occurs if there is a memory access violation. Modern compilers and various configurations optimize for architectures (e.g. 4 or 8 byte) which not necessary leads to a binary to crash. However under other conditions the same code weakness could cause a binary to crash. I would recommend running with address sanitizer (ASAN) enabled

simsong:
> > @joachimmetz - I downloaded the file and the test case does not crash on my mac...
> 
> so a crash (or segfault) only occurs if there is a memory access violation. Modern compilers and various configurations optimize for architectures (e.g. 4 or 8 byte) which not necessary leads to a binary to crash. However under other conditions the same code weakness could cause a binary to crash. I would recommend running with address sanitizer (ASAN) enabled

I compiled on my back with `./configure --enable-address-sanitizer --enable-undefined-sanitizer` and `Apple clang version 16.0.0 (clang-1600.0.26.6)`.  No problem. Let me find a Linux x86 system somewhere...

simsong:
Okay, I have now compiled with `./configure --enable-address-sanitizer` and `g++ (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0` and run fiwalk with git develop b6e53ee6d Merge pull request #3195 from hillu/avoid-bool.

Running fiwalk on `clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336` I get no address error.

I understand that the ossfuzz system generated an error, but I cannot replicate it with the 139791 byte disk artifact with the SHA-256 of cde563ddb5efb41c31bd34e6fd49b56f82d33eac97e5b8fa417edf7f3ec63ee6.

Given this state of affairs, and given @bcarrier's concern about changing behavior, I'm going to change this PR to a draft until I have a test input that reliably exercises the bug and that this PR reliably fixes.

joachimmetz:
> Running fiwalk on clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336 I get no address error.

I have been able to reproduce the ossfuzz finding with valgrind, the output is in https://github.com/sleuthkit/sleuthkit/issues/3198

joachimmetz:
> I have now compiled with ./configure --enable-address-sanitizer 

What does this do? It only appears to set `CXXFLAGS=-fsanitize=undefined` **where does it link with libasan?**

I can reproduce it with:

```
/configure CFLAGS="-fno-omit-frame-pointer -fsanitize=address -g" LDFLAGS="-fsanitize=address -g -lasan" CXXFLAGS="-fno-omit-frame-pointer -fsanitize=address -g"
make
./tools/fstools/fls clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336 
```

```
./tools/fstools/fls clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336 
=================================================================
==1165270==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x533000018700 at pc 0x7fcb00ed900c bp 0x7ffc449a89a0 sp 0x7ffc449a8998
READ of size 1 at 0x533000018700 thread T0
    #0 0x7fcb00ed900b in tsk_UTF16toUTF8 tsk/base/tsk_unicode.c:163
    #1 0x7fcb0110ff48 in ntfs_dent_copy_short_only tsk/fs/ntfs_dent.cpp:316
    #2 0x7fcb01114098 in ntfs_proc_idxentry tsk/fs/ntfs_dent.cpp:541
    #3 0x7fcb0111c17f in ntfs_dir_open_meta tsk/fs/ntfs_dent.cpp:1250
    #4 0x7fcb0104731b in tsk_fs_dir_open_meta_internal tsk/fs/fs_dir.cpp:308
    #5 0x7fcb010488da in tsk_fs_dir_walk_recursive tsk/fs/fs_dir.cpp:661
    #6 0x7fcb0104a8c9 in tsk_fs_dir_walk_internal tsk/fs/fs_dir.cpp:1001
    #7 0x7fcb0104aa8b in tsk_fs_dir_walk tsk/fs/fs_dir.cpp:1045
    #8 0x7fcb0103ed30 in tsk_fs_fls tsk/fs/fls_lib.cpp:246
    #9 0x404a99 in do_it(TSK_FS_INFO*, unsigned long, int, unsigned long, int, char const*, int) tools/fstools/fls.cpp:406
    #10 0x405068 in main tools/fstools/fls.cpp:436
    #11 0x7fcb00010247 in __libc_start_call_main (/lib64/libc.so.6+0x3247) (BuildId: 515c33a35f41020661fea8ac4eb995e26ccd6b00)
    #12 0x7fcb0001030a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x330a) (BuildId: 515c33a35f41020661fea8ac4eb995e26ccd6b00)
    #13 0x4014b4 in _start (/home/lordyesta/Projects/sleuthkit/tools/fstools/.libs/fls+0x4014b4) (BuildId: fee3a092f34bdcf307073316999678ffc3c171ea)

0x533000018700 is located 0 bytes after 98048-byte region [0x533000000800,0x533000018700)
allocated by thread T0 here:
    #0 0x7fcb016c2250 in calloc (/lib64/libasan.so.8+0xc2250) (BuildId: 0505b45e5a5d9a6c8ecb1d529aaaf13cd21fbe4e)
    #1 0x7fcb00ed26b9 in tsk_malloc tsk/base/mymalloc.c:32
    #2 0x7fcb01119857 in ntfs_dir_open_meta tsk/fs/ntfs_dent.cpp:1046
    #3 0x7fcb0104731b in tsk_fs_dir_open_meta_internal tsk/fs/fs_dir.cpp:308
    #4 0x7fcb010488da in tsk_fs_dir_walk_recursive tsk/fs/fs_dir.cpp:661
    #5 0x7fcb0104a8c9 in tsk_fs_dir_walk_internal tsk/fs/fs_dir.cpp:1001
    #6 0x7fcb0104aa8b in tsk_fs_dir_walk tsk/fs/fs_dir.cpp:1045
    #7 0x7fcb0103ed30 in tsk_fs_fls tsk/fs/fls_lib.cpp:246
    #8 0x404a99 in do_it(TSK_FS_INFO*, unsigned long, int, unsigned long, int, char const*, int) tools/fstools/fls.cpp:406
    #9 0x405068 in main tools/fstools/fls.cpp:436
    #10 0x7fcb00010247 in __libc_start_call_main (/lib64/libc.so.6+0x3247) (BuildId: 515c33a35f41020661fea8ac4eb995e26ccd6b00)
    #11 0x7fcb0001030a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x330a) (BuildId: 515c33a35f41020661fea8ac4eb995e26ccd6b00)
    #12 0x4014b4 in _start (/home/lordyesta/Projects/sleuthkit/tools/fstools/.libs/fls+0x4014b4) (BuildId: fee3a092f34bdcf307073316999678ffc3c171ea)

SUMMARY: AddressSanitizer: heap-buffer-overflow tsk/base/tsk_unicode.c:163 in tsk_UTF16toUTF8
Shadow bytes around the buggy address:
  0x533000018480: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018500: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018580: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018600: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018680: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x533000018700:[fa]fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018780: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018800: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018880: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018900: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018980: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==1165270==ABORTING
```

simsong:
Hm. I thought that --enable-address-sanitizer actually enabled address sanitizer. It does that on bulk_extractor. I will check the autoconf file. Thanks.

simsong:
> @simsong I would recommend creating an issue that `./configure --enable-address-sanitizer` is not WAI

@joachimmetz  — Done, and now it is.

joachimmetz:
will take a look when time permits

simsong:
We still don't have a good resolution on this.

simsong:
However, now that we have clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336 in https://github.com/sleuthkit/sleuthkit_test_data, I'll add running fls on it to the checks.

tsk/fs/ntfs_dent.cpp
```diff
@@ -514,7 +514,12 @@ ntfs_proc_idxentry(NTFS_INFO * a_ntfs, TSK_FS_DIR * a_fs_dir,
                 continue;
             }
         }
-
+	// WARNING: tsk_UTF16toUTF8 used in ntfs_dent_copy_short_only cannot check for
+	// all out of bounds cases.
+        if ((uintptr_t) a_idxe + idxlen > endaddr_alloc) {
+            a_idxe = (ntfs_idxentry *) ((uintptr_t) a_idxe + 4);
+            continue;
```
simsong: Shouldn't we actually fix this:
https://github.com/sleuthkit/sleuthkit/blob/e119503b6fd4472dab1ce8209f770ac1aafa9b19/tsk/base/tsk_unicode.c#L155
and make it `while (source+1 < sourceEnd) {` ?

simsong (changes requested):
@joachimmetz — I'm confused by the message. Shouldn't the correct fix be to fix the while statement in `tsk_UTF16toUTF8`?

tsk/fs/ntfs_dent.cpp
joachimmetz: where do you see source ? are you referring to tsk_UTF16toUTF8 ?

this function blindly assumes sourceEnd is a 16-bit pointer so we cannot detect if the underlying buffer is 1 byte short

simsong: Yes. If we change `source` to `source+1` in `tsk_UTF16toUTF8`, I believe it solves this problem. What do you think?

image

joachimmetz: This is unlikely to going to catch all necessary cases:

* if source is an uneven number of bytes?
* if `source+1` overflows?

An alternative is create a tsk_UTF16toUTF8 equivalent that takes source as bytes and checks for `sourceEnd >= 2 && source < sourceEnd - 2`

simsong: I do not understand. It would certainly handle an uneven number of bytes.  If we make the statement:

```
while (source+1 < sourceEnd) {
}
```
It will handle the dereferences to both *source and *(source+1). Is there any dereference to *(source+2) in the loop?

Presumably source+1 would only overflow if source=0xffffffffffffffff. Then it is not valid. Or do you mean something else?

I'm proposing that we fix `tsk_UTF16toUTF8`. Why would we create a new function that takes source as bytes and checks, rather than fixing this?

Do I misunderstand what this function does?

joachimmetz: no but this issue is not in the function itself but how it is called 

https://github.com/sleuthkit/sleuthkit/issues/3198 is an OOB of 1 byte (presumably because the buffer passed to tsk_UTF16toUTF8 was not a multitude of 2 and just hard casted to 16-bit)

if the function is changed and the check is done in the function this issue will no longer occur, given code can no longer make the call unsafely

simsong: Hi. What does OOB mean?

You wrote: 
> if the function is changed and the check is done in the function this issue will no longer occur, given code can no longer make the call unsafely

That's what I'm proposing. Shoud we do that?

joachimmetz: > Shoud we do that?

that works for me, but unfortunately I lack the time at the moment

the reason I proposed a new function (or wrapper) to ease with the code migration, I'm not sure of all the invocations will migrate as easy

simsong: Ah, OOB means `out of bounds`.

simsong: So my plan is to make the change that I identified, which is correct, and which I believe will address this one as well. Then I will close this PR.

simsong: Actually, I can probably keep this as well.

joachimmetz:
@simsong issue still persists even with fix-3198-v2

```
git pull --squash https://github.com/sleuthkit/sleuthkit fix-3198-v2

/configure CFLAGS="-fno-omit-frame-pointer -fsanitize=address -g" LDFLAGS="-fsanitize=address -g -lasan" CXXFLAGS="-fno-omit-frame-pointer -fsanitize=address -g"
make
./tools/fstools/fls clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336 

=================================================================
==790073==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x533000018700 at pc 0x7efc690dd751 bp 0x7ffffb57c900 sp 0x7ffffb57c8f8
READ of size 1 at 0x533000018700 thread T0
    #0 0x7efc690dd750 in tsk_UTF16toUTF8 tsk/base/tsk_unicode.c:163
    #1 0x7efc6931bbf4 in ntfs_dent_copy_short_only tsk/fs/ntfs_dent.cpp:317
    #2 0x7efc6931fd44 in ntfs_proc_idxentry tsk/fs/ntfs_dent.cpp:542
    #3 0x7efc69327e75 in ntfs_dir_open_meta tsk/fs/ntfs_dent.cpp:1250
    #4 0x7efc6924eb57 in tsk_fs_dir_open_meta_internal tsk/fs/fs_dir.cpp:310
    #5 0x7efc6925020a in tsk_fs_dir_walk_recursive tsk/fs/fs_dir.cpp:677
    #6 0x7efc692527a4 in tsk_fs_dir_walk_internal tsk/fs/fs_dir.cpp:1007
    #7 0x7efc69252966 in tsk_fs_dir_walk tsk/fs/fs_dir.cpp:1051
    #8 0x7efc69246517 in tsk_fs_fls tsk/fs/fls_lib.cpp:246
    #9 0x404a99 in do_it(TSK_FS_INFO*, unsigned long, int, unsigned long, int, char const*, int) tools/fstools/fls.cpp:406
    #10 0x405068 in main tools/fstools/fls.cpp:436
    #11 0x7efc6820f487 in __libc_start_call_main (/lib64/libc.so.6+0x3487) (BuildId: 8d0812d9e0a232d8c2f08880afc3718f055805e6)
    #12 0x7efc6820f54a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x354a) (BuildId: 8d0812d9e0a232d8c2f08880afc3718f055805e6)
    #13 0x4014b4 in _start (/home/user/Projects/sleuthkit/tools/fstools/.libs/fls+0x4014b4) (BuildId: 3fa910fabf80009dcae5528faf6c03a2a6acc83c)

0x533000018700 is located 0 bytes after 98048-byte region [0x533000000800,0x533000018700)
allocated by thread T0 here:
    #0 0x7efc698c2230 in calloc (/lib64/libasan.so.8+0xc2230) (BuildId: 89c230c891879ee538159d2e56f56784c84db409)
    #1 0x7efc690d6ceb in tsk_malloc tsk/base/mymalloc.c:32
    #2 0x7efc6932554d in ntfs_dir_open_meta tsk/fs/ntfs_dent.cpp:1046
    #3 0x7efc6924eb57 in tsk_fs_dir_open_meta_internal tsk/fs/fs_dir.cpp:310
    #4 0x7efc6925020a in tsk_fs_dir_walk_recursive tsk/fs/fs_dir.cpp:677
    #5 0x7efc692527a4 in tsk_fs_dir_walk_internal tsk/fs/fs_dir.cpp:1007
    #6 0x7efc69252966 in tsk_fs_dir_walk tsk/fs/fs_dir.cpp:1051
    #7 0x7efc69246517 in tsk_fs_fls tsk/fs/fls_lib.cpp:246
    #8 0x404a99 in do_it(TSK_FS_INFO*, unsigned long, int, unsigned long, int, char const*, int) tools/fstools/fls.cpp:406
    #9 0x405068 in main tools/fstools/fls.cpp:436
    #10 0x7efc6820f487 in __libc_start_call_main (/lib64/libc.so.6+0x3487) (BuildId: 8d0812d9e0a232d8c2f08880afc3718f055805e6)
    #11 0x7efc6820f54a in __libc_start_main@GLIBC_2.2.5 (/lib64/libc.so.6+0x354a) (BuildId: 8d0812d9e0a232d8c2f08880afc3718f055805e6)
    #12 0x4014b4 in _start (/home/user/Projects/sleuthkit/tools/fstools/.libs/fls+0x4014b4) (BuildId: 3fa910fabf80009dcae5528faf6c03a2a6acc83c)

SUMMARY: AddressSanitizer: heap-buffer-overflow tsk/base/tsk_unicode.c:163 in tsk_UTF16toUTF8
Shadow bytes around the buggy address:
  0x533000018480: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018500: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018580: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018600: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x533000018680: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x533000018700:[fa]fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018780: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018800: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018880: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018900: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x533000018980: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==790073==ABORTING
```

joachimmetz:
@simsong PTAL at changes

joachimmetz:
```
/configure CFLAGS="-fno-omit-frame-pointer -fsanitize=address -g" LDFLAGS="-fsanitize=address -g -lasan" CXXFLAGS="-fno-omit-frame-pointer -fsanitize=address -g"
make
./tools/fstools/fls /tmp/clusterfuzz-testcase-minimized-sleuthkit_fls_ntfs_fuzzer-5124116049166336 
r/- 8:	$BaᝤClus
r/r 6:	$Åitmap
r/- 7:	$Boot
d/V 11:	$
r/- 0:	$MFT
r/- 1:	ā
r/- 0:	
V/V 11:	$OrphanFiles
```

simsong (approved):

simsong:
Looks good. We are on our way to fixing this longstanding problem! Thanks.

tsk/base/tsk_unicode.c
```diff
@@ -252,6 +252,38 @@ tsk_UTF16toUTF8(TSK_ENDIAN_ENUM endian, const UTF16 ** sourceStart,
...
+	int retVal = TSKconversionOK;
+
+	if (source_len > 0 && source_len % 2 != 0) {
+		// source_len must be even, since UTF-16 uses 2 bytes per character
+		source_len -= 1;
+	}
+	if (source_len < (number_of_characters * 2)) {
+		number_of_characters = source_len / 2;
+	}
+	retVal = tsk_UTF16toUTF8(endian, (const UTF16 **) source,
```
Taha-Ebrahim: I was making a unit test to try and replicate the malformed buffer bug that we were finding with the address sanitizer and `ntfs_dent_copy_short_only` and this implementation still fails the test and seems to contain an ASAN issue. 

Since `source` is initially a `const UTF16*`, after changing `source` to `&source` it passes the test.

simsong: @joachimmetz could you take a look at this? Thanks.

joachimmetz: let me take a look later, I could have forgotten a & in my time constraints, so good catch if that's the case

joachimmetz: Yeah there is an error in the parameter passing to `tsk_UTF16toUTF8` have a pending fix.

Based on the POC the crux of the issue is in `ntfs_proc_idxentry` `a_idxe` exceeding `endaddr_alloc`

a_idxe: 0x5330000186a4 (at the time of the ASAN flag)
endaddr: 0x533000018700
endaddr_alloc: 0x533000018148

joachimmetz: Note that IMHO the code would be more clear when another variable is used instead of incrementing the function argument `a_idxe`

Taha-Ebrahim: I've re-run the tests on my end with the new fixes it still seems to fail the test. I have made a PR #3367 with the failing test and have also copied it here. 

```c++
TEST_CASE("tsk_safeUTF16toUTF8 prevents out-of-bounds read", "[unicode][safe]") {
    // 1. Set up a small, fixed-size source buffer.
    // The buffer is 4 bytes and can hold 2 UTF-16 characters.
    const size_t source_buffer_len = 4;
    uint8_t source_buffer[source_buffer_len];
    // Fill the buffer with valid UTF-16 characters: "AB"
    source_buffer[0] = 'A'; source_buffer[1] = 0;
    source_buffer[2] = 'B'; source_buffer[3] = 0;
    
    // 2. Set up the target buffer.
    const size_t target_buffer_len = 10;
    UTF8 target_buffer[target_buffer_len] = {0};
    UTF8 *target_ptr = target_buffer;
    UTF8 *target_end_ptr = target_buffer + target_buffer_len;

    // 3. Deliberately provide a malformed number_of_characters.
    // We claim there are 3 characters, but the buffer only holds 2.
    const size_t malformed_num_chars = 3;

    // 4. Call the fixed function with the malformed input.
    // The function is expected to correct the input before any unsafe access occurs.
    SECTION("Function call with malformed input") {
        REQUIRE_NOTHROW(tsk_safeUTF16toUTF8(TSK_LIT_ENDIAN, source_buffer, source_buffer_len,
                                            malformed_num_chars, &target_ptr, target_end_ptr,
                                            TSKlenientConversion));

        // 5. Verify the function's behavior.
        // The function should return TSKconversionOK as it safely converted the available data.
        REQUIRE(tsk_safeUTF16toUTF8(TSK_LIT_ENDIAN, source_buffer, source_buffer_len,
                                    malformed_num_chars, &target_ptr, target_end_ptr,
                                    TSKlenientConversion) == TSKconversionOK);
    }
}
```

simsong:
Please merge #3367 into this. Thanks!

joachimmetz:
@simsong friendly nudge/ping, are we pursuing this direction or not? I added changes to this PR address similar OOB read issues in ntfs_proc_attrseq and fixed an issue in the string end pointer calculation

tsk/fs/ntfs.cpp
```diff
@@ -2437,23 +2438,31 @@ ntfs_proc_attrseq(NTFS_INFO * ntfs,
                 }
                 fs_name->next = NULL;
             }
+	    /* This check does not make sense
```
joachimmetz, line 2444: @simsong please take a look if you agree

simsong:
Coming back to this. SOrry for the delay.

tsk/fs/ntfs.cpp
simsong: Okay. I'm fine with having this code commented out for now.

simsong (approved):

bcarrier:
I plugged this PR into Claude to reload all of this into my memory. 

It summarized the basic issue as callers to 'tsk_UTF16toUTF8()' were passing in end addresses that were not correct. I.e. it was a caller problem and not a tsk_UTF16toUTF8() problem. 

Other fuzzing PRs recently have addressed this in other ways: 
Recently fixed correctly:
- ntfs_dent_copy_short_only — your own commit 89b31cd ("Add bounds checks to fix #3436")
- ntfs_proc_attrseq $FNAME path — offsetof check at line 2435

The remaining gaps
- ntfs_istat: 4839
- usn_journal.c:55

Other callers were already doing bounds checks. 

I haven't had a chance to run this code against the original image yet though. 

#3496 adds those checks to ntfs_istat and usn_journal

joachimmetz:
The problem here is that tsk_UTF16toUTF8 allows for an OOB read, the idea behind having a tsk_safeUTF16toUTF8 that checks the bound is that you don't have to do this everywhere you use tsk_UTF16toUTF8

PR #3221, opened 2025-03-10
Reverts the 'ant' requirement from #3217

bcarrier:
Ant should not be a requirement unless the user wants to build Java. Someone should edit what is now there. I only commented out the error.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -219,10 +219,11 @@ AS_IF([test "x$enable_java" != "xno"], [
 ]) dnl test enable_java
 
 dnl Test is ant is available since we now use it for the build system
+dnl BC - TODO - Add this to the enable Java section and uncomment below
 AC_PATH_PROG([ANT], [ant], [no])
-if test "x$ANT" = "xno"; then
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -219,10 +219,11 @@ AS_IF([test "x$enable_java" != "xno"], [
...
 dnl Test is ant is available since we now use it for the build system
+dnl BC - TODO - Add this to the enable Java section and uncomment below
 AC_PATH_PROG([ANT], [ant], [no])
-if test "x$ANT" = "xno"; then
-   AC_MSG_ERROR([Apache Ant is required but was not found. Install Ant and try again.])
-fi
+dnl if test "x$ANT" = "xno"; then
+dnl    AC_MSG_ERROR([Apache Ant is required but was not found. Install Ant and try again.])
+dnl fi
 AC_SUBST(ANT)
```
simsong, line 227: We can remove this, but then one of the `make clean` targets is going to fail, because many of the things made by `ant` can only be cleaned by ant. So we will need to conditionalize the `make clean` as well.

simsong: And this should be moved into the Java section then, which I will do.

PR #3225, opened 2025-03-14
Add btrfs zstd flag

michael-ashburn:
At the moment `fsstat` can't detect btrfs with zstd compression:

```
btrfs_open: Unsupported superblock incompat_flags:
COMPRESS_LZOv2
...
Cannot determine file system type
```

See https://github.com/torvalds/linux/blob/695caca9345a160ecd9645abab8e70cfe849e9ff/fs/btrfs/sysfs.c#L285C1-L285C25 and https://github.com/torvalds/linux/blob/695caca9345a160ecd9645abab8e70cfe849e9ff/fs/btrfs/compression.h#L119.

### DIFF
--- a/tsk/fs/btrfs.cpp
+++ b/tsk/fs/btrfs.cpp
@@ -4253,7 +4253,7 @@ btrfs_fsstat_print_incompat_flags(FILE * a_file, uint64_t a_flags)
         "DEFAULT_SUBVOL",
         "MIXED_GROUPS",
         "COMPRESS_LZO",
-        "COMPRESS_LZOv2",
+        "COMPRESS_ZSTD",
         "BIG_METADATA",
         "EXTENDED_IREF",
         "RAID56",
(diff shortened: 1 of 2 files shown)


simsong:
Hi @michael-ashburn  — We are super-excited to receive this PR. However, we request that you also give us a disk image that demonstrates the error. Can you do that?

Also, does our BTRFS implementation simply not detect these file systems, or can it also not read them?

Regards,

michael-ashburn:
> we request that you also give us a disk image that demonstrates the error

[btrfs_zstd.zip](https://github.com/user-attachments/files/19250237/btrfs_zstd.zip)

```sh
dd if=/dev/zero of=btrfs_zstd bs=1M count=109
mkfs.btrfs btrfs_zstd
mount -o compress=zstd btrfs_zstd /mnt/tmp
touch /mnt/tmp/test
umount /mnt/tmp
```

```cmd
fsstat btrfs_zstd
```

PR #3228, opened 2025-03-15
Enable java on debian GitHub actions .

simsong:
First step to making a complete deb that includes java support:

1. Enables Java for GitHub actions for MacOS and Ubuntu.
2. Adds various `#pragma` statements to suppress warnings on some code that now gets compiled. 
3. Adds dependencies to `Makefile.am` so that it can build with both `-j1` and `-j`


simsong:
This failed because of the `make -j4`. But I am sure that `make -j1` would have succeded. Time to debug this dependency problem.

PR #3243, opened 2025-03-31
rewrite of ossfuzz implementation

simsong:


simsong:
Let's see if this works tonight.

PR #3246, opened 2025-03-31
Validate AFF and AFF4 in CI/CD system.

simsong:
add support for afflib and aff4 to ubuntu build so this actually gets tested by ci/cd process


simsong:
afflib is not getting installed:
image

PR #3247, opened 2025-04-01
fix distcheck

simsong:


simsong:
Yay! `make distcheck` finally works.

PR #3256, opened 2025-04-03
Prevent handle leaks and double-frees

uckelman-sf:
This PR wraps almost all local uses of TSK handles in `std::unique_ptr`, to ensure that they are freed exactly once along every code path.


simsong:
Hi @uckelman-sf  — thank you so much for your attention to this! Please let me know if there is anything I can do. Please feel free to update the CI/CD system if you think we can get more info about if it.

simsong:
Well, this all works except for the appveyor, which I need to fix, so I'm going to merge this in.

uckelman:
You're probably missing `SLEUTKHIT_TEST_DATA_DIR` in the appveyor build as well.

simsong:
> You're probably missing `SLEUTKHIT_TEST_DATA_DIR` in the appveyor build as well.

Indeed, but I think that I added it. See: https://github.com/sleuthkit/sleuthkit/pull/3258

I'm now working on the CI/CD thing for the most recent SleuthKit release from CyberTriage which demonstrates that the CT fork also has the problem that you just fixed...

PR #3262, opened 2025-04-08
merged in changes from Eran-YT:speed-up-tsk_fs_path2inum

simsong:
This takes over https://github.com/sleuthkit/sleuthkit/pull/2517
Closes #2517 

Adds a new function `tsk_fs_dir_get2()` which allows specifying if attributes should be loaded on get directory entries. `tsk_fs_dir_get()` modified to call `tsk_fs_dir_get2()` with `load_attributes` set to false.

From #2517:

> Currently tsk_fs_path2inum uses tsk_fs_dir_get which loads the attributes for the directory it gets, which tsk_fs_path2inum doesn't actually need in most cases, and this is very expensive (They are read from disk).
> This PR changes this behavior to only load the attributes if needed (only a very specific case involving NTFS), thus speeding up the function.
> By my benchmarks the speed up is about by a factor of >50.
> This also reduces the disk I/O.

### DIFF
--- a/tsk/fs/fs_dir.cpp
+++ b/tsk/fs/fs_dir.cpp
@@ -431,8 +431,21 @@ tsk_fs_dir_getsize(const TSK_FS_DIR * a_fs_dir)
  * @param a_idx Index of file in directory to open (0-based)
  * @returns NULL on error
  */
+TSK_FS_FILE*
+tsk_fs_dir_get(const TSK_FS_DIR* a_fs_dir, size_t a_idx)
+{
+    return tsk_fs_dir_get2(a_fs_dir, a_idx, TRUE);
+}
+
+/** \ingroup fslib
+* Return a specific file or subdirectory from an open directory.
+ * @param a_fs_dir Directory to analyze
+ * @param a_idx Index of file in directory to open (0-based)
+ * @param load_attributes Boolean indicating if the file attributes should be loaded
+ * @returns NULL on error
+ */
 TSK_FS_FILE *
-tsk_fs_dir_get(const TSK_FS_DIR * a_fs_dir, size_t a_idx)
+tsk_fs_dir_get2(const TSK_FS_DIR * a_fs_dir, size_t a_idx, size_t load_attributes)
 {
     TSK_FS_NAME *fs_name;
     TSK_FS_FILE *fs_file;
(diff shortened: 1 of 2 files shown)


tsk/fs/fs_dir.cpp
```diff
@@ -431,8 +431,21 @@ tsk_fs_dir_getsize(const TSK_FS_DIR * a_fs_dir)
...
+/** \ingroup fslib
+* Return a specific file or subdirectory from an open directory.
+ * @param a_fs_dir Directory to analyze
+ * @param a_idx Index of file in directory to open (0-based)
+ * @param load_attributes Boolean indicating if the file attributes should be loaded
+ * @returns NULL on error
+ */
 TSK_FS_FILE *
-tsk_fs_dir_get(const TSK_FS_DIR * a_fs_dir, size_t a_idx)
+tsk_fs_dir_get2(const TSK_FS_DIR * a_fs_dir, size_t a_idx, size_t load_attributes)
```
uckelman, line 448: Why is `load_attributes` a `size_t` instead of a `bool`?

simsong:
Hm. Because I should have fixed it and didn’t.----Sent from my phone.On Apr 8, 2025, at 7:42 PM, Joel Uckelman ***@***.***> wrote:﻿
@uckelman commented on this pull request.



In tsk/fs/fs_dir.cpp:
>  TSK_FS_FILE *
-tsk_fs_dir_get(const TSK_FS_DIR * a_fs_dir, size_t a_idx)
+tsk_fs_dir_get2(const TSK_FS_DIR * a_fs_dir, size_t a_idx, size_t load_attributes)

Why is load_attributes a size_t instead of a bool?

simsong:
> Hm. Because I should have fixed it and didn’t.----Sent from my phone.On Apr 8, 2025, at 7:42 PM, Joel Uckelman ***@***.***> wrote:﻿ @uckelman commented on this pull request. In tsk/fs/fs_dir.cpp:
> TSK_FS_FILE *
> -tsk_fs_dir_get(const TSK_FS_DIR * a_fs_dir, size_t a_idx) +tsk_fs_dir_get2(const TSK_FS_DIR * a_fs_dir, size_t a_idx, size_t load_attributes) Why is load_attributes a size_t instead of a bool? —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you modified the open/close state.Message ID: ***@***.***>

Noted. PR submitted https://github.com/sleuthkit/sleuthkit/pull/3271

tsk/fs/tsk_fs.h
```diff
@@ -676,6 +676,7 @@ extern "C" {
         void *a_ptr);
     extern size_t tsk_fs_dir_getsize(const TSK_FS_DIR *);
     extern TSK_FS_FILE *tsk_fs_dir_get(const TSK_FS_DIR *, size_t);
+    extern TSK_FS_FILE *tsk_fs_dir_get2(const TSK_FS_DIR *, size_t, size_t load_attributes);
```
bcarrier, line 679: This is going into the public API. I don't think it needs a "2" in the name.

simsong: @bcarrier  — 
`tsk_fs_dir_get()` is the legacy API. We can't change it. 
`tsk_fs_dir_get2()` is the new API that takes 2 arguments. What would you rather call it?

PR #3263, opened 2025-04-09
Paul1278 impl/residential layout

simsong:
Closes #2871


simsong:
From @bcarrier:

> Any existing NTFS test image should probably have a resident file. They are small files that are a few hundred bytes and don't get their own block allocated to them. They use space in the fixed-size MFT entry (1024-bytes).
> 
> Some context on this proposed change though. I considered this when we first built it out, but decided it against it because technically those blocks are allocated to $MFT (The master file table, where the MFT entries reside). 
> 
> I assume that with this change, TSK will report that two allocated files are associated with blocks for resident files:
> * The $MFT
> * The resident file that is using space in $MFT
> 
> According to the file system, only $MFT is using that block.
> 
> This may break code that assumes only one allocated file will be using a block.

simsong:
I asked @copilot to review this, but it would be great if @APriestman or @bcarrier could also give it a look.

PR #3264, opened 2025-04-10
fix `make distcheck` and `make check`

simsong:
- test programs now `exit 77` if their data is not present.
- data can reside in `${src_dir}/test/data` (small things) or in `$SLEUTHKIT_TEST_DATA_DIR` (big things)


simsong:
`make distcheck` finally works. @bcarrier , @jayaramcs , @jonstewart and @uckelman-sf may all enjoy seeing just how many changes were involved in getting this to work (and note: it only works because we disable building the Java stuff).

The key problem is that `make distcheck` does an out-of-source tree build, and it makes the source tree read-only to ensure correctness. The `Makefile` is built in a directory called `_build/sub` and both `srcdir` and `top_srcdir` are set to `../..`. (`srcdir` and `top_srcdir` are the same when there is a single `Makefile`, but they are different when there are multiple makefiles.)

Another challenge was in ensuring that all of the Java sources are included when a `make dist` is done. Previously they were not included, as `Makefile.am` referenced then with a `*` glob, but `automake` and `make` do not do glob expansion. This is corrected with a `dist-hook` than manually copies into the distribution directory the relevant `*.java` and other files. Note that this copy also needs to work when doing an out-of-source build.

The real problem getting the out-of-source build to work is that the Java files are built with `ant` and `ant` does not support this well. I would need to modify the automatically-generated `build.xml` files to understand an out-of-source build, and this is more work than I want to do just now. I've added https://github.com/sleuthkit/sleuthkit/issues/3269 as an issue to track this.

PR #3271, opened 2025-04-15
load_attributes should be a `bool`, not a `size_t`

simsong:


simsong:
It's a bool because it's a pure C header, not a C++ header.

PR #3286, opened 2025-05-19
fixed typo "teh" to "the"

Taha-Ebrahim:


tools/fstools/fls.cpp
```diff
@@ -65,7 +65,7 @@ usage()
     tsk_fprintf(stderr, "\t-p: Display full path for each file\n");
     tsk_fprintf(stderr, "\t-r: Recurse on directory entries\n");
     tsk_fprintf(stderr, "\t-u: Display undeleted entries only\n");
-    tsk_fprintf(stderr, "\t-v: Verbose output to stderr\n");
+    tsk_fprintf(stderr, "\t-v: verbose output to stderr\n");
```
simsong, line 68: If you are going to fix to be consistent with mmls (see below) , please fix them all to be consistency. Otherwise this looks weird.
```
 % mmls -h                                                                                                                       (main)mitsearch
usage: mmls [-i imgtype] [-b dev_sector_size] [-o imgoffset] [-BcrvVh] [-aAmM] [-t vstype] image [images]
	-t vstype: The type of volume system (use '-t list' for list of supported types)
	-i imgtype: The format of the image file (use '-i list' for list supported types)
	-b dev_sector_size: The size (in bytes) of the device sectors
	-o imgoffset: Offset to the start of the volume that contains the partition system (in sectors)
	-B: print the rounded length in bytes
	-r: recurse and look for other partition tables in partitions (DOS Only)
	-c: print CSV output
	-v: verbose output
	-V: print the version
	-h: help. print this message
Unless any of these are specified, all volume types are shown
	-a: Show allocated volumes
	-A: Show unallocated volumes
	-m: Show metadata volumes
	-M: Hide metadata volumes
```

simsong (changes requested):
If you are going to change the capitalization of one, please fix them all for consistency. However, I do not think that this meaningfully improves the function, so I recommend not making this change.

simsong:
I'm closing this PR because this change doesn't meaningfully improve the code base, and it changes the output of the command for no apparent reason.

tools/fstools/fls.cpp
Taha-Ebrahim: It caused an issue with one of the tests (see [0af442d]) so I reverted the changes... None of them should be changed anymore.

simsong: > It caused an issue with one of the tests (see [[0af442d](https://github.com/sleuthkit/sleuthkit/commit/0af442d665f46bd425dd8b42c1643f970b06de43)]) so I reverted the changes... None of them should be changed anymore.

Right. This is why I have requested that you work on your own fork and making sure it passes all tests before sending it to this fork.

PR #3290, opened 2025-05-27
Added tests for img_types.c

Taha-Ebrahim:
Made the following changes: 

1) created tsk_tempfile.[cpp, h] files to capture output for testing purposes.
2) created img_types.h file for testing purposes.
3) made changes to improve null input checking for img_types.c
4) created new test_img_types.cpp file for tests themselves.
5) added files to sources of Makefile.am so that tests are run. 
6) made changes to testing readme for clarity.


test/tools/tsk_tempfile.cpp
```diff
@@ -0,0 +1,43 @@
+#include "tsk_tempfile.h"
```
simsong, line 8: This file needs a comment to explain why it is necessary. Also, I thought it was only needed under mingw?

simsong:
Thank you for the submission. I am conducting my review now and I have changed this into a draft request.

simsong:
@Taha-Ebrahim  — Coverage of the develop branch is nowhere near 60%, so please modify the yaml file for GitHub actions so that this PR does not fail because develop is below 60%.

image

tsk/img/img_types.h
```diff
@@ -0,0 +1,35 @@
...
+const char *tsk_img_type_todesc(TSK_IMG_TYPE_ENUM type);
+
+// Return a bitmask of supported image types
+TSK_IMG_TYPE_ENUM tsk_img_type_supported(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif // IMG_TYPES_H
```
simsong: Our TSK convention is that files with names like tsk/img/img_types.h are part of the public API, wheres a file with the name `tsk/img/img_types_i.h` would not be part of the public API. I don't want to make this part of the public API, so please rename to `tsk/img/img_types_i.h`.

simsong: (And if you want, you can start on a README_DEVELOPER.md for internal developer documentation which includes everything you learn/discover during this process.

tsk/img/img_types.c
```diff
@@ -94,6 +98,10 @@ tsk_img_type_toid_utf8(const char *str)
 TSK_IMG_TYPE_ENUM
 tsk_img_type_toid(const TSK_TCHAR * str)
 {
+    if (str == NULL) {
+        return TSK_IMG_TYPE_UNSUPP;
+    }
+
```
simsong: Please use `nullptr`. 

I think that it should throw an exception or return a different error. This is calling error, not an unsupported filetype. My preference is to throw an exception, but we do not have exceptions widely implemented in TSK. However, this is an okay time to start.

test/tsk/img/test_img_types.cpp
```diff
@@ -0,0 +1,144 @@
...
+  REQUIRE(output.find("Single or split raw file") != std::string::npos);
+
+#if HAVE_LIBEWF
+  SECTION("Known format: ewf (if enabled)") {
+      REQUIRE(output.find("ewf") != std::string::npos);
+  }
+#endif
+}
+
+// MWE test case that fails on MinGW systems because of std::tmpfile.
```
simsong: Expand MWE; people may not know what it means.
Is this a MWE test case? I think that it's just a test case. 
I'm okay if you want to use tempfile, but let's bring this back and have it test tsk_tempfile() which should just passthrough to std::tempfile() on all platforms except MinGW.

simsong (changes requested):
Please address comments. Thanks!

simsong:
> [Taha-Ebrahim](/Taha-Ebrahim)

Oh, this wasn't your doing it, was https://github.com/sleuthkit/sleuthkit/commit/e51d8a8a13f9725403a6264f625410295d00a8a5. Fixed in https://github.com/sleuthkit/sleuthkit/pull/3291

tsk/img/img_types.c
Taha-Ebrahim: img_types is written in C so there are a couple of issues because of language incompatibility:
* There is no ``nullptr`` keyword. 
* Exceptions aren't natively supported.

 How should I proceed?

simsong:
I guess you are right. It needs to be NULL and not nullptr. Apologies!

test/tools/tsk_tempfile.cpp
```diff
@@ -0,0 +1,43 @@
...
+        if (!env) env = std::getenv("TMP");
+        temp_dir = env ? env : ".";
+    
+        std::string filename = "tsk_tempfile_" + std::to_string(std::time(nullptr)) +
+                               "_" + std::to_string(GetTickCount64()) + ".txt";
+        std::string full_path = temp_dir + PATH_SEP + filename;
+    
+        FILE* file = std::fopen(full_path.c_str(), "w+");
+        return file;
+    #else
```
simsong: By  convention, C-preprocessor `#` statements are flush left. Please make lines 39 and lines 42 flush left.

test/tsk/img/test_img_types.cpp
```diff
@@ -0,0 +1,139 @@
...
+  TSK_IMG_TYPE_ENUM raw = TSK_IMG_TYPE_RAW;
+  REQUIRE(!strcmp(tsk_img_type_toname(raw), "raw"));
+  TSK_IMG_TYPE_ENUM unsupp = TSK_IMG_TYPE_UNSUPP;
+  REQUIRE(tsk_img_type_toname(unsupp) == nullptr);
+  #if HAVE_LIBEWF
+  SECTION("Known format: ewf (if enabled)") {
+    TSK_IMG_TYPE_ENUM ewf = TSK_IMG_TYPE_EWF_EWF;
+    REQUIRE(!strcmp(tsk_img_type_toname(ewf), "ewf"));
+  }
+  #endif
```
simsong: Same here. All `#ifdef` and  `#endif` statements should be flush-left.

tsk/img/img_types.c
simsong: Well, `TSK_IMG_TYPE_UNSUPP` is wrong, because `NULL` is not a type. See if there is a return code for null pointer inappropriately provided.

simsong (changes requested):
Much better! Let's make it perfect...
You are also welcome to start on other tests while we finish off this one.

tsk/img/img_types.c
Taha-Ebrahim: I couldn't find a return type for a null pointer specifically. I was thinking that I could use ``TSK_IMG_TYPE_DETECT = 0x0000``. Otherwise I could ``exit(1)``.

simsong: Hm. I'm now aware that you are adding code to `img_types.c` rather than adding a test case. You are making the code more robust, which is good, but this is not strictly improving the code coverage.  How about just making this an `assert` statement? Just assert it is not NULL. 

You may find this article useful: https://cacm.acm.org/research/mars-code/

simsong (commented):
Looks better!

simsong:
AppVeyor failed because the Linux VM didn't load its security patches.

simsong (approved):

PR #3293, opened 2025-05-27
Add Codecov workflow

singhpratik5:


simsong (changes requested):
As we discussed:
1. Please rename this file with consistency with the other actions.
2. Please remove the codecov from the build action.
3. Note that this action closes #3293

PR #3309, opened 2025-06-18
Do not use img after move

uckelman-sf:


tools/fstools/fls.cpp
```diff
@@ -368,7 +368,7 @@ open_handles(const Options& opts, const TSK_TCHAR* const* argv, size_t argc) {
 
         img_parent = std::move(img);
 
-        TSK_OFF_T offset = imgaddr * img->sector_size;
+        TSK_OFF_T offset = imgaddr * img_parent->sector_size;
```
simsong, line 371: great going.

simsong: I wonder why our static analysis tools didn't pick this up. Should we add others? This seems like a pretty simple pattern to detect.

simsong (approved):

PR #3316, opened 2025-06-30
Updated CLI test runner proposal

Taha-Ebrahim:
Proposal for a new C++ test runner for TSK CLI tools, tsk_cli_runner, that reads a pipe-delimited test definition file (cli_tests.txt) to validate CLI tool behavior end-to-end.


simsong (approved):

simsong:
@uckelman-sf - Any thoughts on this?

uckelman-sf:
Why do this? What problem does it solve?

simsong:
> Why do this? What problem does it solve?

It's proven to be quite difficult to create new tests, with many files requiring modification, and some quoting problems. 
I would like to simplify creation of new tests and improve reporting.

PR #3320, opened 2025-07-02
AppVeyor Test Data Directory Fix

Taha-Ebrahim:
Ensures make unpack is run after unzipping test data and sets `SLEUTHKIT_TEST_DATA_DIR` before running `make check`. Fixes missing test input issues in CI.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,9 @@
 version: 4.6.0.{build}
 
+# Run on all pull requests to any branch
+# Skip branch builds when there are PRs to prevent duplicates
+skip_branch_with_pr: true
+
 environment:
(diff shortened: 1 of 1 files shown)


simsong:
> Ensures make unpack is run after unzipping test data and sets `SLEUTHKIT_TEST_DATA_DIR` before running `make check`. Fixes missing test input issues in CI.

I'm looking forward to seeing how this works. I'm unclear why we are getting those C++ unit test errors. Any thoughts?

Taha-Ebrahim:
image

It doesn't seem to be skipping any of the tests on the macOS build anymore!

simsong:
Weird, but good!

PR #3325, opened 2025-07-03
Fix develop-4.14 to compile under windows & mac; Removes many warnings;

simsong:
make github action run on develop-4.14 branch


simsong:
It's odd that only three checks had to run. I wonder why.

PR #3347, opened 2025-07-15
add better skipping functionality

Taha-Ebrahim:
Currently, tests would fail when the `SLEUTHKIT_TEST_DATA_DIR` is not set but used in CLI testing. Tests are now skipped more gracefully if it is not set.


simsong:
This is great. 

@Taha-Ebrahim — Also remember — If any test is skipped, the program should do an `exit 77` and not `exit 0`. I don't have time to check the code to see if that happens...

PR #3351, opened 2025-07-24
Fix timezone test

simsong:
* set TZ in the test program and check for correct behavior.
* remove test/tools/test_cli_runner from repo (it's a script that is generated)


simsong:
Well, this took a lot of effort, but I finally got some decent tests and demonstrate the different behavior of `asctime()` on `mingw`.  @Taha-Ebrahim and @singhpratik5 — both of you may be interested.

singhpratik5:
@simsong Sure, It's a lot of effort.

Taha-Ebrahim:
Will take a look!

PR #3357, opened 2025-07-29
noise reduction for tsk_cli_runner

Taha-Ebrahim:
tsk_cli_runner log was getting quite cluttered. fix to keep log more neat.

### DIFF
--- a/test/tools/test_utils.cpp
+++ b/test/tools/test_utils.cpp
@@ -63,9 +63,11 @@ std::string read_file(FILE* file) {
 bool compare_files(FILE* expected, FILE* actual) {
     std::string expected_content = read_file(expected);
     std::string actual_content = read_file(actual);
-
-    std::cout << "Expected File: " << expected_content << '\n';
-    std::cout << "Output: " << actual_content << '\n';
+    
+    if (expected_content != actual_content) {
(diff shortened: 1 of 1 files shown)


test/tools/test_utils.cpp
```diff
@@ -63,9 +63,11 @@ std::string read_file(FILE* file) {
...
     std::string expected_content = read_file(expected);
     std::string actual_content = read_file(actual);
-
-    std::cout << "Expected File: " << expected_content << '\n';
-    std::cout << "Output: " << actual_content << '\n';
+    
+    if (expected_content != actual_content) {
+        std::cout << "Expected File: " << expected_content << '\n';
+        std::cout << "Output: " << actual_content << '\n';
+    }
```
simsong, line 71: This is a good start, but it would be hard to read. I recommend including the byte length, so the output looks like this:

```
== OUTPUT DOES NOT MATCH EXPECTED OUTPUT ==
Expected output NNN bytes in NN lines
Actual output: YYY bytes in YY lines

Expected output:
<output goes here>

Actual output:
<output goes here>
=======
```

(And you could add lines numbers or a diff, but probably more than is needed. We implement this sort of functionality in many places and it is nice to have it centralized.)

simsong (approved):

PR #3360, opened 2025-08-02
added test/tsk/fs/test_fs_io.cpp

singhpratik5:
This PR validates that `tsk_fs_read_decrypt` properly decrypts a block of encrypted data.
- [ ] It starts with a block of data from an encrypted file system for which the decryption is known.
- [ ] It configures the decryption apparatus. 
- [ ] It calls the decryption functions.
- [ ] It verifies that the decryption took place properly.


simsong:
@singhpratik5 — I'm going to close this PR right now because I don't think that you have the necessary data to test the decryption functions. I recommend that you create an end-to-end test first that has an encrypted disk image, sets up the key, and tries to read an encrypted file. Please mail @Taha-Ebrahim, as he was looking at that.

PR #3364, opened 2025-08-07
Clean up log output for readability

Taha-Ebrahim:
Made a number of changes to clean up the log files:

* Introduced global variables `g_tsk_printf_fd` and `g_tsk_stderr_fd` in tsk_printf.c to control the destination of all output from the tsk_printf functions. 
* Changed output in `dcalc_lib.cpp` and `dls_lib.cpp` so that data is written to `g_tsk_printf_fd` if set, instead of directly to stdout (in `fwrite`).
* Standardized the use of tsk_printf and `tsk_fprintf` for all diagnostic output, removing direct use of standard output functions.
* Cleaned up log messages by removing extraneous newlines.
* Made changes to `tsk_base.h` to reflect the changes made in `tsk_printf.c` 

Log files still need some work to get them fully clean.


simsong (changes requested):
When can this temporary comment out in test_cli_runner be removed?

test/tools/test_cli_runner.cpp
```diff
@@ -71,6 +71,7 @@ TEST_CASE("readfile") {
 }
 
 /// Test for print_diff when lines differ
+/* TEMPORARY REMOVAL WHILE UPDATING TEST_UTILS.CPP
```
simsong: when can this be removed?

simsong (approved):

test/tools/test_cli_runner.cpp
simsong: @Taha-Ebrahim - did you see this?

PR #3381, opened 2025-09-11
Support ext4 journal dentry checksums

### LINKED ISSUE #1475: ext4 journal descriptor block parsing error.
You should revise ext2fs_journ_dentry struct from
typedef struct {
	uint8_t fs_blk[4];
	uint8_t flag[4]
}ext2fs_journ_dentry;
to
typedef struct {
	uint8_t fs_blk[4];
	uint8_t flag[4];
	uint8_t fs_blk_hi[4];
	uint8_t checksum[4];
}ext2fs_journ_dentry;
In generall case, your structure doesn't match on ext4 journal system.

bjangelo:
Support JBD2_FEATURE_INCOMPAT_CSUM_V2 and
JBD2_FEATURE_INCOMPAT_CSUM_V3, which result in variable size journal descriptor entries.

https://www.kernel.org/doc/html/latest/filesystems/ext4/journal.html#descriptor-block

Closes #1475

### DIFF
--- a/tsk/fs/ext2fs_journal.cpp
+++ b/tsk/fs/ext2fs_journal.cpp
@@ -24,11 +24,27 @@
 
 
 /* Everything in the journal is in big endian */
+#define big_tsk_getu16(x)	\
+	(uint16_t)((((uint8_t *)x)[1] <<  0) + \
+	(((uint8_t *)x)[0] <<  8) )
+
 #define big_tsk_getu32(x)	\
 	(uint32_t)((((uint8_t *)x)[3] <<  0) + \
 	(((uint8_t *)x)[2] <<  8) + \
 	(((uint8_t *)x)[1] << 16) + \
 	(((uint8_t *)x)[0] << 24) )
+
+#define big_tsk_getu64(x, y)	\
+    ((uint64_t) \
+    ((uint64_t)((uint8_t *)(y))[3] <<  0) + \
+    ((uint64_t)((uint8_t *)(y))[2] <<  8) + \
+    ((uint64_t)((uint8_t *)(y))[1] << 16) + \
+    ((uint64_t)((uint8_t *)(y))[0] << 24) + \
+    ((uint64_t)((uint8_t *)(x))[3] << 32) + \
+    ((uint64_t)((uint8_t *)(x))[2] << 40) + \
+    ((uint64_t)((uint8_t *)(x))[1] << 48) + \
+    ((uint64_t)((uint8_t *)(x))[0] << 56))\
+
 
 
 /*
@@ -74,6 +90,8 @@ load_sb_action(
     jinfo->last_block = big_tsk_getu32(sb->num_blk) - 1;
     jinfo->start_blk = big_tsk_getu32(sb->start_blk);
     jinfo->start_seq = big_tsk_getu32(sb->start_seq);
+    jinfo->feature_compat = big_tsk_getu32(sb->feature_compat);
+    jinfo->feature_incompat = big_tsk_getu32(sb->feature_incompat);
 
     return TSK_WALK_STOP;
 }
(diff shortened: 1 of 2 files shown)


simsong:
@bjangelo what do you think of the @copilot's recommendation?
I've reviewed it and I'm inclined to go with it. But I want to add all of the disk images to the checks so that this code is tested, which is it currently not.

bjangelo:
> @bjangelo what do you think of the @copilot's recommendation? I've reviewed it and I'm inclined to go with it. But I want to add all of the disk images to the checks so that this code is tested, which is it currently not.

This code is ripped directly from the Linux kernel.

Here is the function that calculates the number of bytes per journal descriptor entry:

- https://github.com/torvalds/linux/blob/master/fs/jbd2/journal.c#L2667-L2683

Here are the associated data structures:

- https://github.com/torvalds/linux/blob/master/include/linux/jbd2.h#L179-L199

Here is a function that uses the above code to count the number of journal descriptor entries in a block.

- https://github.com/torvalds/linux/blob/master/fs/jbd2/recovery.c#L198-L223

And here is the code that writes journal block descriptor tags.

- https://github.com/torvalds/linux/blob/master/fs/jbd2/commit.c#L676-L681

So this PR just mirrors the Linux kernel code, which certainly appears to use `sizeof(journal_block_tag_t) + sizeof(__u16)`, or 14 bytes (assuming 64-bit feature), although I’m not sure exactly what the extra 2 bytes are for.

The kernel documentation (https://www.kernel.org/doc/html/latest/filesystems/ext4/journal.html#descriptor-block) states the following and does not mention a 14 byte size.

```
If JBD2_FEATURE_INCOMPAT_CSUM_V3 is NOT set, the journal block tag is defined as struct journal_block_tag_s, which looks like the following. The size is 8, 12, 24, or 28 bytes:
```

The following commit introduced journal checksum v3.

https://github.com/torvalds/linux/commit/db9ee220361de03ee86388f9ea5e529eaad5323c#diff-b91fc16613e147acaa71465410bb3a0f1ea9714eb103332edcc9ae9841625c9e

The following bit of the commit message appears relevant.
```
It turns out that there are some serious problems with the on-disk
format of journal checksum v2.  The foremost is that the function to
calculate descriptor tag size returns sizes that are too big.
```
So it seems that the Linux kernel journal checksum v2 had a bug that over-calculated descriptor entry size and that bug is now a feature for compatibility reasons.

With all that being said, I have not myself parsed an ext4 journal with checksum v2.

simsong:
Well, AI is good about stealing code, is it not?
I'll try to create a test using some ext4 file systems. I want this code to
be tested as part of our CI/CD system! Give me a few days.

bjangelo:
No sweat.

Please let me know if I can help out with anything

simsong:
If you have any ideas on the right way to structure the tests, I'm all ears!

bjangelo:
I’ve primarily been generating test ext2 and ext4 file systems with a small journal, manually parsing the descriptor block via `jcat`, and comparing against output from `jls`.

https://github.com/sleuthkit/sleuthkit/issues/1475#issuecomment-3282621831

So simplest thing might be just comparing `jls` output against expected/known good (via manual checking) output?

simsong:
If you give me your disk images, that would help me a bit. Can you zip them up and attach them to this ticket? I'll do the rest.

bjangelo:
Will do.

Looking over this PR again, I realize we need to support printing 64-bit block numbers with `fs_blk_hi` when 64-bit feature is enabled

simsong:
Great! Our tiny disk images won’t be of help testing those…

bjangelo:
Updated to support printing 64-bit block numbers from journal entries.

bjangelo:
Test ext2/ext4 file-systems along with expected jls output:

- https://github.com/sleuthkit/sleuthkit/issues/1475#issuecomment-3291978904

bjangelo:
Please note that this PR would obviate https://github.com/sleuthkit/sleuthkit/pull/1477

simsong:
This is great, but I need to create a test and I just haven't had a chance to do that yet. I hope to have the time to do it over the next few days. Sorry for the delay.

bjangelo:
@simsong 

Anything I can do to help drive this to merge?

simsong:
Hi @bjangelo  — Thanks for bringing this up again.
I haven't had the time to put together a test for this. However, I can easily put an ext4 disk image into the digitalcorpora zip file if none of the existing test images will work. So if you want to drive this forward:

- [ ] See if there are any test images in our current set of test images that are loaded as part of the test procedure that can be used to craft a test that will test the code that you have added.
- [ ] If not, give me the URL of a test image that does and I'll add it to the zip file.
- [ ] Review the tests that are part of the CI/CD actions.
- [ ] Add a new test or expand one of our current tests so that the code you have contributed is covered under the unit test.

Thanks!

Simson

PR #3394, opened 2025-10-26
Added test_nsrl.cpp in test'/tsk/hashdb for NSRL specific fxns

singhpratik5:
1. Both NSRL formats included
2. SHA1 and MD5 variants included

1) nsrl_test function tests:
NSRL Format 1 recognition
NSRL Format 2 recognition
Rejection of invalid databases
Wrong prefix headers
Too-short files
Empty files
Invalid header formats

2) nsrl_open function tests:
Opening Format 1 databases
Opening Format 2 databases
Correct database type setup
nsrl_makeindex function tests:
Format 1 with SHA1 hash type
Format 1 with MD5 hash type
Format 2 with SHA1 hash type
Format 2 with MD5 hash type
Empty database handling (should fail)
Invalid data lines handling (should skip)

3) nsrl_getentry function tests:
Format 1 with SHA1 lookups
Format 1 with MD5 lookups
Format 2 with SHA1 lookups
Format 2 with MD5 lookups
Callback variations (stop, error, collect)
Hash not found scenarios
Invalid offset handling
Wrong hash length handling

4) Static functions (get_format_ver, nsrl_parse_sha1, nsrl_parse_md5, is_valid_nsrl) covered indirectly


singhpratik5 (commented):
Didn't work.

test/tsk/hashdb/test_nsrl.cpp
```diff
@@ -274,6 +278,10 @@ TEST_CASE("nsrl_open format 2")
 	REQUIRE(f != nullptr);
 	create_nsrl_format2_db_file(f.get());
 
+	// Ensure data is written and file is at beginning
+    fflush(f.get());
+    rewind(f.get());
+
```
simsong, line 285: This is a really non-standard idiom. Why are you using it?

simsong (changes requested):

test/tsk/hashdb/test_nsrl.cpp
singhpratik5: @simsong The create_nsrl helpers write data to the stream but leave the file open. I added fflush to ensure all buffered data is committed before nsrl_open attempts to access it, and rewind to reset the file pointer to the beginning. It didn't work, will remove them.

simsong: @singhpratik5  - tsk_make_simple_tmpfile() returns a FILE *. Can't you do a `fflush(f)` and a `fseek(f,0L,0)`? My understanding is that violating the `FILE *` abstraction layer by manipulating the underlying file descriptor can cause problems with the buffering.

singhpratik5: @simsong fflush(f.get()) and rewind(f.get()), which are both standard stdio.h functions that operate strictly on the FILE* stream, effectively the same as doing fseek(f, 0L, SEEK_SET). Is there a specific reason to prefer fseek over rewind in this context? Either way I will remove these lines.

simsong:
***@***.***, in my experience f.get() isn't used much. If you can
rewrite this code so that it just uses fflush and rewind, without using
f.get(), that would be great. I'm still unclear why you are calling
f.get().*

singhpratik5:
f is defined as a std::unique_ptr (RAII wrapper). I am calling f.get() to extract the raw FILE* pointer needed for fflush. I will move the fflush and rewind calls inside the create_nsrl_ helper functions, or remove them so it won't be called.

simsong:
Ah, that's my confusion. Okay, there's nothing wrong with this. Perhaps a
comment would be useful, but I'm fine with it. Thanks for the clarification.

singhpratik5 (commented):
Completed, no changes needed as of now. Will update later...

simsong:
Is this ready for review?

PR #3398, opened 2025-10-31
Patch more tests

simsong:


simsong:
As a result of this test, we now capture the amount of free space on each of the virtual machines. Yay!

PR #3407, opened 2025-11-29
Added unit tests for sqlite hdb (test_sqlite_hdb.cpp)

singhpratik5:
More to go...


simsong (approved):
Great going. Probably a bit too much whitespace, but no need to fix that for now.

Issue #2641, opened 2021-12-22
apfs: multiple use-after-free issues that need attention


joachimmetz:
@jtsylve @bcarrier can you address multiple use-after-free issues in the APFS code also see: https://bugs.chromium.org/p/oss-fuzz/issues/list?q=sleuthkit&can=2

joachimmetz:
Corresponding [CWE 416](https://cwe.mitre.org/data/definitions/416.html)

jtsylve:
I'm disappointed that the implementation has some issues, but I don't have the extra cycles to maintain the codebase myself for TSK.  We've abandoned this implementation internally in favor of a rewrite that I unfortunately can't open source.

joachimmetz:
@jtsylve that is unfortunate to hear. 

@bcarrier can you add a statement that this part of the codebase is no longer maintained and that from a security perspective people are recommended to add additional controls like a security sandbox.

Issue #2625, opened 2021-11-28
Null-dereference READ in APFSJObjTree::APFSJObjTree


joachimmetz:
@jtsylve PTAL https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34905&q=sleuthkit&can=2 looks like an NULL-deference issue in the initializer list of a constructor https://github.com/sleuthkit/sleuthkit/blame/develop/tsk/fs/apfs_fs.cpp#L24

This makes this issue very time consuming to debug, since you wrote the code it might be quicker for you to solve it.

joachimmetz:
@jtsylve @bcarrier he folks these are easy to address issues from the perspective of someone that knows the code base what is the ETA on this?

joachimmetz:
Corresponding [CWE 476](https://cwe.mitre.org/data/definitions/476.html)

joachimmetz:
Related: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34905&q=sleuthkit&can=2

joachimmetz:
@jtsylve @bcarrier this looks like a more wide spread issue in the APFS code:

* https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=35062&q=sleuthkit&can=2
* https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=33818&q=sleuthkit&can=2

PR #3444, opened 2026-03-01
Fix overflow bug in APFS code. Reported by Mobasi. Also brought in some build changes from develop and fixed some build issues. 

bcarrier:


joachimmetz:
@bcarrier heads up there are many similar (maybe even worse) issues currently unaddressed

* https://github.com/sleuthkit/sleuthkit/issues/2641
* https://github.com/sleuthkit/sleuthkit/issues/2625

PR #3447, opened 2026-03-02
Rejistry++ memory leak and correct bounds checks

bcarrier:


rejistry++/src/ByteBuffer.cpp
```diff
@@ -97,7 +96,7 @@ namespace Rejistry {
             }
 
             memcpy(&dst[0], &_buffer[_position + offset], length);
-            _position += offset;
+            _position += length;
```
crayy8, line 99: This is still wrong? Shouldn't this be _position += (length + offset)

rejistry++/src/ByteBuffer.cpp
```diff
@@ -73,19 +73,18 @@ namespace Rejistry {
         }
 
         /**
-         * Throws exception if offset or length are too large. */
+         * Throws exception if offset or length are too large. 
+         * @param offset Offset within this buffer to read from
```
crayy8, line 77: This isn't correct or at least not clear, it should be offset from the current position in the buffer

rejistry++/src/ByteBuffer.h
```diff
@@ -53,21 +53,21 @@ namespace Rejistry {
 
         /**
          * Copy 'length' bytes from this buffer into the given destination,
-         * starting at the current position in this buffer and at the given
-         * offset in the destination. The position of this buffer is incremented
+         * starting at offset in the current buffer.
+         * The position of this buffer is incremented
```
crayy8, line 57: I believe this should be buffer is incremented by length + offset

crayy8 (changes requested):

rejistry++/src/HBIN.cpp
```diff
@@ -51,13 +52,19 @@ namespace Rejistry {
     Cell::CellPtrList HBIN::getCells() const {
         uint32_t nextCellOffset = FIRST_CELL_OFFSET;
         Cell::CellPtrList cellList;
+        uint32_t hbinSize = getRelativeOffsetNextHBIN();
 
         do {
             Cell::CellPtr nextCell = new Cell(_buf, getAbsoluteOffset(nextCellOffset));
+            uint32_t cellLen = nextCell->getLength();
+            if (cellLen == 0 || nextCellOffset + cellLen < nextCellOffset) {
+                delete nextCell;
```
crayy8, line 61: we need to free everything in cellList as well

jayaramcs (changes requested):
Not ready for merge.

rejistry++/src/RIRecord.cpp
```diff
@@ -70,6 +70,7 @@ namespace Rejistry {
         for (it = subkeyList.begin(); it != subkeyList.end(); ++it) {
             NKRecord::NKRecordPtrList nkRecordList = (*it)->getSubkeys();
             finalNKRecordList.insert(finalNKRecordList.end(), nkRecordList.begin(), nkRecordList.end());
+            delete *it;
```
crayy8, line 73: Note I've made changes to use unique pointers so we do not need to worry about deleting to handle exception cases. So this will be removed in a following commit.

rejistry++/src/RegistryKey.cpp
```diff
@@ -86,11 +87,14 @@ namespace Rejistry {
 
 
     size_t RegistryKey::getSubkeyListSize() const {
-        std::vector<RegistryKey *> subkeys;
         SubkeyListRecord::SubkeyListRecordPtr subkeyListRecordPtr = _nk->getSubkeyList();
         NKRecord::NKRecordPtrList nkRecordList = subkeyListRecordPtr->getSubkeys();
+        size_t sz = nkRecordList.size();
+        for (NKRecord::NKRecordPtrList::iterator it = nkRecordList.begin(); it != nkRecordList.end(); ++it) {
```
crayy8, line 93: this is still good.

rejistry++/src/HBIN.cpp
crayy8: Addressed in linked PR.

rejistry++/src/ByteBuffer.h
crayy8: This method is supposed to be an implementation of the java ByteBuffer class so some of these changes need to be reverted. This will be addressed in the linked PR. 

https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#get-byte:A-int-int-

rejistry++/src/RegistryByteBuffer.cpp
```diff
@@ -177,7 +177,7 @@ namespace Rejistry {
         uint32_t i = 0;
         uint32_t pos = 0;
 
-        while (i < data.size()) {
+        while (i + 1 < data.size()) {
```
crayy8, line 180: this caused a regression when the data is not null terminated. Fixed in the linked PR.

PR #3458, opened 2026-03-30
Compiler warning fixes

bcarrier:


bcarrier:
The TSTRNCPY fixes were all because the source length was being used as the limiter versus the destination.

bcarrier:
This fixes all of the automated build issues with the develop-4.14 branch. Built with Claude. Lots of casting fixes and build infrastructure changes that were ultimately changed to get 'make distcheck' to work. Lots of path issues with java projects with that, so I ultimtely change the CI code to manually unpack and test.

tsk/fs/logical_fs.cpp
```diff
@@ -284,7 +288,7 @@ convert_wide_string_to_utf8(const wchar_t *source) {
 		if (tsk_verbose)
 			tsk_fprintf(stderr,
 				"convert_wide_string_to_utf8: error converting logical file name to UTF-8\n");
-		strncpy(dest, invalidName, strlen(invalidName));
+		strncpy(dest, invalidName, strlen(invalidName) + 1);
```
APriestman: I think this is correct, but it also seems like it would be more clear to have the length be maxUTF8len since dest was allocated as:
char *dest = (char*)tsk_malloc(maxUTF8len);

APriestman (approved):

tsk/fs/logical_fs.cpp
bcarrier: OK. Changed it to snprintf.

APriestman (approved):

Issue #3476, opened 2026-04-18
issues with patch for cve-2026-40024


joachimmetz:
Numerous shortcomings flagged in https://github.com/sleuthkit/sleuthkit/commit/a3f96b3bc36a8bb1a00c297f77110d4a6e7dd31b which is supposed to address [cve-2026-40024](https://nvd.nist.gov/vuln/detail/CVE-2026-40024) have been mailed to security@sleuthkit.org

Unclear if these have been received due to https://github.com/sleuthkit/sleuthkit/issues/3474

This likely means that the path traversal issue in cve-2026-40024 has **NOT** been fully addressed in 4.15 and additional issues have been introduced

bcarrier:
@joachimmetz I think #3477 addresses them. Do you agree?

joachimmetz:
left a comment, there might be still a case not caught; but likely does not occur given the string allowed to truncate. Here is where unit tests could be beneficial.

Also note that replacing certain characters with `^` is not forensically sound, also see: https://github.com/sleuthkit/sleuthkit/issues/1894

PR #3477, opened 2026-04-19
Fix for issue in #3476 - tsk_recover escape sequences

bcarrier:
## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer error reporting when an output path is too long, preventing silent truncation.
  * Strengthened directory-traversal protection: improved sanitization detects and neutralizes leading/traversal patterns and trailing traversal fragments on both Windows and non-Windows systems.

### DIFF
--- a/tools/autotools/tsk_recover.cpp
+++ b/tools/autotools/tsk_recover.cpp
@@ -11,6 +11,7 @@
 
 #include "tsk/tsk_tools_i.h"
 #include <locale.h>
+#include <stdio.h>
 #include <sys/stat.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


joachimmetz:
Will take a closer look, at first glance it seems to address the obvious issues. Also strongly recommended to add tests

tools/autotools/tsk_recover.cpp
```diff
@@ -154,9 +164,12 @@ uint8_t TskRecover::writeFile(TSK_FS_FILE * a_fs_file, const char *a_path)
...
 
-        // make sure there is no \..\ path traversal
-        if (i + 4 < ilen) {
-            if ((path8[i] == '\\') && (path8[i+1] == '.') && (path8[i+2] == '.') && (path8[i+3] == '\\')) {
+        // make sure there is no \..\ or trailing \.. path traversal
+        // note: i+3 == ilen is intentional — path8[ilen] is the null terminator,
+        // which matches '\0' to catch \.. at end-of-string as well as \..\ scenario
+        if (i + 3 <= ilen) {
+            if ((path8[i] == '\\') && (path8[i+1] == '.') && (path8[i+2] == '.') &&
+                (path8[i+3] == '\\' || path8[i+3] == '\0')) {
```
joachimmetz, line 172: what if `..` is the end of the string but the string was too big for snprintf and it truncated the result?

joachimmetz:
style nit: maybe strip trailing white-space in this file

tools/autotools/tsk_recover.cpp
joachimmetz: looks like this might be handled by line 146, assuming `sizeof(path8)` never is larger than INT_MAX

PR #3499, opened 2026-05-04
Logical FS performance: per-directory file-list cache and inum cache …

crayy8:
…improvements

- Add per-directory file-list cache (DIR_FILE_LIST_CACHE, 500 FIFO slots) to avoid repeatedly enumerating the same directory when resolving K files in one folder during inum-based searches.

- Store inum_cache paths relative to base_path instead of full OS paths. Saves memory and keeps LOGICAL_INUM_CACHE_MAX_PATH_LEN evaluating only the meaningful portion of the path.

- Bump LOGICAL_INUM_CACHE_LEN from 3000 to 50000. Combined with break- on-first-empty-slot optimization in scan loops, lookups remain bounded by actual fill level, not array size.

- Opportunistically cache visited directories during inum searches with always_cache=false (only fills empty slots; never evicts useful ones).

- Use alloc-before-evict pattern for inum_cache and dir_file_list_cache to avoid leaving slots in a stale "key present, data empty" state on malloc failure.

- Add get_path_relative_to_base helper with debug assertion to verify the base_path prefix invariant in one place rather than scattering unchecked pointer arithmetic across four call sites.

- Misc fixes: initialize target_inum and check tsk_malloc result in create_path_search_helper; check GetFullPathNameW return value in create_search_path_long_path.

- Add comment to load_dir_and_file_lists_win documenting future FindFirstFileEx + FIND_FIRST_EX_LARGE_FETCH optimization (Win7+).

## Release Notes

* **Bug Fixes**
  * Improved error handling in cache allocation and cleanup to prevent memory leaks and data structure inconsistencies.

* **Performance**
  * Added directory file-list caching to reduce repeated enumeration operations on Windows.
  * Increased cache capacity for inode and path lookups for faster file system operations.

* **Improvements**
  * Optimized path caching semantics for more efficient storage and retrieval.

### DIFF
--- a/tsk/fs/logical_fs.cpp
+++ b/tsk/fs/logical_fs.cpp
@@ -20,6 +20,7 @@
 #include <set>
 #include <string.h>
 #include <cwctype>
+#include <assert.h>
 
 #include "tsk_fs_i.h"
 #include "tsk_fs.h"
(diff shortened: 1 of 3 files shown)


crayy8:
Wrong repo. Closing.

PR #3544, opened 2026-08-17
Fix UBSAN invalid-enum-value in ntfs_proc_attrseq (OSSFuzz #542853444…

### LINKED ISSUE #3529: OSS-Fuzz issue 542853444
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=5702757815484416 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=542853444 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

bcarrier:
… / GH #3529)

ntfs_proc_attrseq reads a 32-bit attribute type directly from the MFT record on disk and passes it to tsk_fs_attr_set_run / tsk_fs_attr_set_str, both of which store the value in a TSK_FS_ATTR_TYPE_ENUM field.

UBSAN caught that the value 0xE000003F is not a valid enumerator for TSK_FS_ATTR_TYPE_ENUM.  The highest defined NTFS attribute type is 0x100 (TSK_FS_ATTR_TYPE_NTFS_LOG / $Logged_Utility_Stream).  Any value above 0x100 from a disk image is either corrupted or a vendor extension not recognized by TSK.

Fix: after reading the type from disk, skip (continue) attributes whose type exceeds 0x100 rather than propagating an invalid enum value into the TSK_FS_ATTR structure.

Fixes: OSSFuzz issue 542853444
Fixes: https://github.com/sleuthkit/sleuthkit/issues/3529

### DIFF
--- a/ossfuzz/status.md
+++ b/ossfuzz/status.md
@@ -1,8 +1,0 @@
-# OSSFuzz Issue Status
-
-This file tracks the status of every OSSFuzz issue we have triaged.
-
-| OSSFuzz ID | GitHub ID | Branch | Status | Notes |
-|------------|-----------|--------|--------|-------|
-| 542909776 | 3530 | 20260817-gh3530-ossfuzz542909776-fatxxfs-name-leak | FIXED | Direct-leak: fs_name leaked on two early-return paths in fatxxfs_dent_parse_buf |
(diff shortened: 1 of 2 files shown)


bcarrier:
@coderabbitai review

PR #3545, opened 2026-08-17
Fix fs_name leak in fatxxfs_dent_parse_buf (OSSFuzz #542909776 / GH #…

### LINKED ISSUE #3530: OSS-Fuzz issue 542909776
OSS-Fuzz has found a bug in this project. Please see https://oss-fuzz.com/testcase?key=6409256514682880 for details and reproducers.

This issue is mirrored from https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=542909776 and will auto-close if the status changes there.

If you have trouble accessing this report, please file an issue at https://github.com/google/oss-fuzz/issues/new.

bcarrier:
…3530)

fatxxfs_dent_parse_buf allocates an fs_name object at function entry but two early-return paths did not free it before returning:

1. Line 380: when tsk_fs_dir_walk_internal fails while searching for a parent directory, the function returned TSK_OK without freeing fs_name.

2. Line 405: when fatfs_dir_buf_add fails, the function returned TSK_ERR without freeing fs_name.

Both paths now call tsk_fs_name_free(fs_name) before returning. The existing path at line 429 (normal exit) was already correct.

Fixes: OSSFuzz issue 542909776
Fixes: https://github.com/sleuthkit/sleuthkit/issues/3530

### DIFF
--- a/ossfuzz/status.md
+++ b/ossfuzz/status.md
@@ -1,7 +1,0 @@
-# OSSFuzz Issue Status
-
-This file tracks the status of every OSSFuzz issue we have triaged.
-
-| OSSFuzz ID | GitHub ID | Branch | Status | Notes |
-|------------|-----------|--------|--------|-------|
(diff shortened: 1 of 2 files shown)


bcarrier:
@coderabbitai review

